PR #2, opened 2016-03-23
Fix live indexing of inventory and prices into search engine after legacy Magento reindex.

romainruaud:


PsmIndia:
You need to remove database tables too, after than remove all modules files, 
Than
1. remove cache, 
2. setup:upgrade
3. de compile
4. genrate static content
5 reindex
6 again clear cache...

thats it

PR #15, opened 2016-04-19
Connection with the Search API

ddattee:
Adding Magento\Api\Search overriding class to use elastic search in the nativ REST API


ddattee:
I'm not quite sure this is the right way to do it.
Also the Request building process need some checking because I don't get any result yet even thought the search seem to go throught.

ddattee:
I updated the PR with the last modification from master to ease acceptance.

afoucret:
Outdated and replaced by PR #700

PR #23, opened 2016-05-12
Fix autocomplete results number

romainruaud:


romainruaud:
No need to merge this one, its also included in #25

PR #29, opened 2016-05-13
Adding a configurable thesaurus module with back-office CRUD.

romainruaud:
Added :

- basic CRUD in Back-Office
- expansions & synonyms types
- index invalidation on save/delete
- basic filtering on grid

Missing : 

- Unicity check between several wordbags of a thesaurus.


afoucret:
As previous of the module has been released, schema is not ok.
We should use update instead of upgrade.

afoucret:
Merged at @b99ee593f5ff428a43f51a5e48d1ecb833be77bf

PR #33, opened 2016-05-19
Feature abstract indexer

Fdec:


afoucret:
I just took a look very quickly at this one but I think we can do a lot better than what @romainruaud have proposed in his EAV refactoring (especially on the index handler part).

I see an abstract class and many implementations where the only parts that are changed is actually the index and the type given to the constructor. 

IMHO, there is a more clever way to get this done by using the DI virtual types which are a good mean to achieve this without declaring new classes for each new index handler.

I will take a look at this tomorrow or Monday and propose an implementation using virtual types.

afoucret:
Replaced by PR #34

PR #39, opened 2016-06-03
Fix inventory management Fix #35

### LINKED ISSUE #35: Inventory management
- Is in stock indexing for configurable products
- Apply the config to autocomplete

romainruaud:


afoucret:
Not huge fan of injecting ResourceConnection into src/module-elasticsuite-catalog/Plugin/Indexer/AbstractIndexerPlugin.php and handling SQL here.

Would be better if we inject a resource model and working with it.

romainruaud:
I have updated the code.

PR #47, opened 2016-06-03
Move the application of configuration to names attributes in the inst…

romainruaud:
…all data instead of install schema.


romainruaud:
This will fix #47

PR #64, opened 2016-06-14
Indexing child attributes when it is not into parent data

delyriand:


afoucret:
I will merge the PR #65 instead of this one.

PR #101, opened 2016-08-31
Conf product may have conf. attribute not used in layered nav

vo1:
When configurable attribute is not used in LayeredNav, L155 fails indexation due to attribute is not in array.

E.g. you have clothing with color and size, where color is always some complex naming "crimson/blood red/peach" and color_group for this is "red". Color is not used in layered while color_group is. Such solution results in Notice: Undefined offset: <colorAttributeId> which is fixable by isset proposed here.


romainruaud:
@afoucret can you merge this one ? This is project critical for @vo1 .

PR #110, opened 2016-09-13
Fix categories scheduled indexing

### LINKED ISSUE #107: Jenkins Build fails due to Elastic Suite
Hey,

on a project where we are using this module, all our custom setups which create new categories fail with the exception _"Could not parse URI"_ in _vendor/elasticsearch/elasticsearch/src/Elasticsearch/ClientBuilder.php_ on line 545, even though these setups have nothing to do with ElasticSearch. The relevant code part causing this exception is rather simple and looks roughly like this (shortened for better Github legibility):

            $category = $this->categoryFactory->create();
            $category->setData($data);
            $this->categoryRepository->save($category);

It appears as if the ElasticSuite module hooks into some event and then tries to establish an (unnecessary?) connection to the local ElasticSuite instance via 127.0.0.1:9200. However, this fails and causes the aforementioned exception to be thrown. The issue only appears to be happening when Jenkins runs the setup scripts, I assume because there is no local instance of ES or similar, but I am not entirely certain.

This raises the question for me why ES is hooking into the setup script at that point. I am also wondering if one possible solution would be to let this part of the code fail more gracefully, though I am not sure what the ramifications would be for actual front-end logic, as I am currently mainly concerned about our Jenkins builds here.

This is on Magento 2.1 and PHP 7. Let me know if there are any open questions.

romainruaud:
Should close #107 .

To merge on master & also 2.2.x

### DIFF
--- a/src/module-elasticsuite-catalog/Plugin/Indexer/Category/Save/ReindexCategoryAfterSave.php
+++ b/src/module-elasticsuite-catalog/Plugin/Indexer/Category/Save/ReindexCategoryAfterSave.php
@@ -53,7 +53,9 @@ class ReindexCategoryAfterSave
 
         if ($subject->getLevel() > 1) {
             $categoryIndexer = $this->indexerRegistry->get(\Smile\ElasticsuiteCatalog\Model\Category\Indexer\Fulltext::INDEXER_ID);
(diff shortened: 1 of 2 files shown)


afoucret:
To be merged into branch 2.2.x

Issue #77, opened 2016-07-10
Memory Exhaustion when Viewing Certain Categories


phronesis:
For a certain category and it's sub categories, we are experiencing memory exhaustion. This is actually happening for 1 category(Home Appliances) and most of its sub categories. It does not have the most number of products but this still happens. Just to test, i have tried increasing the php allowed memory size to 1G just to test. 
When this happens, a blank page shows on the frontend. My Catalog size is 9k and I am running elastic search on 2 aws ec2 servers with 2G each.

phronesis:
Error From Logs:
PHP Fatal error:  Allowed memory size of 1077936128 bytes exhausted (tried to allocate 37748736 bytes) in /var/www/releases/d-ZYPRQ8DEG/vendor/smile/elasticsuite/src/module-elasticsuite-core/Search/Adapter/Elasticsuite/Response/AggregationFactory.php on line 58" while reading response header from upstream, client: 172.31.9.215, server: _, request: "GET /home-appliances.html HTTP/1.1", upstream: "fastcgi://unix:/var/run/php/php7.0-fpm.sock:",...

phronesis:
Taking a closer look at it shows that memory gets exhausted when carrying out price aggregations. It happens in my Home Appliances category because it has the highest prices ranges: between like 5k and 2million. My immediate need will be to disable price aggregation and price slider and instead stick with the price layering my custom theme provides.

phronesis:
I found how to remove prices from layered navigation in product attributes. That solved it, temporarily. I think you should look into changing price steps in layered navigation, your implementation makes price step 1 and does not seem to take into consideration the settings in Back Office Catalog Layered Navigation settings.

afoucret:
Hi @phronesis,

We will have a look at this one.
To be honest it is quite unusual to have such an high price range and we had not managed this case.

phronesis:
Thanks @afoucret. I understand. We are in Africa where the exchange rate of the dollar or Euro is quite high.. lol. That's why we have such price ranges.

afoucret:
I traveled in Africa few months ago and I had the feeling of being a millionaire with the change rate between CFA and Euros.

We will try to provide a fix.

romainruaud:
Hello @phronesis , you can check the PR #111 which takes into account the fact that the config parameter "Price Navigation Step Calculation" is set to "Manual" and then apply the value of "Default Price Navigation Step" for the interval.

[...]

PR #111, opened 2016-09-13
Adding the ability to use the Manual mode for price navigation step, …

romainruaud:
Add the ability to use the Magento configuration for the Price Navigation Step Calculation : the "Manual" mode is now supported and will use the "Default Price Navigation Step" field when rendering the price slider.

This can allow to have much lighter aggregations when having a wide range of prices displayed, like in #77 

Could be merged only in 2.2.x since it's more a new feature than a bugfix.

### DIFF
--- a/src/module-elasticsuite-catalog/Block/Navigation/Renderer/PriceSlider.php
+++ b/src/module-elasticsuite-catalog/Block/Navigation/Renderer/PriceSlider.php
@@ -12,7 +12,9 @@
  */
 namespace Smile\ElasticsuiteCatalog\Block\Navigation\Renderer;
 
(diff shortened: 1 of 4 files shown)


southerncomputer:
how about adding min_doc_countclass Histogram
`
{
    /**
     * Build the aggregation.
     *
     * @param BucketInterface $bucket Histogram bucket.
     *
     * @return array
     */
    public function buildBucket(BucketInterface $bucket)
    {
        return ['histogram' => ['field' => $bucket->getField(), 'interval' => $bucket->getInterval(),'min_doc_count'=>1]];
    }
}
`
this helped with massive amounts of price variance!

PR #119, opened 2016-09-16
Ensure avoiding processing NULL query if there is no relations for in…

### LINKED ISSUE #118: Empty query processed during indexation
Thanks to @southerncomputer I can confirm that there are queries processed with `entity_id IN (NULL)` clauses during index process.

Queries are looking like this in the query log : 

<pre>
## 16820 ## QUERY
SQL: SELECT `t_default`.`entity_id`, `t_default`.`attribute_id`, COALESCE(t_store.value, t_default.value) AS `value` FROM `catalog_product_entity_decimal` AS `t_default`
 LEFT JOIN `catalog_product_entity_decimal` AS `t_store` ON t_default.entity_id = t_store.entity_id AND t_default.attribute_id = t_store.attribute_id AND t_store.store_id= 3 WHERE (t_default.store_id=0) AND (t_default.attribute_id IN (156)) AND (t_default.entity_id IN (NULL))
AFF: 0
TIME: 0.0017
</pre>

This occurs on https://github.com/Smile-SA/elasticsuite/blob/master/src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/AttributeData.php#L46

The cause is there is no parent-child relation for the concerned products. Therefore, $allChildrenIds is empty and cause the useless request to be sent.

romainruaud:
…dexed products.

This will fix #118 .

Should be merged into master and 2.2.x

### DIFF
--- a/src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/AttributeData.php
+++ b/src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/AttributeData.php
@@ -42,16 +42,19 @@ class AttributeData extends AbstractAttributeData implements DatasourceInterface
         $indexData    = $this->addAttributeData($storeId, $productIds, $indexData);
 
         $relationsByChildId = $this->resourceModel->loadChildrens($productIds);
-        $allChildrenIds     = array_keys($relationsByChildId);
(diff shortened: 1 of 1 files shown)


afoucret:
To be merged into branch 2.2.x

Issue #121, opened 2016-09-16
First-time Magento 2 setups fail if they include a category creation


TiEul:
During a first-time setup of Magento, for example when deploying a pre-existing codebase to a new environment, the ElasticSuite module will cause an exception to be thrown if any of the custom setups include a category creation. This cannot be solved by setting the indexers to "scheduled" since the indexer namespace for the the magento binary only becomes available after the first setup has finished, meaning the setup will end up in a deadlock.

afoucret:
Hi @TiEul,

Where did you create the category (InstallSchema or InstallData script) ?

BR,

TiEul:
Hey @afoucret,

they category creations take place in _Data_ scripts. The temporary workaround I came up with looks like [this](https://github.com/TiEul/elasticsuite/commit/99e60216a020158ad345b107af4e9154158489d6#diff-ba938d3152ba4a2c04302bbe42596805), by the way,

romainruaud:
Hello @TiEul .

I am unfortunately unable to reproduce your issue.

What I do : 

I create a raw Magento 2.1 instance with a composer-create-project.
I add smile/elasticsuite 2.2.x to the required modules.
I add a module creating a dummy category (copy pasted your module): 

<pre>
if (version_compare($context->getVersion(), '2.0.0', '<')) {
            $category = $this->categoryFactory->create();
            $category->loadByAttribute('name', 'Homepage Highlighted Products');

            if ($category == false || $category->getId() == false) {
                $category = $this->categoryFactory->create();
                $category->setIsActive(true);
                $category->setName('Homepage Highlighted Products');
                $category->setIncludeInMenu(false);
                $this->categoryRepository->save($category);
            }
        }
</pre>

I run the bin/magento setup:install on the command line.

No problem occurs. Install finishes properly and I can see my category is created.

Elasticsearch is running locally of course, before running the setup. But after the install procedure, no index is created. Indexers are indicated as "invalid" in the back-office as expected after a fresh install.

Are you sure you are using the last version of 2.2.x branch ? For me, this bug has been fixed since the merge of #110 into master and 2.2.x.

Regards,


A

romainruaud:
Now I'll need more informations about the error message you observe, and also the setup process which is running in your environment. 

Regards,

TiEul:
Hey Romain,

thanks for your help, I just updated our composer.json so that we are now grabbing v2.2.5 instead of 2.2.1, I will let you know if that solves the problem.

Thanks again

PR #126, opened 2016-09-20
First-time Magento 2 setups fail if they include a category creation

TiEul:
I suppose this is a valid approach to solving issue #121, since saving and the deletion of specific non-existing indexes is a logical fault, but does not require any further action. If the index does not exist, the owners will still see the exception being thrown when trying to perform a search on the specific store. Pull request should be denied if there's a better option I missed though.


romainruaud:
Hello @TiEul

Fixing by ignoring exceptions is not fixing... 

I'm sorry but we won't merge this one.

Regards,

Issue #100, opened 2016-08-25
RFC : Search products by SKU


fcapovilla:
Searches by SKU currently return 0 results on our Magento 2.1 instances with Elasticsuite 2.2.
Searching for names or other custom attribute values seem to return the correct results.

The "Use in search" configuration is currently activated on our "sku" product attribute, but it seems to have no effect on our results.

The same problem can be observed on the demo store (http://demo.magento-elastic-suite.io/), although I don't know if the SKU attribute was configured to be used in searches there.

Our current configuration is : 
- Elasticsuite 2.2.0-alpha2
- Magento 2.1.0
- Elasticsearch 2.3.5

southerncomputer:
Because SKU is not mapped correctly!
 "sku" : {
            "type" : "string",
            "index" : "not_analyzed",
            "fielddata" : {
              "format" : "doc_values"
            }
          },
          "spelling" : {
            "type" : "string",
            "fields" : {
              "phonetic" : {
                "type" : "string",
                "analyzer" : "phonetic"
              },
              "shingle" : {
                "type" : "string",
                "analyzer" : "shingle"
              },
              "whitespace" : {
                "type" : "string",
                "analyzer" : "whitespace"
              }
            },
            "analyzer" : "standard"
          },

Compared to attribute sprc_sku:

 "sprc_sku_code" : {
            "type" : "string",
            "fielddata" : {
              "format" : "lazy"
            },
            "fields" : {
              "edge_ngram_front" : {
                "type" : "string",
                "fielddata" : {
                  "format" : "lazy"
                },
                "analyzer" : "edge_ngram_front"
              },
              "phonetic" : {
                "type" : "string",
                "fielddata" : {
                  "format" : "lazy"
                },
                "analyzer" : "phonetic"
              },
              "shingle" : {
                "type" : "string",
                "fielddata" : {
                  "format" : "lazy"
                },
                "analyzer" : "shingle"
              },
              "untouched" : {
                "type" : "string",
                "index" : "not_analyzed",
                "fielddata" : {
                  "format" : "doc_values"
                }
              },
              "whitespace" : {
                "type" : "string",
                "fielddata" : {
                  "format" : "lazy"
                },
                "analyzer" : "whitespace"
              }
            },
            "copy_to" : [ "search", "spelling", "autocomplete" ],
            "analyzer" : "standard"
          },

Do you see what is missing?

southerncomputer:
[...]

PR #129, opened 2016-09-21
Adding a "reference" field definition : ability to search by SKU.

romainruaud:
This a proposal of implementation according to discussion in #100 

Tested working against Luma default SKU values (MH06, MS09 etc...) and also with EAN13 values (3459379414678).

Let me know about it.

### DIFF
--- a/src/module-elasticsuite-catalog/Block/Plugin/Adminhtml/Product/Attribute/Edit/Tab/FrontPlugin.php
+++ b/src/module-elasticsuite-catalog/Block/Plugin/Adminhtml/Product/Attribute/Edit/Tab/FrontPlugin.php
@@ -273,6 +273,18 @@ class FrontPlugin
             'search_weight'
         );
 
+        $fieldset->addField(
+            'is_reference_field',
+            'select',
+            [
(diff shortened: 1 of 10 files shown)


romainruaud:
This PR brokes the Travis build because the "Field" class has now a complexity of 51... was 50 before. 
Seems hard to optimize, let me know if we can ignore or not if we decide to merge this one.

southerncomputer:
Have you thought about pushing this patch into 2.3.x? it would help with complex sku field accuracy!

afoucret:
@southerncomputer : 
I am not huge fan of this patch and work on a better solution that should be more generic and address the general problem of using custom analyzer on specific fields.
I fear, you will have to wait 2.4.0 for this patch. At the same time I don't think merging temporary solution is a good idea, since it can have side effect on ascendant compatibility ...

maximehuran:
I need this functionality very quickly, if I switch to the 2.4 branch, will I can search products by SKU ?

afoucret:
This PR will not be merged since a better implementation have been proposed (PR #300)

Issue #143, opened 2016-09-26
Core  / Index / Mapping / Field.php is not compliant with Elasticsearch 2.x


southerncomputer:
function GetPropertyConfig is written for elasticsearch 1.x - 
ie. 'format' => 'doc_values' has no meaning in elasticsearch 2.x ! 

Makes me wonder if any other code areas have been considered since we're on 2.4 version of elasticsearch.

For instance -> I enabled caching on all indexes current - which reduced call time on repetitive searches for 16K per index open.

curl -XPUT localhost:9200/magento2_default_catalog_category,magento2_default_thesaurus/_settings -d'
{ "index.requests.cache.enable": true }
'

Maybe it is time to take a look at the code for further optimization to maximize performance!

I am seeing 3-4 second search times on a fully loaded Java Implementation with many facets!  Viewing a category with 8000+ products is nearly instant but searching for that similar category name of products is taking 3-4 seconds! Seems kind of slow :(

afoucret:
Hi @southerncomputer,

We will check the format of doc_values.

Concerning performances we have a work in progress for huge catalog for Magento 2. It will be focus on both inexing and querying the indices. It is matter of weeks to have something we can communicate on but I have no doubt that ElasticSuite will be able to manage huge catalog (let's say millions of product) then.

afoucret:
PR #144 should provides requested perf enhancement.
Will be part of the 2.3.0 release.

Thx for reporting.

afoucret:
Hi @southerncomputer,

Just for your info, we pushed a lot of performance enhancements on the 2.3.x branch.
I know it is a important concern for you.

A full report in this merged PR : #164

So feel free to test it and report new issue.

southerncomputer:
Yeah I just merged 2.3.x branch with your latest changes. One bug that keeps popping up is XML in catalog.leftnav - It causes the EE version to lose its layered navigation due to incompatibility between ElasticsuiteCatalog/view/frontend/layout/*.xml and the EE configuration. 

In Developer mode, it will throw XML cache merge errors, I will have to research my last time fixing this - I had to move two .xml files away from module-layered-navigation-staging - maybe I need to put this back!

afoucret:
Hi @southerncomputer,

First very glad to see you are continuing to use the module.

1. Can you report us if you observe improvement for your own use case (both a matter of indexing speed and search)

[...]

PR #144, opened 2016-09-27
Enable query caching + Fix doc_values settings

afoucret:


afoucret:
Should improve situation for #143

PR #164, opened 2016-10-13
Feature performance

afoucret:
Global performance improvement :
* Better mappings with removed useless analyzed fields and more control on memory usage
* Better usage of the index settings during full indexing
* Magento caching layer for spellcheck, thesaurus and virtual category rules
* Enforce complex queries caching (eg. bool) for the virtual rules ES filters
* PHP optimization for the option value mapping
* Better SQL loading


Koc:
Awesome work!

@afoucret does wrong typed words still works connectly after removing ngrams?

PR #174, opened 2016-10-18
Add local cache to backend table retrieval : this is called on search…

romainruaud:
… result page and can take some times if there are many searchable attributes.


romainruaud:
I will regroup performance related PR into once when Im done.

Closing this one.

PR #207, opened 2016-11-15
Use price renderer defined in layout to render product prices in auto…

### LINKED ISSUE #194: "Fixed Product Taxe" is not used to display the price
On the auto complete, and search result, if a product has a "fixed product taxe" attribute, then the displayed price does not take it in account.

For example, if a product has a price of 100€ (included the 20% french taxes) + a DEEE of 5€, it displayed:

- Product Page: 105€ (with DEEE 5€)
- Search Result: 100€

thanks

romainruaud:
…complete box.

This should fix #194 but also provide better price display into autocomplete box (special price is now properly rendered).

### DIFF
--- a/src/module-elasticsuite-catalog/Model/Autocomplete/Product/ItemFactory.php
+++ b/src/module-elasticsuite-catalog/Model/Autocomplete/Product/ItemFactory.php
@@ -14,9 +14,11 @@
 
 namespace Smile\ElasticsuiteCatalog\Model\Autocomplete\Product;
 
+use Magento\Framework\App\ObjectManager;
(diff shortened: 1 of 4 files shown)


afoucret:
Hi @romainruaud, the PR is incomplete. Bundle prices are rendered as it : "0,00 €"

romainruaud:
Hello @afoucret 

Can you precise how is your bundle product configured ?

With the sample data I have following rendering for "Sprite Yoga Companion Kit", which seems good to me : 

selection_157

afoucret:
Was actually OK. My bad !!!

PR #224, opened 2016-11-29
Prevent facet html value to be renderer in page.

PascalNoisette:
(more details were supplied by email the 2016/11/18 10:06 UTC+1)


afoucret:
We were not able to merge your PR as it since your base branch was 2.3.x and it was needed from 2.1.x branch. However, your commit have been applied on branch 2.1.x and then merged into branches 2.2.x, 2.3.x and master.

Next time, create the fix from the older branch and we will be able to merge more easily.

afoucret:
I close the PR, since the commit is on the different branches.

Issue #66, opened 2016-06-20
RFC : Virtual categories : improve product page URL when using category path in product URL


PierrickOlivier:
Initial conditions:
- In Store/Configuration > Catalog > Catalog, set "Use Categories Path for Product URLs" to "YES"
- Create a virtual category
- Click on a product

--> The URL of the product page is an URL not rewritten, example: https://www.xxxxxxx.eu/uk/catalog/product/view/id/13672/category/643

We would like to improve the URL in this case to have a "clean" URL like this:
http://base_url/virtual-category-name/product-name

afoucret:
We will try to have this feature released in the 2.2.0 or 2.3.0 version.

Thx for reporting.

afoucret:
Regarding the amount of work still remaining in the PR #227, postponing this one to 2.4.x should be considered to have a cleaner implementation ...

Koc:
@afoucret any news on this issue?

romainruaud:
@androshchuk please have a look on this one, probably my old (and unmerged) PR is a good start since it was nearly finished when I wrote it.

Regards

romainruaud:
Closed by associated PR !

PR #227, opened 2016-11-30
Manage nice url for products displayed in virtual categories when the…

romainruaud:
… category path should be applied to URLs

Implements #66

### DIFF
--- a/src/module-elasticsuite-virtual-category/Controller/Router.php
+++ b/src/module-elasticsuite-virtual-category/Controller/Router.php
@@ -1,0 +1,107 @@
+<?php
+/**
+ * DISCLAIMER
+ * Do not edit or add to this file if you wish to upgrade Smile Elastic Suite to newer
+ * versions in the future.
+ *
+ * @category  Smile
+ * @package   Smile\ElasticsuiteVirtualCategory
+ * @author    Romain Ruaud <romain.ruaud@smile.fr>
+ * @copyright 2016 Smile
+ * @license   Open Software License ("OSL") v. 3.0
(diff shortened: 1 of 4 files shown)


afoucret:
@romainruaud : Build have failed. Can you check please.

romainruaud:
My bad.

I refactored this one and build is now OK.

Let me know

afoucret:
Thx, I will review this one carefully before merging.

romainruaud:
For me it's impossible to use the ```Product\Url``` model to retrieve the Url of product instead of using the ```getUrlKey()``` : 

- The product Url model will try to load from ```url_rewrite``` with the product/category couple and will fail
- The product Url model returns a "full Url" and not a raw "request path"

However, I have added a method to prevent the case where the product does not have any url_key (or when the attribute is not used in the listing) to fallback to the standard ```catalog/product/view/id``` pattern.

Regards

afoucret:
Hi @romainruaud,

I have tested this development and there seems that there is an important test case you did not manage. The following example should be enough to understand it.

I have a virtual category (v1) inside another category and a product matching the rule.
So the product URL will be http://mysite/c1/v1/product.html.
But as every category is anchor in ElasticSuite, the product URL may be http://mysite/c1/product.html. This case seems to be missing especially if c1 is not a virtual category.

afoucret:
Aditionnaly you will have to rebase the branch since there is a conflict with the #231 patch in the `frontend/di.xml` file

afoucret:
This PR is outdated now.
Closed.

Issue #49, opened 2016-06-06
RFC : Virtual categories root


afoucret:
To be done : Automatic generation of a subtree for virtual categories when a root category is set.

romainruaud:
@afoucret I need update/brief about this one.

afoucret:
This one will be postponed to v.2.3

itlinlab:
This feature is essential for two of our projects. Both ecommerce stores sell clothing mixed with other merchandising products, having this feature means that we can have a global category structure for Clothing which every clothing product is linked to a subcategory of that and create Virtual Categories that have a virtual root in Clothing Category that will have as rules something like Gender: Female and Age Group: Kid, in the frontend it would keep the catalog just like a normal category and we can show Girl Products in the virtual category with the user still being able to navigate by for example, t-shirts, skirts, etc.

We were looking for a solution to this and we think it would really be one of the most interesting features to add to Virtual Categories because it really allows for full catalog creation and management flexibility in our ecommerce websites keeping our PIM structures clean and simple and with reduced work load for our managers.

That being said, we greatly appreciate if the community can help us create this, also noticed this is the oldest issue on your issue tracker and saw the request for comments so here is ours. 

We are looking for developers to help out, and created a bounty offer within our magento development partners to see if they can help, the bounty clearly states that if solved they will have to share the solution with you and add a pull request for your use, however we are not using the latest version of smile/elasticsuite in the projects we are working on right now so some refactoring might be needed.

Many thanks to the Smile Team for this absolutely killer extension that we have been using in our magento 2 stores for so long...

Pedro Monteiro - CTO @ LINLAB

romainruaud:
6 years later, this oldest RFC has been implemented in #2489

PR #235, opened 2016-12-07
Feature virtual categories subtree

romainruaud:
This is a PR for #49 

This is partially based on PR #227 so it may need to be refactored if we chose to not integrate this feature at all.

Working : 

- navigating through subtree of a virtual category root on the catalog.
- this is not implemented on search results, since I "lost" the initial virtual category root while drilling through category filter : may be hard to do.
- products URLs are also properly displayed (if configured to be displayed with category path) under category root subtree. 

Feel free to test and let met know.


romainruaud:
I'll fix the build but can you review this one first to ensure the approach was correct ?

afoucret:
This PR is outdated now.

PR #236, opened 2016-12-08
Fixed missing translation in the autocomplete block.

oliverde8:


romainruaud:
Hello @oliverde8 and thank you for submitting this one.

The piece of code you add was removed, intentionnally, to be replaced by the ```translate="true"``` feature of the di.xml file (this is a cleaner approach).

See : https://github.com/Smile-SA/elasticsuite/blob/master/src/module-elasticsuite-catalog/etc/frontend/di.xml#L123

Do you have this piece of code in your project ? Ensure you are up to date if you are using 2.3.x branch. 

If you did inject custom titles via your di.xml, please ensure to have injected them with the ```translate="true"```option.

Best regards,

afoucret:
As a complement, this piece of code was buggy since it broke the i18n collect mechanism of Magento which does not support variables inside the `__` function

oliverde8:
Hi, thank you for the response. 

We are indeed using the latest version of 2.3.x branch and have the translate="true". We have not injected any custom titles. 

We will try and investigate some more. 

Thanks,

PR #254, opened 2016-12-15
Feature relevance configuration refactoring (2.1.3 compatibility)

romainruaud:
This PR is intended to fix the backward incompatible changes made by Magento between 2.1.2 and 2.1.3 versions.

This is essentially occuring on objects inheriting or using the ```Magento\Framework\App\Config``` class (Reader, Scope pools, etc ...).

Some (slights) parts of Relevance Configuration have been refactored to be 2.1.3 compliant.

This piece of code has also been tested against a Magento 2.1.2 version and is working, which should prevent us from having to release several versions of the module in the future.

Let me know,

Best regards,


romainruaud:
This one is closed in favor of #255 which targets the 2.2.x branch

Will be cascaded to 2.3 and master then.

PR #278, opened 2017-01-04
Server-side prevention of setting a virtual category root to herself.

romainruaud:


afoucret:
@romainruaud,

I would better have the category removed from the select in admin.
Then it would be impossible to set a category but also it's child as root.

Additionnaly, having a check while moving a category can be useful to avoid a situation where a category is moved under a category using it as a virtual root.

romainruaud:
Got it, I will propose an enhanced version of this fix then.

romainruaud:
Well you can check this updated PR @afoucret

PR #285, opened 2017-01-05
Feature unit testing

afoucret:


afoucret:
The feature have been merged since the build is broken only because env variables can not be echoed from a PR (security reason) by travis.

PR #305, opened 2017-01-20
Ensure proper usage of row_id/entity_id fields when building relation…

### LINKED ISSUE #304: Bad product reassembled in index
Hi,

We encounter a strange problem in the ES index.

In fact, when we search a bundle (query example :  _search/?q=sku:bundle-403114), an other product is returned as we can see in the name field (which is not one of children of the bundle).

[img of the ES index](http://nsa38.casimages.com/img/2017/01/20/170120125901748909.png)
[img of the Magento BO](http://nsa38.casimages.com/img/2017/01/20/170120125901866764.png)

The wrong product returned  by the query got the row_id field equals to **1568577** which is the entity_id of the researched bundle. It seems there is a confusion between the entity_id and the row_id.

Really thanks for your help.

Théo

romainruaud:
…s data for composite products.

Will fix #304


romainruaud:
I will propose this one on 2.2.x branch instead of 2.3.

Issue #204, opened 2016-11-14
Products and attribute values showing incorrectly


erickmarsh:
I have a client using the extension, and we have 2 related problems:
1. We have attributes appearing in the left nav that are not assigned to products in that category.
Example: We are navigating through men's boots, and seeing a style option for "Ballerina". Filtering by this attribute brings us to a product which does not have Ballerina selected for the Style attribute. (it is blank)
http://www.shoebacca.com/catalog/mens-shoes/boots.html?acu_in_style=Ballerina

2. There are products appearing in filtered results that should not be included.
Example: Filtering to find Women's Running Shoes, and "Ballet Flats" appear in the result set
http://www.shoebacca.com/catalog/womens-shoes/athletic-shoes.html?acu_in_sport=Running&p=2 

I verified that none of the associated simple products have any incorrect attribute values which would be causing the improper filtering.

erickmarsh:
Forgot to mention, we are using:
- Smile Extension 2.2.2
- Mage EE 2.1.0
- PHP 7.0.8
- Ubuntu 16.04
- ElasticSearch 2.4.1

afoucret:
It is quite hard for me to figure what happen there since we have never faced this kind of issue. 

I would suggest that the cause of the problem can be one of those 2 (maybe both together) :

1. Did you installed several instances of Magento / ElasticSuite on the same server (let's say production, staging, testing, ...) ? <br/>In this case, is every instance configured to handle it's own ElasticSearch Index as explained in the [wiki](https://github.com/Smile-SA/elasticsuite/wiki/ModuleInstall) ? At least, you must have a differet settings for the `Indices alias name` configuration param, if you don't want your indices to be messed up between the instances ?

2. Did you try to reindex everything at least once through : <br/>
```bin/magento indexer:reindex```

Hope it helps. Thank you for your feedback on this two ones.

gewaechshaus:
@afoucret - sounds like it's kind of related to #195

afoucret:
@gewaechshaus,

I don't think this tow issue are related. @erickmarsh problem looks more like if the index contains product that are not really synced with the database (wrong facets, wrong ids, ...)

We will wait @erickmarsh responses on this ...

erickmarsh:
We are running 3 completely separate environments (Prod, Stage, Dev) with their own web servers, DB and ES so there should be no conflicts. I have reindexed everything again, just to make sure, and the results are the same. 

[...]

PR #306, opened 2017-01-23
Ensure proper usage of row_id/entity_id fields when building relation…

romainruaud:
…s data for composite products.

This is a fix for #304 . May also fix the old #204 which could be related.


romainruaud:
To be merged on 2.2.x then 2.3.x then 2.4.x

Issue #288, opened 2017-01-09
Latent issue on relevance configuration save


Vedrillan:
The resource model Smile\ElasticsuiteCore\Model\ResourceModel\Search\Request\RelevanceConfig which uses the table smile_elasticsuite_relevance_config_data and inherit from Magento\Config\Model\ResourceModel\Config is not aligned with Magento practice.

The field scope_code differ with Magento usage of scope_id.

```
mysql> desc core_config_data;
+-----------+------------------+------+-----+---------+----------------+
| Field     | Type             | Null | Key | Default | Extra          |
+-----------+------------------+------+-----+---------+----------------+
| config_id | int(10) unsigned | NO   | PRI | NULL    | auto_increment |
| scope     | varchar(8)       | NO   | MUL | default |                |
| scope_id  | int(11)          | NO   |     | 0       |                |
| path      | varchar(255)     | NO   |     | general |                |
| value     | text             | YES  |     | NULL    |                |
+-----------+------------------+------+-----+---------+----------------+
5 rows in set (0.00 sec)

mysql> desc smile_elasticsuite_relevance_config_data;
+------------+------------------+------+-----+---------+----------------+
| Field      | Type             | Null | Key | Default | Extra          |
+------------+------------------+------+-----+---------+----------------+
| config_id  | int(10) unsigned | NO   | PRI | NULL    | auto_increment |
| scope      | varchar(30)      | NO   | MUL | default |                |
| scope_code | varchar(30)      | NO   |     | default |                |
| path       | varchar(255)     | NO   |     | general |                |
| value      | text             | YES  |     | NULL    |                |
+------------+------------------+------+-----+---------+----------------+
```

The method Magento\Config\Model\ResourceModel\Config::saveConfig is expecting to have a field scope_id in the table.

```
$newData = ['scope' => $scope, 'scope_id' => $scopeId, 'path' => $path, 'value' => $value];

if ($row) {
    $whereCondition = [$this->getIdFieldName() . '=?' => $row[$this->getIdFieldName()]];
    $connection->update($this->getMainTable(), $newData, $whereCondition);
} else {
    $connection->insert($this->getMainTable(), $newData);
}
```

When saving configuration from the BO there is no issue because it is using an entity factory, but this issue will happen if using the Magento\Framework\App\Config\ConfigResource\ConfigInterface.

romainruaud:
Fixed on 2.4.x.

PR #310, opened 2017-01-23
Implements the saveConfig method to use correct "scope_code" field.

romainruaud:
Will fix the latent issue described on #288 

Low impact since the bug can occurs only when calling explicitely the ```saveConfig``` or ```deleteConfig``` method, which is not the case when editing the configuration in the Back-Office.

This could only happen on specific code written to use this functions.


afoucret:
As the bug is only a "latent one", no emergency to release a patch for 2.3.x branch.
Merged only into the 2.4..x branch for release with 2.4.0.

PR #314, opened 2017-01-25
Externalize coding standards

romainruaud:
They are now located on dedicated packages : 

https://packagist.org/packages/smile/magento2-smilelab-phpcs
https://packagist.org/packages/smile/magento2-smilelab-phpmd

They have been tagged 1.0.0 in the state they were legacy used for elasticsuite.

Configuration for Travis have been updated, so does our (internal) Spbuilder tool.

PHMD has been removed for CodeClimate since it cannot run with external repositories (and PHPMD is already done via Travis).

I have added Phan for CodeClimate since it can give us more insights for the quality, let me know if you are ok or if I should remove it.


romainruaud:
You can see that the Travis build performed correctly with the new configuration based on external repositories.

We keep in touch,

regards,

PR #331, opened 2017-02-13
Index all prices data for each product.

romainruaud:
This one allows to index all prices related data (with the min and max price).

This could allow later to build price facet to the min_price data if we want : for websites using configurable and bundle products which want to have the price slider values correlated to the way they chose to display prices, eg. using the "From 'min_price' $ " display.

This is also a prelude to the availability of loading prices and stock data directly from elasticsearch instead of doing it via the _applyProductLimitation method.

We can discuss this one together later.

Regards


afoucret:
This is a deprecated PR. I close.

PR #333, opened 2017-02-17
Remove is_used_in_autocomplete from install.

afoucret:


romainruaud:
Are you sure that it does not make the following line fail if the column is not created during install ? 

https://github.com/Smile-SA/elasticsuite/blob/2.4.x/src/module-elasticsuite-catalog/Setup/UpgradeSchema.php#L99

afoucret:
I am sure that the previous version was breaking the install from scratch (InstallData try to set is_used_in_autocomplete on a non existing field).

I think the new version does not break since the `removeIsUsedInAutocompleteField` is only called if module version is  `< 1.2.2`.

The new installer have been played on a Magento fresh install.

PR #336, opened 2017-02-24
Prevent fatal error  when viewing a category not set to is anchor

romainruaud:
Maybe this is not the best way to deal with it, but it prevent the layer plugin to fail with a fatal error when viewing a category not set to is_anchor=1.

Let's discuss about this together,

Regards


romainruaud:
Closed in favor of #338

PR #338, opened 2017-03-01
Fix categories anchor related issues

romainruaud:
- Ensure we can apply methods on the layer plugin : this will prevent fatal error when using a theme or custom code which is instantiating a layer anywhere (home page, etc...)

- Automatically set categories to is_anchor=1 when saving it.

- Automatically set categories to is_anchor=1 just before reindexing the category/products association.

This should get rid of any remaining confusion or mis-usage of the is_anchor and layer strange instanciations.

Let me know


afoucret:
This one is replaced by PR #356.

Reopen if I am wrong.

PR #345, opened 2017-03-07
Debuging preview in BO for virtual categories

ddattee:


afoucret:
This PR will not be merged.
The preview does not contains all children categories products (especially products contained in non virtual category). As a result front will display a product list that is different from what have been previewed in admin.

Issue #342, opened 2017-03-03
Create Categorie Virtual slow the category view on Back-office


zkliouel:
Hi, I have an issue, you can find the test case below : 

Test Case

Get 3000 simple categories (non virtual) with max deep level of 5
Create one virtual category linked to Root category with a simple condition like SKU contains '1' and save it.
Then when I try to display categories view on BO I have the loader loading indefinitely. Even if I click on a non virtual category.

After a code inspection it's due to the method "getQueryFilter" on class "Smile\ElasticsuiteVirtualCategory\Model\Preview"
The conditions : 
```
if ($this->category->getIsVirtualCategory() || $this->category->getId()) {
            $query = $this->category->getVirtualRule()->getCategorySearchQuery($this->category);
        }

        if ((bool) $this->category->getIsVirtualCategory() === false) {
```
It seems you always do an "virtual search" even when the category is not virtual, i change the conditions above with
```

if ($this->category->getIsVirtualCategory()) {
            $query = $this->category->getVirtualRule()->getCategorySearchQuery($this->category);
        } else  { ....
```
And now my non virtual categories show my some products on section "product preview and sorting" instead of the message : "Your product selection is empty for the selected Store View. If you are running a multi-store setup, please check this manual page for more informations."

PS : i already change the elastic config index.query.bool.max_clause_count to 10240 because default config of 1024 cause the error "too_many_clauses: maxClauseCount is set to 1024"

Regards

ddattee:
I had the same issue and this resolved the problem from me to.
Thanks @zkliouel 

Will this be merged quickly ?

romainruaud:
Hello @zkliouel or @ddattee 

thank you for submitting this issue and provide the fix :)

Can one of you propose this via a PR ? 

Regards

ddattee:
Here is a PR #345

romainruaud:
Hello, the PR #345 will not be merged, since it changes the way the module is working, you can check @afoucret answer in the PR.

We did provide another way of optimizing this part via 7c91485dad015e5117853edda3e3ee4dede3a46d which consist in aggregating all non-virtual categories child to a unique query clause, and then processing only the virtual categories children apart.

Your PR #346 has also been integrated, so does the #349 .

Please consider using the 2.3.x branch directly if you want to benefit this fix asap.

Otherwise this will be released in the next minor version of the module.

Best regards,

PR #352, opened 2017-03-14
Virtual category optimization

afoucret:
- Reduce the recursion level for standard categories.
- Merge children categories into an unique should clause when possible


afoucret:
This one fix #342

PR #356, opened 2017-03-16
Fixing direct usage of Layer to prepare a product collection which th…

romainruaud:
…ey did not instantiate themselves.

This is mostly related to the ```Call to undefined method Magento\Catalog\Model\ResourceModel\Product\Collection\Interceptor::addSortFilterParameters()``` error.

This error occurs when Magento uses the Layer object as a filter, to prepare an external product collection.

This can be seen especially on ```Magento\Catalog\Block\Navigation``` : 

`       $productCollection = $this->_productCollectionFactory->create();
        $this->_catalogLayer->prepareProductCollection($productCollection);`

the `$productCollection` is created by the standard product collection factory. Therefore it is not possible to use ElasticSuite method on it when the layer calls the prepareProductCollection.

This can be fixed by switching the factory used to the CatalogSearch one, and will then work properly.

Imho, since we cannot ensure which type of collection is sent to the Layer, and since prepareProductCollection is a public method, we should go further on this one.

Let me know @afoucret

### DIFF
--- a/src/module-elasticsuite-catalog/etc/di.xml
+++ b/src/module-elasticsuite-catalog/etc/di.xml
@@ -191,4 +191,20 @@
         </arguments>
     </type>
 
+    <!-- Ensure automatic instantiation of layer will properly use a fulltext product collection -->
+    <virtualType name="Smile\ElasticsuiteCatalog\Model\Layer\Category\Context" type="Magento\Catalog\Model\Layer\Context">
+        <arguments>
+            <!-- This provider uses the Magento\CatalogSearch\Model\ResourceModel\Fulltext\CollectionFactory as Collection Factory -->
+            <argument name="collectionProvider" xsi:type="object">Magento\CatalogSearch\Model\Layer\Category\ItemCollectionProvider</argument>
+            <argument name="stateKey" xsi:type="object">Magento\Catalog\Model\Layer\Category\StateKey</argument>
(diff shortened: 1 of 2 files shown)


afoucret:
The automatic instantiation of product collection using  \Magento\Catalog\Model\Layer\Category\ItemCollectionProvider should be managed.

I think it can be useful in some case.

romainruaud:
Hello, I just tested the PR with this block : 

```
class Test extends \Magento\Framework\View\Element\Template
{
    private $catalogLayer;

    private $catalogLayer;

    public function __construct(
        \Magento\Framework\View\Element\Template\Context $context,
        \Magento\Catalog\Model\Layer\Resolver $layerResolver,
        array $data
    ) {
        $this->catalogLayer = $layerResolver->get();
        parent::__construct($context, $data);
    }

    protected function _prepareLayout()
    {
        $writer = new \Zend\Log\Writer\Stream(BP . '/var/log/rorua.log');
        $logger = new \Zend\Log\Logger();
        $logger->addWriter($writer);

        $collection = $this->catalogLayer->getProductCollection();
        $collection->addIsInStockFilter();

        $logger->info(" Class : " . get_class($collection));

        return parent::_prepareLayout();
    }
}
```

Which produces no error and outputs : 

```
2017-03-20T15:00:41+00:00 INFO (6):  Class : Smile\ElasticsuiteCatalog\Model\ResourceModel\Product\Fulltext\Collection\Interceptor
```

Issue #376, opened 2017-03-31
Bugs: price filter, price slider


testrafa:
### Preconditions

Magento Version : Magento ver. 2.1.5

ElasticSuite Version : 2.3.4 release

Bugs:
- price filter doesn't work properly, wrong count products amounts (upper interval)

![price_filter](https://cloud.githubusercontent.com/assets/26811962/24541301/252017ee-15f7-11e7-9c6a-998380554e9b.png)

- link 'show more' in filters:

changing value and save doesn't work for some attributes f.e material and style_general.

![show_more](https://cloud.githubusercontent.com/assets/26811962/24541320/3b1ba22a-15f7-11e7-84a8-b86b2d233718.png)

Reproduction: Stores -> Attributes -> Product -> choose any Attribute -> storefront properties -> change "value in facet max size" -> save.
And there is no expected results

- price slider
manually settings doesn't work. After turn on this option, "price" attribute dissapear from frontend .
Option 'display product count' also does not work. There is no results on frontend when I change this option for "no" and save it

![manual_i_ilosc_produktow_w_price](https://cloud.githubusercontent.com/assets/26811962/24541340/5967baca-15f7-11e7-919a-c7ab8b2d7614.png)

romainruaud:
Hello, and thank you for all those feedbacks.

- Price slider : I confirm that there is something to fix on the interval.

- Facet max. size : I do not reproduce your issue. Which value did you test ? I set "3" to "material" and "style_general" and this is working properly.

- Price slider and manual interval : If you keep the default value for "Default Price Navigation Step" which is 100, yes, the price disappear. This is due to the interval being bigger than the product values : all products prices are between 0 and 100, so there is no way to display values with an interval of 100 on the slider. If you set it to 10, everything is working properly.

I wait for more informations on the second point.

best regards,

romainruaud:
I close this one due to being inactive for more than one week.

Please check the PR #381 which is fixing the issue on the price slider.

I am still waiting for an update about the facet max size.

Do not hesitate to reopen the issue when you will have more informations.

regards,

PR #381, opened 2017-04-03
Set upper interval as included in product count.

romainruaud:
This is a fix for #376


romainruaud:
Seems that you were right.

Something is weird here : 

Product A is 68$
Product B is 69$
Product C is 69$
Product D is 69,10$

When building the Slider, the intervals values are given as this : 
```
{"value":68,"count":1}
{"value":69,"count":3}
```

On the slider if I choose "67-69" : it show 4 products, meaning that the product D is included. So my PR is OK here. Without this PR the slider shows 1 product only.

BUT, if I submit the page, I only see products A B and C. Because the query sent to ES is : 

```
"query":{"range":{"price.price":{"gte":"67","lte":"69"}}}
```

Which cannot match a product having a price of 69.10$

So I think this PR should not be merged as is, but rather be reworked a little to cover all cases.



If I set a product to 69.10$ and select eg 67$-69$

afoucret:
As you have said this PR need to be reworked.
I close for being inactive since a while.

Issue #341, opened 2017-03-02
Set "special_price" attribut to "filterable" generate exception 


zkliouel:
Hi,

Follow the stacktrace below : 

```
Exception #0 (BadMethodCallException): Missing required argument $name of Smile\ElasticsuiteCore\Search\Request\Aggregation\Bucket\Histogram.

Exception #0 (BadMethodCallException): Missing required argument $name of Smile\ElasticsuiteCore\Search\Request\Aggregation\Bucket\Histogram.
#0 /var/www/project/vendor/magento/framework/ObjectManager/Factory/Dynamic/Developer.php(82): Magento\Framework\ObjectManager\Factory\Dynamic\Developer->_resolveArguments('Smile\\Elasticsu...', Array, Array)
#1 /var/www/project/vendor/magento/framework/ObjectManager/ObjectManager.php(57): Magento\Framework\ObjectManager\Factory\Dynamic\Developer->create('Smile\\Elasticsu...', Array)
#2 /var/www/project/var/generation/Smile/ElasticsuiteCore/Search/Request/Aggregation/Bucket/HistogramFactory.php(43): Magento\Framework\ObjectManager\ObjectManager->create('Smile\\Elasticsu...', Array)
#3 /var/www/project/vendor/smile/elasticsuite/src/module-elasticsuite-core/Search/Request/Aggregation/AggregationFactory.php(58): Smile\ElasticsuiteCore\Search\Request\Aggregation\Bucket\HistogramFactory->create(Array)
#4 /var/www/project/vendor/smile/elasticsuite/src/module-elasticsuite-core/Search/Request/Aggregation/AggregationBuilder.php(91): Smile\ElasticsuiteCore\Search\Request\Aggregation\AggregationFactory->create('histogramBucket', Array)
#5 /var/www/project/vendor/smile/elasticsuite/src/module-elasticsuite-core/Search/Request/Builder.php(157): Smile\ElasticsuiteCore\Search\Request\Aggregation\AggregationBuilder->buildAggregations(Object(Smile\ElasticsuiteCore\Search\Request\ContainerConfiguration),
```

I finaly found that setting special_price attribut filterable is not managed in the mapping.

Best Regards

romainruaud:
Hello,

we will not "fix" this one since there is no need to set the special_price attribute to filterable : the price displayed by the price slider in front-office already takes the special_price value into account (since we are using the price index table values).

[...]

PR #390, opened 2017-04-11
Allow to index prices attributes.

romainruaud:
They were not indexed previously, which was preventing our fellow users to create custom "prices" attributes and setting them as filterable in front office.

This has been discussed on #341 : this issue occurs in fact not for special_price only, but for any attribute having a price backend.

Regards


dersam:
I've tested this PR against my data (as discussed in #341). It fixes my issue with `price` types. Thanks for the quick response!

PR #396, opened 2017-04-19
Change prepareProductCollection scope of autocomplete DataProvider 

romainruaud:
Change the scope to public : this will allow people to add plugins on it.

This is mandatory for being able to filter products based on offers for the ElasticSuite for Retailers module.


afoucret:
Replaced by #404

PR #400, opened 2017-04-21
Adding Magento's EQP to build

romainruaud:
This PR embeds the Magento's marketplace EQP check (https://github.com/magento/marketplace-eqp) to the build.

It also contains two fixes to perform the build without error with a severity level of 10 (the minimum level to be published on the marketplace).

We could consider fixing all the other warnings shown for lower severity level later. Let me know.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ cache:
     - $HOME/.composer/cache
 
 script:
+    - vendor/bin/phpcs --config-set installed_paths vendor/magento/marketplace-eqp/
+    - vendor/bin/phpcs --standard=vendor/magento/marketplace-eqp/MEQP2 --severity=10 src/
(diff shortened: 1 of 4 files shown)


romainruaud:
Well that's kinda strange, Travis does not seem to be able to install magento/marketplace-eqp ...

But this worked perfectly on my personal build : https://travis-ci.org/romainruaud/elasticsuite/jobs/224354367

Maybe the Magento repo is down for now... I will take a look on this one later.

afoucret:
You can not have a successful build when changing the .travis.yml file.

This is a security policy of Travis to avoid disclosure of secrets (Magento authentication can be stolen if you append `echo $MAGENTO_USERNAME` or `echo $MAGENTO_PASSWORD` during the build). 

I will merge this one.

PR #407, opened 2017-04-25
Fixing autocomplete refactoring.

romainruaud:
Your refactoring was causing a fatal error because of using a non existent variable.


romainruaud:
I added one more commit : the argument name to inject the proper collection on the DI was not the same as used in the object.

romainruaud:
I added another one : the attributes provider was not using the ProductCollection provider, so the facets for attributes were not added to the collection and therefore not displayed.

PR #417, opened 2017-05-04
Feature optimizer preview

romainruaud:
Hello @afoucret ,

let me know about this one.

It was quite hard to inject conditionnally the optimizers because of the existing plugin who does it on a very low-level (on the QueryBuilder).

Maybe we could consider moving this plugin elsewhere which could allow me to write smaller piece of code for this feature.

Let me know


afoucret:
Merge have been reverted because UI is not polished enough.

Issue #409, opened 2017-04-26
Question: How to translate "Search terms" in autocomplete?


shazul:
Magento Version : CE 2.1.6
ElasticSuite Version : 2.3.0
Environment : Developer and Production

This is actually a question, and probably not a bug:
I've been unsuccesfully trying to translate the block titles in the autocomplete results: **Search terms, Products, Categories**.

I have tried adding them to my theme csv translations file (dictionary.csv) which is where I think they should be. Just to try I also addes them to the js translation (No success). Please excuse me if I'm just doing something wrong, I just can't figure out how to translate those "words".

romainruaud:
Hello,

I can confirm this one is a bug, I am not able to get the translations working properly in my testing environment either.

We will try to provide a fix soon. 

Best regards

romainruaud:
The ```translate="true"``` node was missing into the DI for the "Search terms" title.

But the other have this tag and does not get translated properly either.

I think that the ```ŧranslate="true"``` into Magento DI has no effect : they are not using in the core.

I asked Magento how this is supposed to work : https://github.com/magento/magento2/issues/9411

But I guess we will have to refactor slightly this part. 

We keep you in touch

afoucret:
PR merged

PR #437, opened 2017-06-01
Fixing the way the translations of the autocomplete  are built.

romainruaud:
This is a fix for #409


romainruaud:
I updated the PR : injecting the block's titles via layout instead of DI allow us to have them properly translated.

So imho this is better now.

PR #454, opened 2017-06-20
Fixed issue with "is one of" rules not working.

oliverde8:
is One of rules are of input type string, but should still use a TERMS query.


oliverde8:
Thanks,

Issue #450, opened 2017-06-12
Can we add custom sort ordering for product attribute option values ?


vipingarg12:
Hi Support,

We need to add custom sort ordering in results. How we can do that ? Currently sorting working on attribute option id.

Magento version: 2.1.1
Elastic Version: 2.1.1
Elastic Suite Version: 1.2.0

Thanks,
Vipin

romainruaud:
Hello @vipingarg12 

You just have to set any attribute you want to sort on, as "Used For Sorting in Product Listing".

![selection_315](https://user-images.githubusercontent.com/15340849/27033775-1895c858-4f7b-11e7-9427-7e16431a4a1a.png)


This is standard Magento's behavior and we are actually keeping it functional.

Best regards. I close the issue, do not hesitate to reopen it if you have any other question.

vipingarg12:
We need to sort ordering on option values. for example Color is attribute and values are Red, Green, Blue then when we sort it should sort by alphabetically.  So Blue, Green, Red results should come.

I hope you understand the problem. Default magento sort gives results based on option id not alphabetically.

romainruaud:
Sorry but I do not get it.

What do you want to sort ? The options inside the filter ? Or the products ?

vipingarg12:
I wants to sort products collection on category page. Go to below page and sort by color

http://custom.pratikshajewelry.com/diamond.html

romainruaud:
Ok I got it.

You are right, when using a select-based attribute for sort, the optionId is used, instead of the option label.

I reopen the issue, we will try to provide a fix for this one.

Regards

mautz-et-tong:
We have the same Problem on Integer Attributes.
We have set up an attribute "alt_position" type integer and the frontend is sorted by value_id not by the value itself.
We use Magento 2.1.7
Smile 2.3.5
Elastic 2.4.5

afoucret:
PR have been merged into branch 2.3.x

andros89cy:
Hello vipingarg12,

Are you a magento developer?

Thanks

jeremy-donson:
**  Has this customizable sorting issue been addressed?  **
I have great ideas, and want great options!  :D 
I can fix it solo, but would prefer to have this be a github enhancement for our planet.

vishal-magento360:
@vipingarg12  did you get any solution for this ? I am also experiencing same issue.

manishgoswamig:
[...]

PR #463, opened 2017-07-05
Ensure sort order is set on option text field for attributes using so…

romainruaud:
…urce.

This is a fix for #450 

See my comment explaining the problem : https://github.com/Smile-SA/elasticsuite/issues/450#issuecomment-307786591

Imho, this fix is sufficient for 2.3.x branch.

But we should consider refactoring a bit the way we handle field name mapping in the collection for the next 2.4.x release. What I have in mind would be a FieldNameMapperInterface or something like this, which could handle these operations, and we could inject into the collection but also everywhere else we need to access the mapping (virtual categories, search API, and so on...) instead of duplicating a dirty "mapFieldName" function everywhere.

Let me know,

### DIFF
--- a/src/module-elasticsuite-catalog/Model/ResourceModel/Product/Fulltext/Collection.php
+++ b/src/module-elasticsuite-catalog/Model/ResourceModel/Product/Fulltext/Collection.php
@@ -99,11 +99,6 @@ class Collection extends \Magento\Catalog\Model\ResourceModel\Product\Collection
     private $originalPageSize = false;
 
     /**
-     * @var \Smile\ElasticsuiteCore\Helper\Mapping
-     */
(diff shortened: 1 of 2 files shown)


romainruaud:
@afoucret any new about this one ?

romainruaud:
PR Updated : 

- inject proper field name for attributes directly via addSortFilterParameter

- little change to prepareSortOrder function : 

previously, the $sortField was erased because it was recomputed from $attribute even if found in _productLimitationFilters. 

This was leading to the following issue : first "option_text_manufacturer" was properly pulled from _productLimitationFilters, but then overwritten by mapFieldName($attribute) where $attribute is the "real" attribute code without the "option_text_" prefix

PR #475, opened 2017-07-23
Add Magento_LayeredNavigationStaging as a dependency to fix blank filter 

tunght13488:
With certain config.php module sequence where Magento_LayeredNavigationStaging come before Smile_ElasticsuiteCatalog we got blank filter for product listing page and search result page


afoucret:
Merging this PR would made the module incompatible with Magento CE (the module `Magento_LayeredNavigationStaging` is not available for Magento CE). 

I don't see any way to fix module loading order while keeping the dependency optional.
We will need to dig in this one.

BR,

Issue #412, opened 2017-05-03
Unable to search by SKU


robbanl:
I've tried search using SKU with the 2.4.x branch but it's still a no go. No results turn up. From what I've seen there have been other tickets that says they are solved but then commented later on that they do not work.

Magento Version : CE2.1.4
ElasticSuite Version : 2
Environment : both

southerncomputer:
https://github.com/wardcapp/elasticsuite/commit/2ac9f8e0bf9f05eb038f7e748496072dcf186abd

apply that patch to 2.4.x branch and reindex

robbanl:
I've tried that but it still won't work. I can see the SKU in the ES index when doing a manual check :(

southerncomputer:
Works for me i've got 1.5M sku's and it works a-okay!

romainruaud:
Hello @robbanl 

The solution given by @southerncomputer may be dependant of the way your SKUs are "built" ? Are they fully numeric or a mix from letter and numbers ?

Regards

indiebytes:
Hi @romainruaud,

I work with @robbanl and can tell that the SKUs consist of an mix of letters, numbers and some other characters, at least dashes and underscores.

Thanks for helping out!

southerncomputer:
It would be nice to allow SKU like search for meta keywords as well, since I end up putting the SKU,MFGPART#,UPC,EAN,GTIN,Competitor part # in that location as well.  Easy way to increase matching capacity without increasing too many attributes!

I will say the dash does cause problems since ABC-V1P80 will not match at all whilst V1P80 will match. I just copied romainruaud 's solution from is_reference module.

If you have suggestion on how to match PREFIX-MFGPARTNO with all sorts of garbage in the sku, please let us know! ie HEW-V1P80UTR#ABA which does not get matched very well!

robbanl:
I tried adding a new attribute (sku_search) and for one product I added the SKU (just numbers) and tried the search. But that didn't work either. Isn't it possible to have custom attributes in Smile ElasticSearch module? Or do I need to define them in some XML?

samc2810:
Is there any solution to this?
The solution of @southerncomputer doesnt seem to work for V2.3.5

romainruaud:
@samc2810 please give me some examples of SKUs you are using, I will take a look at this one.

Best regards,

romainruaud:
@indiebytes and @robbanl please provide me some examples of SKU too

robbanl:
[...]

PR #479, opened 2017-07-26
2.4.x - search by sku corrections

ericrisler:
This commit applies the patch and updates created by @southerncomputer in this Issue: https://github.com/Smile-SA/elasticsuite/issues/412.

Also found a bug in the setup:upgrade processes when InstallData for the virtual categories executed.


src/module-elasticsuite-virtual-category/Setup/InstallData.php
```diff
@@ -45,11 +51,13 @@ class InstallData implements InstallDataInterface
...
      * @param EavSetupFactory             $eavSetupFactory             Eav setup factory.
      * @param VirtualCategorySetupFactory $virtualCategorySetupFactory Virtual Category setup factory.
+     * @param Config                  $eavConfig       EAV Config.
      */
-    public function __construct(EavSetupFactory $eavSetupFactory, VirtualCategorySetupFactory $virtualCategorySetupFactory)
+    public function __construct(EavSetupFactory $eavSetupFactory, VirtualCategorySetupFactory $virtualCategorySetupFactory, Config $eavConfig)
     {
         $this->eavSetupFactory      = $eavSetupFactory;
         $this->virtualCategorySetup = $virtualCategorySetupFactory->create();
+        $this->eavConfig       = $eavConfig;
```
Koc: This property seems unused

houdang:
Any news on this merge?

houdang:
@ericrisler Any plan when the build will be fixed?

madonzy:
@Koc @ericrisler @southerncomputer does this pull request fixes error with partial searching of SKU? I mean if I have SKUs like RM-123, RM-32, RM-5345 and I'll type in the search field: "RM-" and find all products where SKU starts from "RM-" ?

southerncomputer:
What I did eventually - was just throw the sku's into meta_keywords and make it searchable, but when importing the sku's i would break them up into their component pieces. I believe the problem with this solution is that the Analyzer_REFERENCE should break the sku into pieces and then the normal search should search against that field, but I don't believe the design of the indexer supports this method of breaking the sku into pieces then using the default search against that field

ericrisler:
@madonzy - I don't think that this pull works on partial matching sku's with hypens. I just did a quick test on a site we use this for and it gave no results for a partial match. However, the autocomplete did pickup the correct partial matches. 

@southerncomputer - I didn't notice you gave feedback on improving the patch. I'll take a look at those those soon. Can you summarize what will the suggestions you make improve upon? Thanks.

southerncomputer:
@ericrisler Since I could not figure out the edge n gram against SKU field, i just break the sku into chunks and throw it into searchable field meta_keywords. This allows the full-text search to work against the chunks of sku - basically what the EDGE NGRAM filter was trying to do to SKU field -
so sku HEW-ABCD1234 is broken on import to meta_keywords as:
hew-A,HEW-AB,HEW-ABC,HEW-ABCD,HEW-ABCD1,HEW-ABCD12,HEW-ABCD123, HEW-ABCD1234 plus the same for MFGPARTNO! It did not require all these modifications for search by sku and my import routines handle it with ease!

afoucret:
Time for me to have an update on this one.

Search by SKUs is considered as fixed in branch 2.4.x of ElasticSuite :

* You can now find a product whose SKU is "AA-999df" even if you type "aa999 df". That a pretty big update.

* Search by prefix exists even if is not complete.
With the previous example the product can be found if you type "aa" or "aa999" not if you type "aa99". We don't support ngram and we will not in the future as explained bellow.

* Search by suffix is not supported out of the box and you will not find a product whose SKU is 001883 with the search 1883. It can be fixed with custom analyzer on the SKU field. In the B2C space it is a bad idea to enable search by suffix (exemple of 007 becoming 7) so we will not add it out-of-the box.

* We do not use ngram and are not huge fan of it. You can do it by yourself with custom analyzer but it will not be part of ElasticSuite. Here are some reasons :

    * Indexing performances are really bad when using ngram (and wasted space is too important).

    * It would undermine precision of the search as long we are not able to make the difference between search and index analyzer

* In version 2.5.x, I plan to introduce an new architecture for the autocomplete. It could allow a better product suggestion when typing SKUs.

afoucret:
The proposed implementation is very different of what we plan to do with search by SKUs.
As a result I close this PR.

PR #482, opened 2017-07-31
Search exact term but still get "No exact results found for: '<query>'. The displayed items are the closest matches."

tunght13488:
If doc_freq doesn't exist in term stats, the default value should be 1, because min_doc_freq is default to 1 if omitted in the query param.
See https://www.elastic.co/guide/en/elasticsearch/reference/2.4/docs-termvectors.html#_terms_filtering
for more detail


afoucret:
Hi @tunght13488 ,

0 means that the term is not present into the index. It is exactly what we want to do.

For your issue you should have a look at #490

Issue #465, opened 2017-07-05
Changing category to virtual not triggers reindex


Koc:
### Preconditions

Magento Version : 2.1.7

ElasticSuite Version : 2.3.6

Environment : PHP 7.0.x

Third party modules :

### Steps to reproduce

1. Change some category to virtual
2. Apply some rules
3.

### Expected result

1. Some indexes marked as required reindex and will update by cron

### Actual result

1. no updates, all indexes are green, should run reindex manually by cli

not sure, maybe happens not only on mode switching but also when rules changed.

romainruaud:
Hello,

I will check this one.

For me : 

- when switching a category from "standard mode" to "virtual category" : we should invalidate the index, to ensure products previously assigned to this category are correctly set out of the category.

- when creating a virtual category, or changing a previous rule : there is no need to invalidate the index, since nothing is indexed for virtual categories : The query is processed in real time when navigating to the category page, which allow the category to evolve by itself when products matching the rule are changing (Eg. a category set to "Sale : Yes" will automatically display products that have just been set as "Sale : Yes").

Regards

Koc:
Agreed with both. Thanx for fast response

romainruaud:
PR merged.

Please use 2.3.x branch or wait for the next minor release.

Regards

PR #488, opened 2017-08-07
Enforce reindexing when category is set to virtual.

romainruaud:
This is a fix for #465 

Enforce reindexing (if set to Update on Save) when switching a category from normal mode to virtual mode.

Nothing done for mview, since this is actually working properly (due to the SQL trigger).

### DIFF
--- a/src/module-elasticsuite-virtual-category/Plugin/Catalog/Category/ReindexOnChange.php
+++ b/src/module-elasticsuite-virtual-category/Plugin/Catalog/Category/ReindexOnChange.php
@@ -46,10 +46,13 @@ class ReindexOnChange
     {
         $proceed();
 
-        if ($category->dataHasChangedFor('is_virtual_category')) {
+        // Reindex only if attached product list has changed.
+        // This prevent reindexing if the category is just created and set to virtual.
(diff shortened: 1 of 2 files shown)


afoucret:
It seems there is a problem with this line :

```
$this->getIndexer()->reindexList($category->getPathIds());
```

`$category->getPathIds()`returns category ids and you reindex it using the product indexer. IMHO you don't reindex the products as expected.

romainruaud:
Hmmm, this is the "catalog_category_product" indexer, and it's taking category ids for reindex : https://github.com/magento/magento2/blob/develop/app/code/Magento/Catalog/Model/Indexer/Category/Product/Action/Rows.php

This is done same way by Magento on standard category model (but only if there are products attached, which is not good in our case) : https://github.com/magento/magento2/blob/develop/app/code/Magento/Catalog/Model/Category.php#L1097

This is indexing properly the products in the catalog_category_product_index table (removing them, exactly), and then the fulltext indexer does the job and removes the "category" field from documents.

romainruaud:
PR Updated : 

- prevent reindexing of root category

- prevent reindexing if no products were attached previously (this manage the case of a virtual category creation, since its done in 2 steps).

PR #490, opened 2017-08-08
Fix term vector usage to use also default search field.

romainruaud:
Potential fix for #412 

Two points here : 

- The detection of spelling type (exact or fuzzy) was based on the "spelling" analyzer. It cannot work for fields which are not used for spellcheck.

=> For this one, I have a question, why using "spelling" analyzer for term vectors ? Changing it for both 'search' and 'spelling' gives me better results for fields which are not used for spellchecking.

- The term vector query did not use "dfs" : this implies it is ran to a random shard. By doing this way, a query can be exact one time (matches document in the randomized shard) but fuzzy next time (not matching document in the new randomized shard).

=> About this one, I can understand this may be due to performances issues when using 'dfs'. Let me know where the cursor should be put between exact spelling calculation or less response time. (But the term vector result is cached for later usage, so I would say that using 'dfs' is fine and will give better relevancy).

Let me know !

### DIFF
--- a/src/module-elasticsuite-core/Search/Adapter/Elasticsuite/Spellchecker.php
+++ b/src/module-elasticsuite-core/Search/Adapter/Elasticsuite/Spellchecker.php
@@ -141,9 +141,13 @@ class Spellchecker implements SpellcheckerInterface
             'type'            => $request->getType(),
             'id'              => '',
             'term_statistics' => true,
+            'dfs'             => true,
         ];
 
-        $termVectorsQuery['body']['doc'] = [MappingInterface::DEFAULT_SPELLING_FIELD => $request->getQueryText()];
+        $termVectorsQuery['body']['doc'] = [
+            MappingInterface::DEFAULT_SEARCH_FIELD   => $request->getQueryText(),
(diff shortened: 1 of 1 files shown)


afoucret:
* For the first point, it is by design. We try to improve precision by using only some fields during the spellechecking process (for both detection an fuzziness). The precision is better since some field that may contain mispelling (eg : description) or irrelevant data are excluded from the spellchecker. You can always append more attributes as "Used in spellchecker" if you want.

* I would consider the dfs patch since it seems relevant.

romainruaud:
I understand the first point, but it would imply to set the "sku" field as "Used in spellchecker" if we want to have relevant search results (exact matches) for it, which is imho a bit "strange" for the users. This is true for any other fields.

But since I know this is necessary for it to work, I can add this by default for the SKU field (this would ensure proper search on this field and allow closing the related issue), and I'll be able to provide support for such questions to our fellow users.

I can update the PR to contain only the dfs patch if you want, let me know

afoucret:
Used in spellchecker on the sku field would undermine spellchecker precision IMHO.
Please, update the PR for the dfs stuff.

romainruaud:
PR updated

PR #538, opened 2017-09-21
Fixed sort order for attributes with manual sort

sky-hub:
Issue: when adding "0" value option to a multiselect attribute the "$optionLabel" will be considered empty and "adminSortIndex" will not be set

Changes: 
1. Select attribute options without empty value as empty value is not used in layered navigation
2. Trim option label as options may contain spaces. In this case facet value is already trimmed and attribute option will not causing undefined index for "adminSortIndex" while usort().
3. Check if "$optionLabel" is "0" string as attribute options may contain "0" value


afoucret:
Replaced by #564

PR #545, opened 2017-09-29
Fix PHPUnit test for Magento 2.2.x

afoucret:


afoucret:
Forced merge (travis.yml have been modified and it always provoke a build failure)

Issue #524, opened 2017-09-05
Numeric sku not found


rikwillems:
We have some products that have a numeric sku (123456). These products are not found through search when querying for that sku. Products that have letters in the sku (abc123) are found when searching for the sku.

IPeredero:
https://github.com/Smile-SA/elasticsuite/issues/412

rikwillems:
Hi @IPeredero thanks for pointing me here. I must admit I have too little knowledge of Elasticsearch to make out what I need to do with all the output shown.

southerncomputer:
One option that you can do is place the ski into meta_keywords and make it searchable but break it down into its component - ie SKU 1122334455 as 1122334455,112233445,11223344,1122334,112233,11223,1122,112,11

Then make the field searchable - should do the trick, it is essentially what the reference method was doing automatically but didn't work out so hot! 

Try it!

romainruaud:
I close it since it's a duplicate of #412 which is the main topic about SKU based problematics.

Feel free to test also the attached PR #548 and tell me if it's better.

Regards

PR #548, opened 2017-09-29
Compute spellcheck apart for fields not using default search analyzer.

romainruaud:
This one is meant to enhance "search by sku" but could theorically work with any field which is using a custom "defaultSearchAnalyzer". I choose to split the spellcheck request, which was previously performed only on "spelling" field, to "spelling" + "the fields which are using a custom search analyzer".

This has been tested working OK with several patterns of SKU : 

- mix of letters and number : 101177BLG
- full letters : HEWPUTRABC
- full numbers : 11610018
- mixed with dash : ABC-V1P80
- overly complicated sku : HEW-V1P80UTR#ABA

So this should gladly address #412 and #524 

However, this actually does NOT work with "partial searches", eg searching for just "ABC" or "V1P80" or BLG will not match against my examples.

So let me know if you want me to enhance more this PR by adding support of partial searches, which seems a little bit more tricky to do.

Regards

### DIFF
--- a/src/module-elasticsuite-catalog-optimizer/Model/Optimizer/Preview/SearchQuery.php
+++ b/src/module-elasticsuite-catalog-optimizer/Model/Optimizer/Preview/SearchQuery.php
@@ -144,6 +144,7 @@ class SearchQuery
             'type'            => $containerConfig->getTypeName(),
             'queryText'       => $queryText,
             'cutoffFrequency' => $containerConfig->getRelevanceConfig()->getCutOffFrequency(),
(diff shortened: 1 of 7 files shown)


madonzy:
@romainruaud Yes, I think "partial searches" is very important for implementation especially for B2B commerce.

southerncomputer:
How can we use REFERENCE analyzer to break SKU into pieces but normal analyzer to search for sku results?

so sku is broken into pieces by reference (EDGE-NGRAm) correctly?

romainruaud:
Closed, in favor of #586 which is by far an easier solution.

Issue #362, opened 2017-03-21
Bad sort order for products with no value for the attribute used for sort


fabiendaubanes:
### Preconditions

Magento Version : 2.1.4

ElasticSuite Version : 2.3.2

Environment : developer

### Steps to reproduce

1. Create a custom product attribute "custom_attribute"
2. Enable "used_for_sort_by" on this custom attribute
3. Set a value for this custom attribute on some products (but not for all products) linked to a same category "MyCategory"
4. `php bin/magento indexer:reindex`
5. `php bin/magento cache:clean`
6. Go to the "MyCategory" view page, and sort the product listing by your "custom_attribute", and order by DESC

### Expected result

The process can't sort the products with no value for the "custom_attribute" => it's normal.
But I think it would be more logical if this products are displayed after products with value (and before for an ASC sorting)

### Actual result

The sorting works, but products with no value for the "custom_attribute" are displayed first, before products with value. 
See my listing with DESC sorting :

![actual_result](https://cloud.githubusercontent.com/assets/26573993/24153976/329642ee-0e50-11e7-9701-f065d0cc2998.png)

PR #580, opened 2017-10-23
Feature missing sort

### LINKED ISSUE #560: Option to change "missing" option in "sort" clause.
We needed to reverse the "missing" option behaviour of the "sort" clause within a products query but there is no per attribute option to achieve this.

Probably related to https://github.com/Smile-SA/elasticsuite/issues/62, but not sure, so we opened a new Issue.

### Preconditions

We have a date attribute we use to sort catalog products which is optional so many products are missing a value. When we sort by descendant order on this attribute we get products with null value first but they should show last.
Current solution is to override buildSortOrder inside `Smile\ElasticsuiteCore\Search\Adapter\Elasticsuite\Request\SortOrder\Builder`, changin this lines:
```
        $sortOrderConfig = [
            'order'         => $sortOrder->getDirection(),
            'missing'       => $sortOrder->getDirection() !== SortOrderInterface::SORT_ASC ? '_first' : '_last',
            'unmapped_type' => FieldInterface::FIELD_TYPE_STRING,
        ];
```
Please note that "**!==**" was originally a "==".

What we achieve was to show products with null attribute last, but it affected the catalog globally (on every attribute). What we preferred was to have a backend option to tune on a per attribute basis.

Magento Version : 2.1.7

ElasticSuite Version : 2.3.8

Environment : Developer and production mode.

### Steps to reproduce

1. Add an optional date catalog product attribute.
2. Set this value only on certain products.
3. Try to sort in DESCENDING order.

### Expected result

1. We expected to see products with attribute value first (obviously sorted in desc order) followed by products without a value on that attribute.

### Actual result

1. Products without a value for that attribute are shown first.

romainruaud:
This one allow to fix for each attribute (and both sorting direction) how we are supposed to set the "missing" parameter of the ES sort order.

Default values are equal to what was done previously (DESC => _last, ASC => _first) so it will be transparent after upgrade.

Will fix #560 and #362 by letting our fellow users to define how they want the sort order to behave.

### DIFF
--- a/src/module-elasticsuite-catalog/Block/Plugin/Adminhtml/Product/Attribute/Edit/Tab/FrontPlugin.php
+++ b/src/module-elasticsuite-catalog/Block/Plugin/Adminhtml/Product/Attribute/Edit/Tab/FrontPlugin.php
@@ -99,7 +99,7 @@ class FrontPlugin
         $this->addSearchFields($fieldset);
         $this->addAutocompleteFields($fieldset);
         $this->addFacetFields($fieldset);
-
+        $this->addSortFields($fieldset);
         $this->appendSliderDisplayRelatedFields($form, $subject);
 
         if ($this->getAttribute()->getAttributeCode() == 'name') {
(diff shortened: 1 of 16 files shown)


afoucret:
To be rebased

romainruaud:
Rebase done

afoucret:
UX is too complicated.
Having many parameters for sorting is not really easy to understand.
Think we should work on the missing value instead.

PR #582, opened 2017-10-24
Adding a special attribute to build rules on new products.

### LINKED ISSUE #410: [Suggestion] Add is_new nested field for Virtual Category rules
Hello,

I was wondering if it's possible to add a new field "is_new" for Virtual Categories so that it can be possible to create categories containing only new stuff. I think it will follow the same implementation as the two other nested fields except for the logic which will be more or less like this:
```php
	$now = date("Y-m-d");
	$newsFrom= substr($_product->getData('news_from_date'),0,10);
	$newsTo=  substr($_product->getData('news_to_date'),0,10);
	if ($newsTo != '' || $newsFrom != ''){
		if (($newsTo != '' && $newsFrom != '' && $now>=$newsFrom && $now<=$newsTo) || ($newsTo == '' && $now >=$newsFrom) || ($newsFrom == $
			$is_new=true;
		}
	}
```
PS: I can try and do a PR with the suggestion if that's permitted
### Preconditions

Magento Version : Magento CE 2.1.6

ElasticSuite Version : 2.4.x (dev-master)

Environment : Production

Third party modules : none relevant

romainruaud:
fix #410 

I deliberately choose to set ElasticsuiteCatalog version to 1.2.4 to not collision the other PR #580

### DIFF
--- a/src/module-elasticsuite-catalog-rule/Model/Rule/Condition/Product.php
+++ b/src/module-elasticsuite-catalog-rule/Model/Rule/Condition/Product.php
@@ -141,7 +141,7 @@ class Product extends \Magento\Rule\Model\Condition\Product\AbstractProduct
     public function getInputType()
     {
         $inputType = 'string';
-        $selectAttributes = ['attribute_set_id', 'stock.is_in_stock', 'has_image', 'price.is_discount'];
+        $selectAttributes = ['attribute_set_id'] + array_keys($this->getSpecialAttributes());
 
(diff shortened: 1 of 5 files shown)


romainruaud:
@rorua : Think about delivering an "extendable" rule system instead of this PR, to allow any user to customize the rules by adding their attributes with their own rules.

romainruaud:
closed in favor of #595

PR #595, opened 2017-10-30
Refactoring "special attributes" to allow extending them easily.

romainruaud:
Moving all hardcoded reference to "special attributes" (is_in_stock, price.is_discount, has_image) to a SpecialAttributeProvider extendable via the DI.

Here is also a proposal to implement a special attribute "is_new" injectable via DI : 

di.xml : 

```
    <type name="Smile\ElasticsuiteCatalogRule\Model\Rule\Condition\Product\SpecialAttributesProvider">
        <arguments>
            <argument name="attributes" xsi:type="array">
                <item name="is_product_new" xsi:type="object">Smile\ElasticsuiteCatalogRule\Model\Rule\Condition\Product\SpecialAttribute\IsNew</item>
            </argument>
        </arguments>
    </type>
```

"IsNew" class : 

```
<?php
/**
 * DISCLAIMER
 * Do not edit or add to this file if you wish to upgrade Smile Elastic Suite to newer
 * versions in the future.
 *
 * @category  Smile
 * @package   Smile\ElasticsuiteCatalogRule
 * @author    Romain Ruaud <romain.ruaud@smile.fr>
 * @copyright 2017 Smile
 * @license   Open Software License ("OSL") v. 3.0
 */
namespace Smile\ElasticsuiteCatalogRule\Model\Rule\Condition\Product\SpecialAttribute;

use Smile\ElasticsuiteCatalogRule\Api\Rule\Condition\Product\SpecialAttributeInterface;
use Smile\ElasticsuiteCore\Search\Request\Query\QueryFactory;
use Smile\ElasticsuiteCore\Search\Request\QueryInterface;

/**
 * "Is New" special attribute class
 *
 * @category Smile
 * @package  Smile\ElasticsuiteCatalogRule
 * @author   Romain Ruaud <romain.ruaud@smile.fr>
 */
class IsNew implements SpecialAttributeInterface
{
    /**
     * @var \Smile\ElasticsuiteCore\Search\Request\Query\QueryFactory
     */
    private $queryFactory;

    /**
     * @var \Magento\Config\Model\Config\Source\Yesno
     */
    private $booleanSource;

    /**
     * Image constructor.
     *
     * @param \Smile\ElasticsuiteCore\Search\Request\Query\QueryFactory $queryFactory  Query Factory
     * @param \Magento\Config\Model\Config\Source\Yesno                 $booleanSource Boolean Source
     */
    public function __construct(QueryFactory $queryFactory, \Magento\Config\Model\Config\Source\Yesno $booleanSource)
    {
        $this->queryFactory  = $queryFactory;
        $this->booleanSource = $booleanSource;
    }

    /**
     * {@inheritdoc}
     */
    public function getAttributeCode()
    {
        return 'is_product_new';
    }

    /**
     * {@inheritdoc}
     */
    public function getSearchQuery()
    {
        $now = (new \DateTime())->format(\Magento\Framework\Stdlib\DateTime::DATETIME_PHP_FORMAT);

        $clauses = [];

        $newFromDateEarlier = $this->queryFactory->create(
            QueryInterface::TYPE_RANGE,
            ['field' => 'news_from_date', 'bounds' => ['lte' => $now]]
        );

        $newsToDateLater = $this->queryFactory->create(
            QueryInterface::TYPE_RANGE,
            ['field' => 'news_to_date', 'bounds' => ['gte' => $now]]
        );

        $missingNewsFromDate = $this->queryFactory->create(QueryInterface::TYPE_MISSING, ['field' => 'news_from_date']);
        $missingNewsToDate   = $this->queryFactory->create(QueryInterface::TYPE_MISSING, ['field' => 'news_to_date']);

        // Product is new if "news_from_date" is earlier than now and he has no "news_to_date".
        $clauses[] = $this->queryFactory->create(
            QueryInterface::TYPE_BOOL,
            ['must' => [$newFromDateEarlier, $missingNewsToDate]]
        );

        // Product is new if "news_to_date" is later than now and he has no "news_from_date".
        $clauses[] = $this->queryFactory->create(
            QueryInterface::TYPE_BOOL,
            ['must' => [$missingNewsFromDate, $newsToDateLater]]
        );

        // Product is new if now is between "news_from_date" and "news_to_date".
        $clauses[] = $this->queryFactory->create(
            QueryInterface::TYPE_BOOL,
            ['must' => [$newFromDateEarlier, $newsToDateLater]]
        );

        // Product is new if one of previously built queries match.
        return $this->queryFactory->create(QueryInterface::TYPE_BOOL, ['should' => $clauses]);
    }

    /**
     * {@inheritdoc}
     */
    public function getOperatorName()
    {
        return ' ';
    }

    /**
     * {@inheritdoc}
     */
    public function getInputType()
    {
        return 'select';
    }

    /**
     * {@inheritdoc}
     */
    public function getValueElementType()
    {
        return 'hidden';
    }

    /**
     * {@inheritdoc}
     */
    public function getValueName()
    {
        return ' ';
    }

    /**
     * {@inheritdoc}
     */
    public function getValue()
    {
        return 1;
    }

    /**
     * {@inheritdoc}
     */
    public function getValueOptions()
    {
        return $this->booleanSource->toOptionArray();
    }

    /**
     * {@inheritdoc}
     */
    public function getLabel()
    {
        return __('Only new products');
    }
}

```

### DIFF
--- a/src/module-elasticsuite-catalog-rule/Api/Rule/Condition/Product/SpecialAttributeInterface.php
+++ b/src/module-elasticsuite-catalog-rule/Api/Rule/Condition/Product/SpecialAttributeInterface.php
@@ -1,0 +1,72 @@
+<?php
+/**
+ * DISCLAIMER :
+ *
+ * Do not edit or add to this file if you wish to upgrade Smile Elastic Suite to newer
+ * versions in the future.
+ *
+ * @category  Smile_Elasticsuite
+ * @package   Smile\ElasticsuiteCatalogRule
+ * @author    Aurelien FOUCRET <romain.ruaud@smile.fr>
+ * @copyright 2017 Smile
+ * @license   Open Software License ("OSL") v. 3.0
+ */
+
+namespace Smile\ElasticsuiteCatalogRule\Api\Rule\Condition\Product;
+
+use Smile\ElasticsuiteCore\Search\Request\QueryInterface;
+
+/**
+ * Definition of "special attributes" that can be used for building rules with Elasticsuite.
+ *
+ * @category Smile_Elasticsuite
+ * @package  Smile\ElasticsuiteCatalogRule
+ * @author   Aurelien FOUCRET <aurelien.foucret@smile.fr>
+ */
+interface SpecialAttributeInterface
+{
+    /**
+     * @return string
+     */
+    public function getAttributeCode();
+
+    /**
+     * @return QueryInterface
+     */
+    public function getSearchQuery();
+
+    /**
+     * @return string
+     */
+    public function getOperatorName();
+
+    /**
+     * @return string
+     */
+    public function getInputType();
+
+    /**
+     * @return string
+     */
+    public function getValueElementType();
+
+    /**
+     * @return string
+     */
+    public function getValueName();
+
+    /**
+     * @return mixed
+     */
+    public function getValue();
+
+    /**
+     * @return array
+     */
+    public function getValueOptions();
+
+    /**
+     * @return string
+     */
+    public function getLabel();
+}
--- a/src/module-elasticsuite-catalog-rule/Model/Rule/Condition/Product.php
+++ b/src/module-elasticsuite-catalog-rule/Model/Rule/Condition/Product.php
@@ -1,10 +1,8 @@
 <?php
 /**
  * DISCLAIMER
- *
  * Do not edit or add to this file if you wish to upgrade Smile Elastic Suite to newer
  * versions in the future.
- *
  *
  * @category  Smile
  * @package   Smile\ElasticsuiteCatalogRule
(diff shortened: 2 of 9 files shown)


romainruaud:
This is a better version of the fix proposed in #582

Koc:
@romainruaud  Is it possible do all calculations on indexation phase and filter by bool field? This could be performance optimization.

romainruaud:
@Koc I'm not sure it would be worth it, doing it during indexation would slow down the indexing process, whereas doing it on query allow to benefit from Elasticsearch query cache. Also, the queries are not overly complicated and should not take long for the engine to compute.

@afoucret any opinion about this ?

Koc:
For the most ecommerce projects reading from index doing more ofthen than reindexing IMHO.

jsterling:
@romainruaud I'm using this code successfully in a custom module - however only the first clause (Product is new if "news_from_date" is earlier than now and he has no "news_to_date".) has any effect. 

Due to this, products which have a 'news_from_date' in the past are still visible. Can you help?

gjportegies:
I've implemented this solution but I'm not getting any results on frontend and backend when creating a Virtual Category with the 'Only new products' rule. Is this solution should still relevant? I'm currently using ElasticSuite 2.8.0 on Magento 2.3.1

jesse-thomas:
I had this working 6-9 months ago. After updating Magento and other modules in my development environment, I just noticed that I do not get new product results anymore. I get a no products found message.

jesse-thomas:
@romainruaud can you update this to include the proper way to add a product attribute to the ES index? In this case we don't want the attribute to be searchable or sortable, just available for the virtual category query. Thanks!

romainruaud:
@jesse-thomas just set your attribute to "is used for promo rules" and it will be indexed. It's a bit hacky actually but most of time it's enough on our projects.

regards

JoostWan:
@romainruaud Thanks for this code. I'm busy to get this working for a client. But it's match the dates wrong. I get old products with a different new date in the category. 

When checked the data in Kibana a see the fields as a array instead of a string. I change for both attributes option _Use for Promo Rule Condition_ to **yes**. 

```
"news_from_date": [
      "2020-06-20 00:00:00",
      "2027-06-20 00:00:00"
    ],
    "news_to_date": [
      "2020-07-20 00:00:00",
      "2027-07-20 00:00:00"
    ],
```

Screenshot 2020-11-20 at 15 18 55
Screenshot 2020-11-20 at 15 18 25

Mbertu:
@JoostWan I think that are all data from child products (one of your child products has the year setted to 2027)

krunal2212:
How to add Date Range field for this Filter ? Is that possible ?

Issue #587, opened 2017-10-26
Manage facets by category.


romainruaud:
New feature : ability to manage facets by category.

For each category, display a list of all filterable attributes (not filterable in search for now).

The list contains the following data for each attribute : 

- **Sort Order** (implicit, the list is drag-n-droppable)
- **Attribute Name** (just the label)
- **A selector** to choose between 3 display mode. It contains 3 values : 
    - **Auto** : let the engine do the job, via the "facet coverage"
    - **Yes** : the attribute is displayed in any case, even if facet coverage is not reached.
    - **No** : the attribute is never displayed for this category.

afoucret:
Hi @romainruaud 

A prototype of the UI I want in : https://github.com/afoucret/elasticsuite/tree/facet-category-config

This code need a lot of cleaning and a proper management of the drag and drop and additional fields (coverage rate, sort order and facet size).

I let you finish the work based on this.

romainruaud:
acknowledged @afoucret 

What about hide/show of the attributes considered as irrelevant ? This was nice for UX especially with a huge diversity of universes in the catalog.

afoucret:
Start with implementing without this feature. It would be added after.

afoucret:
Feature is now in the master branch and will be released with version 2.5.0

PR #602, opened 2017-11-06
Adding filterable attributes list by category.

romainruaud:
This is my proposal for #587 

It's actually working but I wanna have your opinion before going eventually further.

I have listed points that I could add to this PR before merging it if you are ok : 

- Staging is not managed. Maybe it is not worth it to allow to stage such settings ? Let me know.

- Actually the data is not inheritable. I think it could be a cool feature to be able to inherit from parent category : probably customers would only configure filterable attributes for top level categories and let the children inherit.

- This has not been plugged into Search results (when filtering on a category). We could do it but I'm not sure how we should manage attributes that are only filterable in search.


afoucret:
Does look at the implementation detail but I see some usability optimizations :

* Design more consistent with color swatches (white header, more readable stripped display, more compact rows)

* Remove facet that will never be displayed because no products are matching. At least display a warning (No products : the facet will not be displayed)

afoucret:
Display is better now.

You still have to rework : 

* the display of facet that are not currently available

* refactoring the Navigation block as I explained to you yesterday

afoucret:
Problem when creating a new category : 
```
Fatal error: Method Magento\Ui\TemplateEngine\Xhtml\Result::__toString() must not throw an exception, caught Error: Call to a member function getFilterableAttributeList() on null in /Users/afoucret/local-apps/www/magento2-ee/vendor/magento/module-ui/Component/Wrapper/UiComponent.php on line 0
```

romainruaud:
Ok @afoucret I think I'm nearly done.

One point remains : how should we deal with newly created categories ? With the "irrelevant attributes" filter, it's actually showing an empty table which is not really good. I'm thinking about simply removing the fieldset when creating a new category (as we do with virtual category rules).

Any opinion ?

romainruaud:
I'm done @afoucret 

Ready for your final code review. 

Let me know !

afoucret:
Please rebase and I will review it.

romainruaud:
rebase done !

afoucret:
Close since you have started refactoring of this PR

PR #627, opened 2017-11-22
Improve input checkbox

minhluan259:


romainruaud:
I understand this is aimed to produce cleaner html by setting an `id` to the option value and a proper `for` to the associated label.

But imho there is no need to fetch the option_id from the database when you could just use a dynamically generated id.

This one could be done in full javascript for me.

minhluan259:
Yes, It could be done by javascript with dynamically generated id.
I'm assume it can use for Ajax Layered Navigation to get value from input field.

I've a question. Do you have a plan to make Ajax Layered Navigation?
Thank you.

afoucret:
Your PR has been replaced by PR #662 

Option ids are generated in JS. It is a better option if we want to build facet on attribute without source model (they don't have option ids).

PR #631, opened 2017-11-27
Extension service contracts (Product Datasource)

vo1:
Adds service contract of data extension with dummy product data service contract.
This service contract can be used for storing customized extra information in ES document source,
for example product configuration (configuration matrix - something like Product.Config in M1), so Magento 2 product listing could be fully loaded from ES database instead of joining extra stuff in Magento collections.
The service contract only manipulates extended_data field in ES, and will not require any overrides of Smile indexer to work, as it is right now.

It is built in 2.3.x branch (latest tag of 2.3.10), but should not be hard to also build it for 2.4.x


vo1:
if needed, I can also add Product.Config data array generation as an example of how it can be used.

afoucret:
Hi @vo1,

I don't see what the advantage of the proposed implementation compared to the existing one.

The ability to append additional data (and control the mapping) already exists in the module. 

As an example you can have a look at this extension built to bring B2B support to ElasticSuite : https://github.com/Smile-SA/magento2-module-elasticsuite-shared-catalog :

1. The mapping is modified through : https://github.com/Smile-SA/magento2-module-elasticsuite-shared-catalog/blob/master/etc/elasticsuite_indices.xml file to append the `customer_group_id` in the catalog_product_entity index.

2. The `\Smile\ElasticsuiteSharedCatalog\Model\Product\Indexer\Fulltext\Datasource\SharedCatalog`is configured as an additional datasources and append additional data.

Since `elasticsuite_indices.xml` are merged you can modify default behavior of the ElasticSuite module (battery are included but removable).

Since mapping is extensible and datasources can be append / overidden, I don't see clearly the advantage of your PR. What have I missed ?

vo1:
For a project I am working on, I have to build up extra field from multiple datasources: attribute, inventory and price. This service contract allows you depending on values received from all product datasources to generate your own record such as product configuration, e.g.:

```
{
  configuration: {
    "21": {
      "77": 1776,
      "78": 1787
    },
    "22": {
       "77": 1777,
       "78": 1778
    },
    "prices": {
      "1776": 220,
      "1777": 219,
      "1778": 219,
      "1787": 220
    },
    "salable": {
      "1776": 1,
      "1777": 1,
      "1778": 1,
      "1787": 0
    },
    "images": {
      "1776": "/f/i/color1image.png",
      "1777": "/f/i/color2image.png",
      "1778": "/f/i/color2image.png",
      "1787": "/f/i/color1image.png"
    }
}
```
This JSON is built from the same datasources that were used in process. When implementing own datasource you will have to run same amount of queries again, while using service contract you can pass all the datasources through it to generate extra stuff for your document.
The idea of having this JSON in ES document is to get rid of using collection in magento category/search page. So the whole configurable product block can be generated by javascript using the predefined template from JSON values received from ES.
Using such JSONConfig object will allow you to see price, image and availability of a child product when clicking different configurations in product block.

If I chose not the best approach for building this functionality, please let me know.
PS: Initially I made this as an override to attributeDataSource but the problem is to make it working I had to copy paste whole class, because addChildData is a private method, and most of the stuff there private, so can't use plugin.

afoucret:
2 points here :

1. I think the same result can be achieved using datasources at indexing. You can retrieve document source when the collection is loaded through `$product->getDocumentSource()`.

2. The document structure you proposed will generate a lot of new fields in the engine. In you example you will add the following fields in the mapping : 
* `configuration.21.77`
* `configuration.21.78`
* `configuration.22.77`
* `configuration.22.78`
* `prices. 1776`
* `prices. 1777`
* ....
 
If you index a full catalog it can result in thoudands new fields. Even if ElasticSearch can handle it, creating so many fields can hurt performance. 

So I will not merge this one.

PR #635, opened 2017-11-28
Fix JS error when hitting enter key before autocomplete is loaded.

### LINKED ISSUE #634: JavaScript error when hiting enter before autocomplete loads
### Preconditions

Magento Version: 2.2.0 but all are probably affected.

ElasticSuite Version: 2.3.0 but all versions are probably affected.

Environment: All modes are affected.

### Steps to reproduce

1. Go to any shop with elasticsuite search extension installed e.g. on the demo: http://demo.magento-elastic-suite.io/. 
2. Type any search string into mini search form.
3. Quickly hit enter key before autosuggestion items appear.

### Expected result

1. User should be redirected to search results page for given query string.

### Actual result

1. `TypeError: this.responseList.selected is null` or similar depending on browser is thrown in the console originating from this line: https://github.com/Smile-SA/elasticsuite/blob/master/src/module-elasticsuite-core/view/frontend/web/js/form-mini.js#L359

Above error is due to `this.responseList.selected` being null until autosuggestion results are loaded. The solution is to check if `this.responseList.selected` is truthy before accessing `.attr('href')` method.

I will create a pull request with fix for this in the nearest hours.

krzksz:
Fixes #634.

### DIFF
--- a/src/module-elasticsuite-core/view/frontend/web/js/form-mini.js
+++ b/src/module-elasticsuite-core/view/frontend/web/js/form-mini.js
@@ -356,8 +356,9 @@ define([
          * @private
          */
         _validateElement: function(event) {
-            if (this.responseList.selected.attr('href') !== undefined) {
-                window.location = this.responseList.selected.attr('href');
+            var selected = this.responseList.selected;
+            if (selected && selected.attr('href') !== undefined) {
+                window.location = selected.attr('href');
                 event.preventDefault();
                 return false;
             }


romainruaud:
@afoucret this one is OK for me to be merged. Maybe we'd better merge it on 2.4.x instead of master and then cascade.

Regards

afoucret:
PR updated to target 2.4.x branch.

PR #638, opened 2017-11-29
Fixing class reference issue on Magento EE.

### LINKED ISSUE #637: Autoload Error on Magento EE
Impossible to compile in production mode, or to access catalog pages in dev mode.

The following error happens : 

Cannot use Magento\Catalog\Model\ProductFactory as ProductFactory because the name is already in use in /var/www/magento/src/vendor/smile/elasticsuite/src/module-elasticsuite-catalog-rule/Model/Rule/Condition/Product.php on line 17

This is due to this commit : https://github.com/Smile-SA/elasticsuite/commit/8fdd30f73f3656b86e3cea227e153b590a35fc4a

I did rationalise the Class names and this is now causing an error. For unknown reasons this is happening only in Magento2 EE version, I guess this is due to a module of the EE suite which is changing the include order of files.

I will push a PR fast to fix this one.

romainruaud:
Fix #637


afoucret:
Merged in 2.4.x branch.

Issue #639, opened 2017-11-29
Some bundle products not being indexed when sharing same child


sambolek:
When the same child product is associated to multiple bundle products, only one of the bundle products is saved in ES. This causes bundle products to not appear in categories / search if they share the same child product.

### Preconditions

Not relevant

Magento Version : Magento 2.2.1

ElasticSuite Version : 2.4.x-dev @ 55becb9146fb7b78370c27f9ab00c074683a60f3

Environment : Either

Third party modules : Not relevant

### Steps to reproduce
1. Create category "Test"
2. Create simple products "Cell phone", "Cell phone 2" and "Cable" that are in stock, enabled etc
3. Create one bundle product "Cell pack 1" containing "Cell phone" and "Cable" and assign it to category "Test"
4. Create another bundle product "Cell pack 2" containing "Cell phone 2" and "Cable" and assign it to category "Test"
5. Open category "Test"

### Expected result
1. Both "Cell pack 1" and "Cell pack 2" products should be visible in category

### Actual result
1. Only one "Cell pack" product is visible in category

The actual cause for this is in https://github.com/Smile-SA/elasticsuite/blob/2.4.x/src/module-elasticsuite-catalog/Model/ResourceModel/Product/Indexer/Fulltext/Datasource/AttributeData.php#L208 - the "group by" statement groups query by child products, meaning if one child product is assigned to multiple bundle products, all except the first connection are lost (group concat affects only "configurable_attributes" field. 

Believe the query result should be grouped by parent_product_id as well?

Query before:
> SELECT `parent`.`entity_id` AS `parent_id`, `child`.`entity_id` AS `child_id`, GROUP_CONCAT(DISTINCT super_table.attribute_id SEPARATOR ',') AS `configurable_attributes` FROM `catalog_product_bundle_selection` AS `main`
 INNER JOIN `catalog_product_entity` AS `parent` ON parent.entity_id = main.parent_product_id
 INNER JOIN `catalog_product_entity` AS `child` ON child.entity_id = main.product_id
 LEFT JOIN `catalog_product_super_attribute` AS `super_table` ON super_table.product_id = main.parent_product_id
 INNER JOIN `catalog_product_website` AS `websites` ON websites.product_id = main.product_id AND websites.website_id = '1' 
 WHERE (parent.entity_id in (2399, 1534)) GROUP BY `main`.`product_id`
 
Result: http://prntscr.com/hh0ewj

-------

Query after:
[...]

PR #640, opened 2017-11-29
Fix group statement used to get bundle products data to index

sambolek:
See issue #639


afoucret:
Seems good to me.
Merged in branch 2.4.x instead of master.

Will be released in the next bugfix release.

PR #648, opened 2017-12-06
Feature data ingest

afoucret:
Create additional indices to store data from the tracker :

- `tracking_log_event` : Store individual tracking events

- `tracking_log_session` : Store aggregated session data for easier parsing


romainruaud:
@afoucret I do not see any low priority or locking prevention system on your resource model responsible for event data saving. Is this normal ?

afoucret:
Locking should not be a problem there as the table should be kept small by deleting event when indexed. So every minute, events are indexed and deleted from the DB.

PR #649, opened 2017-12-06
Refactoring product collection coverage.

### LINKED ISSUE #645: Facet coverage rate only accounting attribute set hits?
### Preconditions

Magento Version : 2.1.9CE

ElasticSuite Version :  "2.3.6"

Environment : Production / Developer

### Steps to reproduce

1. Have thousands of products divided over 6 main attribute sets
2. Create a new product attribute in Admin (in our case it's an option list) and set the field for ~100 products that can be in any attribute set.
3. Set the facet to lets say 50%

### Expected result

1. The filter not shown when all products are listed
2. The filter should not be shown when there are less than 50% products in the list that have the field set. (For example with these results, no list with more than 200 products could be logically valid to show the filter)

### Actual result

1. The filter IS shown in all category views

When I dug into the code that is responsible for this, I found `Smile\ElasticsuiteCatalog\Block\Navigation::addFacets()`, where every filter candidate is counted using the `$countBySetId` count. This couldn't logically work for anything other than attribute_set_id facets, because for every filter, the count for the specific attribute set only is used. Every filter gets the same `facetCoverageRate` and therefore the specific hit rate for that attribute is totally pointless. Looking at the code the problem should be worst when all products are in the same attribute set, because then all the filters must have the same `$currentCoverageRate` (if it's not 0) and therefore yield the same results.

I do hope, I've terribly misread something, but for now, I definitely need help with this one.

romainruaud:
As we discussed together, I refactored the product collection coverage.

This time, no additional PHP classes, I simply put it at the same place of the "countByAttributeSetId".

I decided to let the count by attribute set id in place, and flag it as deprecated (since it was a public method). I just added a new bucket into the loadProductCount method, to be used when calculating coverage rate in the Navigation block.

This will also fix #645 btw

### DIFF
--- a/src/module-elasticsuite-catalog/Block/Navigation.php
+++ b/src/module-elasticsuite-catalog/Block/Navigation.php
@@ -156,23 +156,8 @@ class Navigation extends \Magento\LayeredNavigation\Block\Navigation
      */
     private function addFacets()
     {
-        $productCollection    = $this->getLayer()->getProductCollection();
-        $countByAttributeCode = $productCollection->getProductCountByAttributeCode();
-        $totalCount           = $productCollection->getSize();
-
         foreach ($this->filterList->getFilters($this->_catalogLayer) as $filter) {
-            try {
-                $attribute           = $filter->getAttributeModel();
-                $facetCoverageRate   = $attribute->getFacetMinCoverageRate();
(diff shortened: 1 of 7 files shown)


romainruaud:
This one will be a pre-requisite for the "Filterable Attributes List by Category" refactoring, so let me know quickly if it has to be reworked.

romainruaud:
@afoucret I updated the PR.

PR #654, opened 2017-12-08
Feature category facet table

romainruaud:
A proposal of implementation for the loading of filterable attributes from a table associated to each category.

Just the table and the frontend logic : 

- everything should function as previously with an empty table.
- you can add some data into the table to test (set the display_mode to 0 or 1 to force hide/display of an attribute).

Let me know.


romainruaud:
I updated the PR. For now it uses what I've done for #656 but I may change it if you want.

Post-filtering and re-sorting an object of the FilterableAttributesListInterface is not easy since this one is aimed to be an immutable object (only getList method is public and has no parameters).

I'm kinda stuck by Magento design on this one.

let's talk about it

afoucret:
Replaced by #668

PR #656, opened 2017-12-11
Moving the  filtering of filters according to their coverage in anoth…

romainruaud:
…er method.

I decided to create a new method in the FilterList object.

This is not possible to proceed post filtering of filters according to the coverage inside getFilters(), since the getFilters() method is called also before applying the filters : coverage is inconsistent when calculated at this step.

I'm not a big fan of having another method doing the post-filtering, but I do not see any other option here. But with this approach, this new method has only one purpose and is clear to understand.

Let me know


afoucret:
Replaced by #665

PR #658, opened 2017-12-14
Prevent facet autocomplete search field to have duplicated ids.

### LINKED ISSUE #657: Duplicate element of 'search' with layered navigation autocomplete boxes
### Preconditions
Magento Version : 2.2.2 CE
ElasticSuite Version : 2.4.2
Environment : Developer
Third party modules : None (clean install)
OS: Ubuntu 17.10
PHP: 7.1.11-0ubuntu0.17.10.1

### Steps to reproduce
Go to a category with at least one attribute with enough values to bring up an autocomplete box (e.g. material on the sample data.)

### Expected result
There should not be any elements on the page with the same element id.

### Actual result
The layered navigation autocomplete inputs all have the id of 'search', as does the standard search box in the header.  This doesn't seem to cause any problems, but ids are supposed to be unique and Chrome 63 is now eager to remind you of this, logging a warning to the console on every catalogue page.

The id is defined in https://github.com/Smile-SA/elasticsuite/blob/8062b32f9c18fed164d3f47785588f35d765e580/src/module-elasticsuite-catalog/view/frontend/web/template/attribute-filter.html#L3

Everything seems to work without any id being defined, so maybe just remove it?

romainruaud:
Fix #657

### DIFF
--- a/src/module-elasticsuite-catalog/view/frontend/web/template/attribute-filter.html
+++ b/src/module-elasticsuite-catalog/view/frontend/web/template/attribute-filter.html
@@ -1,6 +1,6 @@
 <div class="field search" data-bind="if: displaySearch">
     <div class="control">
-        <input id="search"
+        <input class="filter-search"
                type="text" 
                data-bind="attr: {placeholder: searchPlaceholder}, event: {keyup: onSearchChange.bind(this), focusout: onSearchFocusOut}" />
     </div>


romainruaud:
Hmmm nope. I will provide a fix for 2.4.x branch instead of master. Sorry

PR #666, opened 2017-12-20
Fix invalid document order node

Fdec:
Datasources may be declared after mapping in document type definition.


afoucret:
Will be merged in branch 2.4.x and part of the next bugfix release.

PR #684, opened 2018-01-10
Adding Category facet config UI and persistency.

romainruaud:
I polished your prototype of UI and added persistency for data.

I also fixed little things : 

- 'display_mode' column was not added to select in collection.
- I renamed 'display_mode' to 'facet_display_mode' to be more consistent with other fields.
- 'facet_sort_order' is now using a dedicated source model, so I also refactored the Attribute Edit plugin to use it.

Let me know


romainruaud:
I refactored the way "use_default" checkboxes are made. I also added the "use_default" for facet_min_coverage as we said together.

What remains to be done now is to stylish a bit the UI.

I keep you in touch.

afoucret:
@romainruaud 
Waiting for the fix we talked about yesterday to merge it.

romainruaud:
The fix we talked about yesterday was in previous commit.

Now I have polished the UI, even if it remains quite "complex" when looking at it.

Let me know if it pleases you.

Regards

src/module-elasticsuite-catalog/Observer/PrepareFilterableAttributeList.php
```diff
@@ -0,0 +1,80 @@
...
+            $sortOrder = $request->getParam('facet_config_order', []);
+            $position  = [];
+
+            foreach ($data as $key => &$item) {
+                $item['position'] = isset($sortOrder[$key]) ? $sortOrder[$key] : $item['position'];
+                $position[] = (int) $item['position'];
+            }
+            array_multisort($position, SORT_ASC, $data);
+
+            $category->getResource()->addCommitCallback(
```
afoucret: I think this should not be done after commit.
If an error occurs during attribute config save, nothing should be saved.

afoucret:
Started a review that need to be addressed.
Additionally :
* On large screen : use default are not displayed under the field but on it's side

* There is not relevant row in the DB after the category save :

```
SELECT * FROM smile_elasticsuitecatalog_category_filterable_attribute;
+-----------+--------------+----------+--------------------+-------------------------+----------------+------------------+
| entity_id | attribute_id | position | facet_display_mode | facet_min_coverage_rate | facet_max_size | facet_sort_order |
+-----------+--------------+----------+--------------------+-------------------------+----------------+------------------+
|        20 |           77 |     NULL |                  0 |                    NULL |           NULL | NULL             |
|        20 |           83 |     NULL |                  0 |                    NULL |           NULL | NULL             |
|        20 |           93 |     NULL |                  0 |                    NULL |           NULL | NULL             |
|        20 |          134 |     NULL |                  0 |                    NULL |           NULL | NULL             |
|        20 |          135 |        2 |                  0 |                    NULL |           NULL | NULL             |
|        20 |          136 |     NULL |                  0 |                    NULL |           NULL | NULL             |
|        20 |          137 |        1 |                  0 |                    NULL |           NULL | NULL             |
|        20 |          138 |     NULL |                  0 |                    NULL |           NULL | NULL             |
|        20 |          139 |     NULL |                  0 |                    NULL |           NULL | NULL             |
|        20 |          140 |     NULL |                  0 |                    NULL |           NULL | NULL             |
|        20 |          141 |     NULL |                  0 |                    NULL |           NULL | NULL             |
|        20 |          142 |     NULL |                  0 |                    NULL |           NULL | NULL             |
|        20 |          143 |     NULL |                  0 |                    NULL |           NULL | NULL             |
|        20 |          144 |     NULL |                  0 |                    NULL |           NULL | NULL             |
|        20 |          145 |     NULL |                  0 |                    NULL |           NULL | NULL             |
|        20 |          146 |     NULL |                  0 |                    NULL |           NULL | NULL             |
|        20 |          147 |     NULL |                  0 |                    NULL |           NULL | NULL             |
|        20 |          148 |     NULL |                  0 |                    NULL |           NULL | NULL             |
|        20 |          149 |     NULL |                  0 |                    NULL |           NULL | NULL             |
|        20 |          150 |     NULL |                  0 |                    NULL |           NULL | NULL             |
|        20 |          151 |     NULL |                  0 |                    NULL |           NULL | NULL             |
|        20 |          152 |     NULL |                  0 |                    NULL |           NULL | NULL             |
|        20 |          153 |     NULL |                  0 |                    NULL |           NULL | NULL             |
+-----------+--------------+----------+--------------------+-------------------------+----------------+------------------+
```

* UI should be hidden on root categories

src/module-elasticsuite-catalog/Observer/PrepareFilterableAttributeList.php
romainruaud: I got it. But this looks like we'll have to refactor the product sorter too, because it is actually saving product positions the same way... That's where I picked the addCommitCallback solution :)

I will refactor saving of filterable attributes, probably via a plugin.

Regards.

romainruaud:
- On large screen : use default are not displayed under the field but on it's side

Hmm... I did it intentionally... felt it was prettier like this. I can change if you want.

- There is not relevant rows in the DB after the category save

Okay, I'll look at this one.

- UI should be hidden on root categories

This is the case on my instance. Hidden for new and root categories. See DataProviderPlugin::aroundGetMeta where the filtering is done.

romainruaud:
PR Updated : 

- now saving values via a plugin, to ensure failure are properly handled
- now saving only relevant rows

Let me know about the rest, for me the UI is properly hidden on root categories (double checked this). About the "use default" displayed aside, as I said it was by design, let me know if it's mandatory to change it.

Regards

romainruaud:
PR Updated : 

- moved attributes saving inside an extension savehandler
- set the "use default value" checkbox to always be on a new line even on large screens.

PR #685, opened 2018-01-11
Fix autocomplete special price

### LINKED ISSUE #682: Special price doesn't show by autocomplete
### Preconditions
When you search for an product and it has an special price. the normale price wil showed in the dropdown screen.

i mean the top search balk in the header.
when you type your product. you see different products that the customer means. and the it didn't show an special price if the product has an special price.

Magento Version : 2.1.9
ElasticSuite Version : 2.3.0
Environment : production and development

romainruaud:
This one fix #682 

Main problem is due to the fact that "special_price" is not selected in autocomplete. I thought `addPriceData()` would do this, but it's not the case.

The only reason the same bug is not appearing on catalog listing is that special_price is set to `used_in_product_listing=1` and get automatically loaded by the layer.

I did not want to load all attributes `used_in_product_listing=1` in the autocomplete (for performances concern, however, this would have fixed the issue easily), so I added only 'special_price' to the list of default selected attributes.

I intentionally omitted to add "special_from_date" and "special_to_date".

This is due to the fact that they are used only with Magento CE and this could cause errors when these attributes are missing (if they were deleted, or when using Magento EE).

To allow proper handling of the special price when using these attributes, one will have to add them manually via DI like this : 

```
    <type name="Smile\ElasticsuiteCatalog\Model\Autocomplete\Product\AttributeConfig">
        <arguments>
            <argument name="additionalSelectedAttributes" xsi:type="array">
                <item name="special_from_date" xsi:type="string">special_from_date</item>
                <item name="special_to_date" xsi:type="string">special_to_date</item>
            </argument>
        </arguments>
    </type>
```

Let me know if this solution is acceptable for you, or if we should do it another way ?

### DIFF
--- a/src/module-elasticsuite-catalog/Model/Autocomplete/Product/AttributeConfig.php
+++ b/src/module-elasticsuite-catalog/Model/Autocomplete/Product/AttributeConfig.php
@@ -32,7 +32,7 @@ class AttributeConfig
     /**
      * @var string[]
      */
-    private $defaultSelectedAttributes = ['name', 'thumbnail'];
+    private $defaultSelectedAttributes = ['name', 'thumbnail', 'special_price'];
 
     /**
      * @var string[]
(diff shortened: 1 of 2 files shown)


afoucret:
@romainruaud Can you please append date to the PR and test it against Magento EE.

BR,

romainruaud:
PR Update : 

- I added date fields.

Tested working OK with EE version.

PR #693, opened 2018-01-17
Add ES server configuration options to Magento setup:install command

### LINKED ISSUE #692: Installing the module with Docker : suboptimal step to define ES location
While installing the module on a dockerized development environment, the ES instance is not situated in localhost but in a custom alias. We should apply [this step](https://github.com/Smile-SA/elasticsuite/wiki/ModuleInstall#optional--installing-the-module-when-using-a-custom-elasticsearch-server-location) but it would require to set an environment specific value in a global setting...

### Preconditions

Magento Version : any (but here : 2.2 EE B2B)

ElasticSuite Version : any

Environment : any

Third party modules : N/A

### Tests done

* I tried to use Magento 2.2 `config:set` before installing the module, but it complain the configuration does not exists (and we cannot install the module before adding this configuration !!)
* I could set this value in env.php, [according to new patterns from Magento](http://devdocs.magento.com/guides/v2.2/config-guide/cli/config-cli-subcommands-config-mgmt-import.html) but I wanted to commit this step somewhere, for easy dev env install
* Eventually I just load the value with `INSERT INTO vr_core_config_data (path, value) VALUES ('smile_elasticsuite_core_base_settings/es_client/servers', 'elasticsearch:9200')` in an install script... but it's far from awesome :(

### Ideas

I know it's not a easy thing to fix, but first, the documentation is wrong. It should at least use Magento 2.2 patterns...

In a perfect world, the module should be installable easily, and this configuration should not be used in the upgrade script, but at least in a separate command, or triggered when doing the first indexation...

careys7:
Allows users to specify connection flags, such as `--es-server` to `setup:install`.

Unsure of the best way to approach the use of deprecated ClientFactory. 

## Fixes 
Fixes #692 - issue with installation process (as currently described in the wiki) requiring a new module to define a custom elasticsearch endpoint (where it is not localhost:9200) to allow install to succeed. 

---

## Integration Test Compatibility

This PR also allows the Magento integration test framework  to run when elasticsuite is installed (which is our immediate motivation for adding this).

Example integration test configuration:

`dev/tests/integration/phpunit.custom.xml`
```xml
<phpunit>
   <testsuites>
      <!-- Define custom test suite, eg app/code/Vendor -->
   </testsuites>
   <php>
      <!-- Specify a custom install configuration file --->
      <const name="TESTS_INSTALL_CONFIG_FILE" value="etc/install-config-mysql.custom.php"/>
   </php>
</phpunit>
```

`dev/tests/integration/etc/install-config-mysql.custom.php`
```php
<?php
return [
    'db-host' => 'mysql',
    'db-user' => 'root',
    'db-password' => 'root',
    'db-name' => 'magento_integration_tests',
    'db-prefix' => '',
    'backend-frontname' => 'backend',
    'admin-user' => \Magento\TestFramework\Bootstrap::ADMIN_NAME,
    'admin-password' => \Magento\TestFramework\Bootstrap::ADMIN_PASSWORD,
    'admin-email' => \Magento\TestFramework\Bootstrap::ADMIN_EMAIL,
    'admin-firstname' => \Magento\TestFramework\Bootstrap::ADMIN_FIRSTNAME,
    'admin-lastname' => \Magento\TestFramework\Bootstrap::ADMIN_LASTNAME,
    // Specify the --es-server install flag to setup::install
    'es-server' => 'elasticsearch:9200',
];
```

`dev/tests/integration/etc/config-global.php`
```php
<?php
// Set the default search engine to Elasticsuite
return [
    'catalog/search/engine' => 'elasticsuite',
];
```

Running as follows:

```bash
$ vendor/bin/phpunit -c dev/tests/integration/phpunit.custom.xml --testsuite Custom
```

### DIFF
--- a/src/module-elasticsuite-core/Helper/ClientConfiguration.php
+++ b/src/module-elasticsuite-core/Helper/ClientConfiguration.php
@@ -14,6 +14,7 @@
 
 namespace Smile\ElasticsuiteCore\Helper;
 
+use Magento\Framework\App\Helper\AbstractHelper;
 use Smile\ElasticsuiteCore\Api\Client\ClientConfigurationInterface;
 
 /**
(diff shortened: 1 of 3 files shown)


afoucret:
Hi @careys7,

I have fully refactored the client instantiation and removed the deprecated ClientFactory class.

As a result, the PR #716 does the job and handle es config from the setup:install CLI.

Issue #506, opened 2017-08-18
Add preloader/spinner for ajax requests of autocomplete


Koc:
It would be nice display some gif-spinner before sending ajax request and hide it on ajax complete. With selector option.

romainruaud:
Hello @Koc 

This is not something we plan to add on the module, since this is not the way Magento is working actually.

Feel free to add by yourself on your project if you need it. I close the issue for now.

Best regards,

PR #717, opened 2018-01-31
Add 'processing' class to the autocomplete search form when searching.

### LINKED ISSUE #669: Proposal: add class to input or form when searching
it would be useful to be able to have a class applied to the input or to the form when the search starts and obviously will be removed once completed, this would allow to assign a different style or an icon to improve the UX

romainruaud:
Fix #669


Koc:
Seems this PR also fixes #506

Issue #729, opened 2018-02-06
Missing addOptionsData in filter options


Leone:
When you change option label in the filter attribute in the admin panel, filter navigation on the frontend disappears.

### Preconditions

Magento Version : 2.1.10 CE

ElasticSuite Version : 2.3.0, but does not matter

Environment : does not matter

Third party modules : does not matter

### Steps to reproduce

1. edit attribute in the admin panel
2. change label in some option
3. save changes

### Expected result

1. Filters should be still visible on the frontend.

### Actual result

Filter nav (the whole block) on the frontend disappears.
We found in logs:

> main.CRITICAL: Notice: Undefined index: adminSortIndex in /*/vendor/smile/elasticsuite/src/module-elasticsuite-catalog/Model/Layer/Filter/Attribute.php on line 277 [] []

The problem is that script doesn't set key adminSortIndex to all items in array $items and a few lines below the script tries to sort these array using this key.

### Possible solution

In class Smile\ElasticsuiteCatalog\Model\Layer\Filter\Attribute, method addOptionsData(),
change this:
```
usort($items, function ($item1, $item2) {
    return $item1['adminSortIndex'] <= $item2['adminSortIndex'] ? -1 : 1;
});
``` 
to this:
```
usort($items, function ($item1, $item2) {
    // check if key adminSortIndex exists
    if(!isset($item1['adminSortIndex']) or !isset($item2['adminSortIndex'])){
        return 0;
    }
    return $item1['adminSortIndex'] <= $item2['adminSortIndex'] ? -1 : 1;
});
```

afoucret:
Hi @Leone 

I will try to look at your PR quickly now.

afoucret:
PR have been merged and will be part of the next maintenance release (2.5.1).

Thx for contributing.

PR #736, opened 2018-02-08
Check if adminSoertIndex key exists

Leone:
Issue: https://github.com/Smile-SA/elasticsuite/issues/729

This one fix the missing adminSortIndex issue.
If you change the option label in the filter attribute in the admin panel, the script doesn't see this option, doesn't set key adminSortIndex and throws an error in usort() function.

This issue exists in all versions.

Regards,
Grzegorz

### DIFF
--- a/src/module-elasticsuite-catalog/Model/Layer/Filter/Attribute.php
+++ b/src/module-elasticsuite-catalog/Model/Layer/Filter/Attribute.php
@@ -267,10 +267,29 @@ class Attribute extends \Magento\CatalogSearch\Model\Layer\Filter\Attribute impl
                 }
             }
 
-            usort($items, function ($item1, $item2) {
-                return $item1['adminSortIndex'] <= $item2['adminSortIndex'] ? -1 : 1;
-            });
+            $items = $this->sortOptionsData($items);
         }
+
+        return $items;
+    }
+
+    /**
+     * Sort items by adminSortIndex key.
+     *
+     * @param array $items to be sorted.
+     *
+     * @return array
+     */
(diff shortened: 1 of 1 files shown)


afoucret:
Hi @Leone,

I think it would be nice to release this patch in the 2.5.1 release since the next major release (2.6.0) is not coming before 2 months.

Could you rebase this PR from branch 2.5.x, please ?

Leone:
Hi @afoucret,

I rebased pull request from branch 2.5.x.
Please keep in mind that in previous versions this problem was also present.

afoucret:
Hi @Leone,

I have merged it in the branch 2.5.x and it will be part of the next maintenance release (2.5.x).
Our maintenance policy is to apply this kind of non-security patch only on the latest release.

Thx for contributing,

src/module-elasticsuite-catalog/Model/Layer/Filter/Attribute.php
```diff
@@ -267,11 +267,30 @@ private function addOptionsData(array $items)
...
+     */
+    private function sortOptionsData(array $items)
+    {
+
+        usort($items, function ($item1, $item2) {
+            if (!isset($item1['adminSortIndex']) or !isset($item2['adminSortIndex'])) {
+                return 0;
+            }
+
+            return $item1['adminSortIndex'] <= $item2['adminSortIndex'] ? -1 : 1;
```
Koc: Looks like this can be replaced with single line:

```php
return $item1['adminSortIndex'] ?? 0 <=> $item2['adminSortIndex'] ?? 0;
```

Issue #743, opened 2018-02-12
Table not found : module doesn't manage table prefixes in 2.5.0 


shavounet:
### Preconditions
Magento Version : 2.2
ElasticSuite Version : 2.5
Environment : dev

### Steps to reproduce
1. Install a basic Magento with table prefixes & sample data
2. Go on product list page
3. Try to reindex

### Actual result
1. For now (following stacks), I've seen 3 places where table prefixes where not correctly implemented
   * `src/module-elasticsuite-catalog/Model/ResourceModel/Product/Search/Position.php:64`
   * `src/module-elasticsuite-catalog/Model/ResourceModel/Product/FilterableAttribute/Category/Collection.php:129`
   * `src/module-elasticsuite-catalog/Model/ResourceModel/Product/FilterableAttribute/Category/Collection.php:170`
2. I've still an issue (showing twice) while reindexing : 
``` $ ./bin/magento indexer:reindex
Design Config Grid index has been rebuilt successfully in 00:00:00
Customer Grid index has been rebuilt successfully in 00:00:00
Category Products index has been rebuilt successfully in 00:00:00
Product Categories index has been rebuilt successfully in 00:00:00
Product Price index has been rebuilt successfully in 00:00:06
Product EAV index has been rebuilt successfully in 00:00:00
Catalog Rule Product index has been rebuilt successfully in 00:00:03
Catalog Product Rule index has been rebuilt successfully in 00:00:00
Stock index has been rebuilt successfully in 00:00:02
Catalog Search indexer process unknown error:
mapper_parsing_exception: No handler for type [keyword] declared on field [type_id]
Product/Target Rule index has been rebuilt successfully in 00:00:04
Target Rule/Product index has been rebuilt successfully in 00:00:00
Sales Rule index has been rebuilt successfully in 00:00:00
ElasticSuite Category Indexing indexer process unknown error:
mapper_parsing_exception: No handler for type [keyword] declared on field [path]
ElasticSuite Thesaurus Indexing index has been rebuilt successfully in 00:00:00
```

afoucret:
PR have been merged and will be part of the next maintenance release (2.5.1).
Thx for reporting and contributing.

shavounet:
Note that there is another error that I haven't fixed (without knowing if normal or not).
I haven't done a full test either.

afoucret:
For your reindexing issue, I suppose you are working on branch master and not on the 2.5.x branch.
[...]

PR #744, opened 2018-02-12
#743 - fix missing table prefix while looking up db

shavounet:


afoucret:
Thank for the contrib. Very valuable.

PR #768, opened 2018-02-20
Fix setup install options

### LINKED ISSUE #763: Cannot deploy module to a fresh production install
This might or might not be a native Magento issue... but it prevent your module from being deployed in a standard production procedure.
Might be an edge case, or not... I'm not sure what is triggered

### Preconditions
Magento Version : 2.2 (EE B2B, but unrelevent)
ElasticSuite Version : 2.5 (with new --es-hosts install flags)
Environment : prod but reproducible in dev

### Steps to reproduce
1. Get yourself a new env, with no db, ES, environment var... just a composer install & right permissions
2. `./bin/magento setup:di:compile`
3. `./bin/magento setup:install` (with no params, only to trigger param validation)

### Expected result
1. An error message should tell us there is no alive ES node found

### Actual result
1. `Unable to connect ElasticSearch server : Notice: Undefined index: data in /var/www/html/vendor/magento/framework/App/Config/Initial.php on line 60`

Note that this message prevent totally the install, even with right parameters (including es-hosts).

Here are the workarounds, I found : 
* delete generated folder (not ideal due to specific hosting platform...)
* create a patch for Magento core to fix this silly notice what shouldn't exists anyway... (there is a second place to patch before it can work : `Notice: Undefined index: websites in /var/www/html/vendor/magento/module-config/App/Config/Type/System.php on line 248`)
   * Once applied, the error message is `Unable to connect ElasticSearch server : No alive nodes found in your cluster` which is fine

I think it might be due to your code checking configuration, looking for es-hosts definition.

romainruaud:
This one should fix #763 

It will also prevent various problems detected while investigating on this issue : 

- ensure to pass all configuration values to the client to ensure he will not call ScopeConfig->getValue during install process. Otherwise this leads to a fatal error due to the config not being already built by Magento.

- skip validation step of the reachable ES server if no parameter is submitted on the command line.
- ensure useless fields are not written into the env.php

And I also decided to cover this class with unit testing. 

This one is meant to be merged on 2.5 and cascaded to master.

Let me know @afoucret

### DIFF
--- a/src/module-elasticsuite-core/Setup/ConfigOptionsList.php
+++ b/src/module-elasticsuite-core/Setup/ConfigOptionsList.php
@@ -50,7 +50,7 @@ class ConfigOptionsList implements ConfigOptionsListInterface
     const CONFIG_PATH_ES_PASS  = self::CONF_PREFIX . '/http_auth_pwd';
 
     /**
-     * @var ClientFactoryInterface
+     * @var \Smile\ElasticsuiteCore\Client\ClientFactory
      */
     private $clientFactory;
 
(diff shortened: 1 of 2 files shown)


afoucret:
Closed in favor of #769

PR #773, opened 2018-02-20
Fix tracker injection

afoucret:


rbayet:
FYI, fixes work as expected

PR #778, opened 2018-02-21
Fix results qty accuracy for slider-based filters

mborkowski:
Range-slider has a bug with expected results counter. This leads to displaying wrong quantity of expected results. For example when I have filtered an attribute to values between 70 and 80 (my_attribute=70-80) this method will return 70-79.9999... instead of 70-80. This contribution will will fix accuracy.


afoucret:
Your PR have been rebased and completed here : #782

PR #781, opened 2018-02-22
Integration of search context

romainruaud:
This PR contains a first version of the SearchContext integration.

I also added default Visibility/Stock filter to ensure it is working as intended.

It actually works without breaking anything. Filters are only applied in frontend for now, and most of time duplicate already existing ones.

Let me know if the architecture seems ok for you now, and then I'll be able to process additional cleaning and update this PR (or create a new one).

Regards


romainruaud:
I updated the PR as we said yesterday (rename DefaultFilter to Filter, load the filters from XML files, and let the Container build the filter queries from the current search context).

Let me know if it suits you best this way.

For now the Context is still "immutable", I wait for your thought about this point too.

Regards

src/module-elasticsuite-core/Api/Search/Request/ContainerConfigurationInterface.php
```diff
@@ -74,4 +74,13 @@ public function getRelevanceConfig();
...
     public function getStoreId();
+
+    /**
+     * Retrieve filters for the container (visibility, in stock, etc ...) and the current search Context.
+     *
+     * @param \Smile\ElasticsuiteCore\Search\Context $searchContext Search Context
+     *
+     * @return \Smile\ElasticsuiteCore\Search\Request\QueryInterface[]
+     */
+    public function getFilters(\Smile\ElasticsuiteCore\Search\Context $searchContext);
```
afoucret: \Smile\ElasticsuiteCore\Search\Context should be an interface as it is used in the API area.

src/module-elasticsuite-core/Search/Context.php
```diff
@@ -0,0 +1,151 @@
...
+use Magento\Store\Model\StoreManagerInterface;
+
+/**
+ * Elasticsuite Search Context
+ *
+ * @category Smile
+ * @package  Smile\ElasticsuiteCore
+ * @author   Romain Ruaud <romain.ruaud@smile.fr>
+ */
+class Context
```
afoucret: The implemented method are OK but I want to be able to write arbitrary data in the context.
Maybe add customer_group by default in the context.

src/module-elasticsuite-catalog/Model/Product/Search/Request/Container/Filter/Visibility.php
```diff
@@ -0,0 +1,75 @@
...
+use Smile\ElasticsuiteCore\Search\Request\QueryInterface;
+
+/**
+ * Product Visibility Default filter.
+ *
+ * @category Smile
+ * @package  Smile\ElasticsuiteCatalog
+ * @author   Romain Ruaud <romain.ruaud@smile.fr>
+ */
+class Visibility implements FilterInterface
```
afoucret: Bad logic.
You should not have an if but 2 classes instead.

The visibility is not defined by having or not a search query but by the container itself.

romainruaud:
I updated the PR (and also rebased it) to take your requests into account.

Let me know your advices.

regards

afoucret:
Test are still broken !!!

romainruaud:
tests fixed !

PR #783, opened 2018-02-23
Fix nested filter in agg and sort orders.

afoucret:
Do not assume the generated query type is always the same.


afoucret:
@delyriand : Will be part of 2.5.3

PR #785, opened 2018-02-23
Add current nested path in filter query builder to condition the gene…

delyriand:
…ration nested query


afoucret:
Merged in branch 2.5.x

PR #794, opened 2018-03-02
Compute children SKUs into parent product.

### LINKED ISSUE #791: Configurable product won't filter based on associated product SKU's
If I try to search product using associated SKU then I am not getting any result.  

All associated products are not visible individually, can you please help urgently to fix this issues? 

It should show Configurable products in the list because the SKU I am searching belongs to the Configurable product.

romainruaud:
Fix #791 

I just added children SKUs to the parent document one, as seen together.

Regards

### DIFF
--- a/src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/AttributeData.php
+++ b/src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/AttributeData.php
@@ -61,6 +61,7 @@ class AttributeData extends AbstractAttributeData implements DatasourceInterface
                         $indexData[$parentId]['children_ids'][] = $childId;
                         $this->addRelationData($indexData[$parentId], $childrenIndexData[$childId], $relation);
                         $this->addChildData($indexData[$parentId], $childrenIndexData[$childId]);
+                        $this->addChildSku($indexData[$parentId], $relation);
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/AttributeData.php
```diff
@@ -205,4 +206,26 @@ private function filterCompositeProducts($indexData)
...
+     * Append SKU of children product to the parent product index data.
+     *
+     * @param array $parentData Parent product data.
+     * @param array $relation   Relation data between the child and the parent.
+     */
+    private function addChildSku(&$parentData, $relation)
+    {
+        $sku = $relation['sku'] ?? false;
+
+        if ($sku) {
```
afoucret: Is there really case where the sku can be missing ?

src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/AttributeData.php
```diff
@@ -205,4 +206,26 @@ private function filterCompositeProducts($indexData)
...
+            if (!isset($parentData['sku'])) {
+                $parentData['sku'] = [];
+            } elseif (!is_array($parentData['sku'])) {
+                $parentData['sku'] = [$parentData['sku']];
+            }
```
afoucret: May be written in a more concise way :

```php
if (isset($parentData['sku'] && !is_array($parentData['sku'])) {
    $parentData['sku'] = [$parentData['sku']];
}
```

romainruaud:
PR updated

PR #799, opened 2018-03-07
Frontend blacklist implementation.

romainruaud:


src/module-elasticsuite-catalog/Model/Product/Search/Request/Container/Filter/CategoryBlacklist.php
```diff
@@ -0,0 +1,95 @@
...
+ * Do not edit or add to this file if you wish to upgrade Smile Elastic Suite to newer
+ * versions in the future.
+ *
+ * @category  Smile
+ * @package   Smile\ElasticSuiteCatalog
+ * @author    Romain Ruaud <romain.ruaud@smile.fr>
+ * @copyright 2018 Smile
+ * @license   Open Software License ("OSL") v. 3.0
+ */
+namespace Smile\ElasticsuiteCatalog\Model\Product\Search\Request\Container\Filter;
```
afoucret: The namespace seems to be wrong.
The field `is_blacklisted` you are referring to is declared in the `Smile_ElasticsuiteVirtualCategory`module and you are using the filter in the `Smile_ElasticsuiteCatalog`module.

+ `di.xml` file need to be fixed.

romainruaud: I'm not sure what you want me to do with the di.xml ?

The field `is_blacklisted` is defined (for the ES index) in the `ElasticsuiteCatalog` module, whereas the DB column of the table is defined in `ElasticsuiteVirtualCategory`.

I agree, that's strange.

Imho there is more to refactor here, since the `smile_virtualcategory_catalog_category_product_position` table is used for virtual AND non-virtual category product positioning, which is quite unclear (for us but also for others) and forces us to add some code in the `VirtualCategory` module even if we are just modifying the merchandiser behavior.

I can move the filter declaration (in `search_requests.xml` file) into the VirtualCategory module if you want, but I'm not sure this was exactly what you were saying ?

let me know

afoucret: You're right. I was talking about `search_requests.xml` and not `di.xml`

1. Category merchandising comes with the VirtualCategory module.

2. All related fields should be defined in the VirtualCategory module, so the `'is_blacklisted` field MUST be in declared in the VirtualCategory just like the ìs_virtual_field`.
In the end, the indexation of this field is not managed in `Smile\ElasticsuiteCatalog\Model\ResourceModel\Product\Indexer\Fulltext\Datasource\CategoryData`

3. The filter should be declared in the`search_request.xml` file of the VirtualCategory module. It has not been tested yet but these files should be merged accross different module and you should not have to specify `label`, `index`, `type` or `fulltext` again in the VirtualCategory module.

4. You have missed to implement the `$_idAttributes` in `Smile\ElasticsuiteCore\Search\Request\ContainerConfiguration\BaseConfig\Reader``

romainruaud: Just one point remains about file merging : actually, this is mandatory to specify `label`, `index`, `type` and `fulltext` because the XSD schema says it's required.

I could set the field as "optional" in the XSD schema to handle this case, is it ok for you ?

afoucret: Wonder if it could be fixed if  `Smile\ElasticsuiteCore\Index\Indices\Config\SchemaLocator::getPerFileSchema` method return null.

romainruaud: I tested to return null for the `getPerFileSchema` of `Smile\ElasticsuiteCore\Search\Request\ContainerConfiguration\BaseConfig\SchemaLocator` and yes it works, I can define more search request without specifying their index, label, and so on..

Seems legit, for the `system.xml` there are 2 different schemas : one for the "merged" file, and another for each file declared in modules.

Some core modules are also using `null` for the per-file schema.

We could consider : 

- return `null` for the `getPerFileSchema`

- or using another schema (but this would imply to change the SchemaLocation in the file headers).

Let me know

romainruaud: Ok I added `null` as return for the `getPerFileSchema` as seen together.

I can re-use an already defined search container without error.

If I try to create a new search container without label,index, type or fulltext, the XSD validation of the merged file throws an error, so I think we are fine.

Regards.

PR #804, opened 2018-03-08
Cleanup tracking cookies when customer logs out.

romainruaud:


src/module-elasticsuite-tracker/Observer/CleanCookiesOnLogout.php
```diff
@@ -0,0 +1,95 @@
...
+     * @param \Magento\Framework\Event\Observer $observer The observer
+     *
+     * @return void
+     */
+    public function execute(\Magento\Framework\Event\Observer $observer)
+    {
+        $cookieNames = $this->getCookies();
+
+        $metadata = $this->cookieMetadataFactory->createPublicCookieMetadata();
+        $metadata->setPath('/');
```
afoucret: What about site hosted under a different path ? Bug ?

romainruaud: Indeed, it may be a bug, but probably in the tracker himself because it is actually enforcing the cookie path on "/" : https://github.com/Smile-SA/elasticsuite/blob/master/src/module-elasticsuite-tracker/view/frontend/web/js/tracking.js#L35

afoucret: Should be fixed for both.

romainruaud: ok for me, I was not sure if the arbitrary "/" on tracker was by design or a bug. I will use default magento cookie path for the tracker so this should work server/client side in the same way.

PR #818, opened 2018-03-13
Feature gdpr reconciliation table

romainruaud:
- Create the table
- Adding a Customer Reconciliation Service
- Call this service when the tracking controller gets hit
- Cleanup customer data in other code location


romainruaud:
PR Updated : I added a configurable delay, and a cronjob processing cleaning of reconciliation data based on this delay.

src/module-elasticsuite-tracker/Api/Customer/ReconciliationServiceInterface.php
```diff
@@ -0,0 +1,62 @@
...
+    public function setDeleteAfterByCustomerId(int $customerId, \DateTime $deleteAfter);
+
+    /**
+     * Delete all reconciliation data that are set to expire before a given date.
+     *
+     * @param \DateTime $deleteAfter The date after which all the customer data must be cleared.
+     *
+     * @return void
+     */
+    public function deleteByDeleteAfter(\DateTime $deleteAfter);
```
afoucret: I think the service should expose a simple deleteExpired method without param.

src/module-elasticsuite-tracker/Api/Customer/ReconciliationServiceInterface.php
```diff
@@ -0,0 +1,62 @@
...
+namespace Smile\ElasticsuiteTracker\Api\Customer;
+
+/**
+ * Customer Reconciliation : used to handle correlation between customer_id and session_id/visitor_id
+ *
+ * @category Smile
+ * @package  Smile\ElasticsuiteTracker
+ * @author   Romain Ruaud <romain.ruaud@smile.fr>
+ */
+interface ReconciliationServiceInterface
```
afoucret: Not sure if I want to name this ReconciliationServiceInterface.
I think that CustomerTrackingServiceInterface would be a better name.

src/module-elasticsuite-tracker/Api/Customer/ReconciliationServiceInterface.php
```diff
@@ -0,0 +1,62 @@
...
+    public function addByCustomerId(int $customerId, string $sessionId, string $visitorId);
+
+    /**
+     * Delete all reconciliation data for a given customer.
+     *
+     * @param int $customerId The customer Id.
+     *
+     * @return void
+     */
+    public function deleteByCustomerId(int $customerId);
```
afoucret: Think this two method should be merged into a single one : anonymizeCustomerData(int customerId, delay = null) 
* If delay is null : delete now
* Else place the deleteAfter

src/module-elasticsuite-tracker/Api/Customer/ReconciliationServiceInterface.php
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * Add a reconciliation tuple into the database.
+     *
+     * @param int    $customerId The customer Id.
+     * @param string $sessionId  The session Id.
+     * @param string $visitorId  The visitor Id.
+     *
+     * @return void
+     */
+    public function addByCustomerId(int $customerId, string $sessionId, string $visitorId);
```
afoucret: Replace by : addEvent($eventData)
* The event data is stored in the queue
* The reconciliation is saved if needed

We hide the implementation.
Replace the \Smile\ElasticsuiteTracker\Api\EventQueueInterface by our CustomerTrackingServiceInterface in the hit controller.

romainruaud:
I revamped the service according to what you said.

Not sure about "where" the customer session should be handled. Actually it's injected in the service, but I'm not a huge fan of it, maybe the work should be done by the controller before, and just pass the customer_id in the event (to be removed later).

Let's discuss about this later

romainruaud:
I just rebased this one

romainruaud:
- `getVisitorIds(int $customerId)` has been added
- I renamed all "reconciliation" stuff to `CustomerLink` (including the table itself).

romainruaud:
squashed

PR #823, opened 2018-03-16
Refactor user consent

afoucret:


afoucret:
@romainruaud : let you review and test

Give me your feedback when done.
I will merge.

romainruaud:
Ok for me, I am able to : 

- override the user-consent component to a specific one (tested with dummies true/false)
- add custom config to the component via the layout and read it.

You can merge

PR #824, opened 2018-03-19
Fix for boolean filters with „NO“ value

annakoehler:
- index: support 0 values
- filter collection with values = 0
- "NO" option label

### DIFF
--- a/src/module-elasticsuite-catalog/Helper/AbstractAttribute.php
+++ b/src/module-elasticsuite-catalog/Helper/AbstractAttribute.php
@@ -167,7 +167,8 @@ abstract class AbstractAttribute extends Mapping
         }
 
         $value = array_map($this->attributeMappers[$mapperKey], $value);
-        $value = array_filter($value);
+        // not using array_filter here because it could remove "0" string from values
+        // $value = array_filter($value);
         $value = array_values($value);
         $values[$attributeCode] = $value;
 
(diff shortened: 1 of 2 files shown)


afoucret:
hi @dmcanna, 

Can you please describe what this PR fix and how existing features are modified ?

romainruaud:
Imho this is a fix to index 0 values (what he calls "No") into the ES index.

However, I'm not sure if this is intended to index 0 or "0" ... So yes more details are needed.

dmcanna:
Hi,
we had the problem that filter (on category page or catalogsearch page) with all our boolean attributes was only with "yes" values, but we need "no" values too. For example you have an boolean attribute "Humidity resisstant" with "yes" and "no" values and customer would like to see products with "Humidity resisstant = NO" too. 
After research we have found that the boolean attributes with "no" values were not saved in elasticuite.  After we fixed this, we found that 0 option in filter have a label "0" instead of "NO". And collections were not filtered if we had your like ?humidity_resisstant=0
I hope I described it good, but if you need more details please let me know.

sky-hub:
Detailed description for this issue is provided and fixed in this PR: https://github.com/Smile-SA/elasticsuite/pull/1305

romainruaud:
I close this one (way too old).

#1305 seems to fit better.

PR #826, opened 2018-03-20
Use is_used_in_promo_rules instead of is_searchable for the image attribute.

### LINKED ISSUE #825: image attribute should not be set as searchable
We should set **is_used_in_promo_rules** to true instead.

afoucret:
Fix #825

### DIFF
--- a/src/module-elasticsuite-catalog/Setup/CatalogSetup.php
+++ b/src/module-elasticsuite-catalog/Setup/CatalogSetup.php
@@ -288,7 +288,7 @@ class CatalogSetup
 
         $setup->getConnection()->update(
             $setup->getTable('catalog_eav_attribute'),
-            ['is_searchable' => 1],
+            ['is_used_for_promo_rules' => 1, 'is_searchable' => 0, 'is_used_in_spellcheck' => 0],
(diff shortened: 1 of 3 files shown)


afoucret:
@romainruaud : can you test it please.

romainruaud:
This one works as intended for me : 

- db schema properly updated
- 'image' field still in mapping and indexed

This can be merged.

Just one point, on my dev environment, 'image' was (this is legacy, I can't remember why) "used_in_spellcheck=1".

This implies that an attribute can be "is_searchable=0" BUT "is_used_in_spellcheck=1" and would then have "copy_to : 'spelling'" in the mapping.

Maybe this should be also fixed in a separate issue ?

afoucret:
I fixed the is_used_in_spellcheck

romainruaud:
Great for the "image" attribute, but I was thinking a more global fix, shouldn't we ?

Imho we should not be able to set an attribute to "is_searchable=0" and "is_used_in_spellcheck=1".

Your thoughts ?

afoucret:
Yep. Propose a patch for this.

romainruaud:
Acknowledged, I will work on a separate PR for this one.

You can merge this one for now.

PR #828, opened 2018-03-20
Fix prevent spellcheck fields

### LINKED ISSUE #827: Non searchable attributes should not be used for spellcheck.
### Preconditions

Magento Version : any

ElasticSuite Version : any 

Environment : irrelevant

Third party modules :

### Steps to reproduce

1. Set an attribute to "is searchable : No" but "is used for promo rules : yes"
2. Save it & reindex
3.

### Expected result

1. The field on ES should not be copied to "spelling" meta field

### Actual result

1. The field has the "copy_to : spelling"

romainruaud:
Fix #827 

I discovered there was also a bug in the BO attribute edit block, this PR embeds the fix (dependency block was broken).

I changed the isUsedInSpellecheck method for FieldInterface and added the associated unit test.

Let me know


romainruaud:
squashed !

PR #852, opened 2018-04-06
Avoid null pointer exception when storeId is not defined

shavounet:
I have an issue that I can't exactly reproduce on a test environment... All I know, is once some setup has been done with virtual categories, I have a null pointer error on search results, coming from this snippet.

I've fixed this with the most logic code (and it works) for me, but I can't tell if something else should be done.
Feel free to send me feedback.


Koc:
@shavounet thank you for debugging this. Hope it will help me for fixing my issue, but I cann't check it now because my store locked with Magento 2.1.x and ElasticSuite 2.5.9 supports only 2.2.x

Issue #855, opened 2018-04-10
Suggest for Quick Order functionality doesn't work


pakoordoniez:
Suggest doesn't work for "Enter SKU or Product Name" field in Quick Order B2B functionality when product name or incomplete SKU is provided.

### Preconditions
- Debian 8
- PHP 7.1.12
- nginx 1.13.7
- Percona 5.7.20
- ElasticSearch 5.6.7

Magento Version : Commerce (EE) 2.2.3 with B2B modules in version 1.0.3

ElasticSuite Version : 2.5.6

Environment : Developer and Production

Third party modules : Same problem at clean Magento instance.

### Steps to reproduce
1. Make sure you have enabled "Quick Order" functionality in "Configuration > General > B2B Features".
2. Register and login as B2B customer.
3. Go to Quick Order page with form to provide products SKU.
4. Open browser developer tools and filter requests by XHR type.
5. Type product name in "Enter SKU or Product Name" and check response for `/quickorder/ajax/product_suggest/?q=joust` request. It should contain call stack with PHP error posted below.

### Expected result
1. Suggest with products list should be shown when product name or incomplete SKU was provided in "Enter SKU or Product Name" field (??).
2. There is no error in response for product search XHR request.

### Actual result
1. Error message is shown: "The SKU was not found in the catalog".
2. In response for `/quickorder/ajax/product_suggest/?q=joust` above call stack is returned.

```
( ! ) TypeError: Argument 3 passed to Magento\QuickOrder\Model\FulltextSearch::__construct() must implement interface Magento\Search\Api\SearchInterface, instance of Smile\ElasticSuiteCore\Model\Search given, called in /var/www/vendor/magento/framework/ObjectManager/Factory/AbstractFactory.php on line 111 in /var/www/vendor/magento/module-quick-order/Model/FulltextSearch.php on line 57
1  0.2007374848   {main}( ) .../index.php:0
2  0.377110523928 Magento\Framework\App\Bootstrap->run( ) .../index.php:37
3  0.382011155680 Magento\Framework\App\Http\Interceptor->launch( ) .../Bootstrap.php:256
4  0.382311161320 Magento\Framework\App\Http\Interceptor->launch( ) .../Interceptor.php:24
5  0.484517470888 Magento\Framework\App\FrontController\Interceptor->dispatch( ) .../Http.php:135
6  0.486217518344 Magento\Framework\App\FrontController\Interceptor->___callPlugins( ) .../Interceptor.php:26
[...]

PR #873, opened 2018-04-16
Change the implemented interface of the Elasticsuite Search Engine.

romainruaud:
This one is a pre-requisite for fixing #855 

This is just changing the implemented interface from `\Magento\Framework\Api\Search\SearchInterface` to `\Magento\Search\Api\SearchInterface`.

The `\Magento\Search\Api\SearchInterface` is an `@api` flagged one also, and is empty and just inheriting the `\Magento\Framework\Api\Search\SearchInterface` : https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/Search/Api/SearchInterface.php

So for me, this is not too much risky to inherit this interface instead of the previous one, since they are the same for now. Also, the new interface is part of the `Magento_Search` module but we already depend on it.

Regards

### DIFF
--- a/src/module-elasticsuite-core/Model/Search.php
+++ b/src/module-elasticsuite-core/Model/Search.php
@@ -20,7 +20,7 @@ namespace Smile\ElasticSuiteCore\Model;
  * @package  Smile\ElasticsuiteCore
  * @author    Aurelien FOUCRET <aurelien.foucret@smile.fr>
  */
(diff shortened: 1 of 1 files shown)


romainruaud:
It's also more logical to implement this interface since this is the one we declared on our config here : https://github.com/Smile-SA/elasticsuite/blob/master/src/module-elasticsuite-core/etc/di.xml#L65

Issue #853, opened 2018-04-06
The "[terms] query does not support [0] within lookup element" error when Layered Navigation elements have random keys in the URL


mcspronko:
### Preconditions
Google and Bing bots when crawling website use different combinations of a layered navigation filter. Somehow, bots can check the valid URL, however, the way bots generate these URLs might not work with the Layered Navigation logic (ElasticSearch) when keys of the array aren't going 0,1 but 0,2. ElasticSearch can't process such requests and as a result, no products are shown on a category page.

Magento Version : Magento Open Source 2.2.2

ElasticSuite Version :  2.5.0

Environment : Production and Developer

Third party modules : In-house modules for Function Score

### Steps to reproduce
These steps are for demo purpose:
1. Install Magento Open Source 2.2.2 with sample data
2. Navigate to a category with products where Color attribute shows at least 2 color values (exact values doesn't matter)
3. Select 2 color attributes from the layered navigation
4. Observe URI parameters (should be something like ?color%5B0%5D=353&color%5B1%5D=Navy)
5. Change the second key of the color parameter e.g. ?color%5B0%5D=353&color%5B2%5D=Navy in the URL
6. Observe the result

### Expected result
1. Products are shown in the category exactly same list as from step #5.
2. No errors in the system.log file

### Actual result
1. The error in the system.log file with the following information:

```
{"error":{"root_cause":[{"type":"parsing_exception","reason":"[terms] query does not support [0] within lookup element","line":1,"col":982}],"type":"parsing_exception","reason":"[terms] query does not support [0] within lookup element","line":1,"col":982},"status":400}
```

2. Also, the Request Body is logged where you can see keys for multi-select filter attribute:

```
    'terms' => 
    array (
      'option_text_type.untouched' => 
      array (
        0 => 'Belts',
        2 => 'Scarves',
        4 => 'Caps & Hats',
      ),
      'boost' => 1,
    ),
```

The full request body:
```
[...]

PR #875, opened 2018-04-17
Fix the remove urls

Asfolny:
Currently when the remove urls are generated, from 3 filters and upwards, the array difference is taken
Causing the query array to have keys with [0] and [2], which messes up elasticsearch and returns 0 results


afoucret:
Hi @Asfolny,

I recently merged a PR (#870) that fix issue #853 applying facet with arbitrary keys ([0] and [2]).
It will be released with version 2.5.8 (coming in few minutes).

I think it solves your issue in a different way you propose here.
Your PR is still valuable since it make  query params more predictable. I consider merging it and release with the next maintenance release (2.5.9).

Do you confirm your main problem is solved by PR #870 ?

src/module-elasticsuite-catalog/Model/Layer/Filter/Item/Attribute.php
```diff
@@ -52,6 +52,12 @@ public function getRemoveUrl()
             }
 
             $query = [$this->getFilter()->getRequestVar() => $resetValue];
+
```
afoucret: Think the code can be more concise if written this way :

```php
    public function getRemoveUrl()
    {
        $query = [$this->getFilter()->getRequestVar() => $this->getFilter()->getResetValue()];

        if (is_array($this->getApplyValue())) {
            $resetValue = array_values(array_diff($this->getApplyValue(), [$this->getLabel()]));
            if (count($resetValue) < 2) {
                $resetValue = current($resetValue);
            }

            $query = [$this->getFilter()->getRequestVar() => $resetValue];
        }

        $params = [
            '_current'     => true,
            '_use_rewrite' => true,
            '_query'       => $query,
            '_escape'      => true,
        ];

        return $this->_url->getUrl('*/*/*', $params);
    }
```

afoucret:
Hi @Asfolny,

See my review above.
Would also be nice if you could rebase to branch 2.5.x in order to release with version 2.5.9.

BR,

afoucret:
Hi @Asfolny,

Since you have rebased your PR all the new commits of the master branch in it.
I can't merge it while it is in this state.

Issue #885, opened 2018-04-19
Results in ElasticSearch, but not in Magento


edwinbos:
### Preconditions
Magento Version : 2.2.2
ElasticSuite Version : 2.5.6
Environment : Production

### Steps to reproduce
When I follow the ElasticSearch link http://example.com:9200/__search?q=barcode_main:079038257922_ I get 1 result, the expected product with the barcode: **079038257922**.

When I search for **079038257922** in Magento 2 I get zero results.

The attribute "barcode_main" is enabled for search, autocomplete, advanced search.

### Expected result
The same results as the actual ElasticSearch result in de search of Magento.

### Actual result
Zero results

edwinbos:
If you search on the name of the product (which corresponds to the bar code) I get the expected result. So the product is there, but can't be found by barcode.

Attribute configuration:

![image](https://user-images.githubusercontent.com/1267356/38992975-9a4412c2-43e2-11e8-92bb-7fc807b33082.png)

![image](https://user-images.githubusercontent.com/1267356/38993012-ad859676-43e2-11e8-920d-36796153018e.png)

afoucret:
What about frontend input ? Validation ?

afoucret:
I am almost sure there is a problem with the input validation of your attribute that cause it to be indexed like an integer. Can you please update your issue to provide required information.

BR,

afoucret:
What append if you configure the field as being used in spellcheck ?

afoucret:
Hi @edwinbos,

Any feedback ?

afoucret:
Hi @edwinbos,

Almost sure making your attribute "Used in spellcheck" would solve this issue.
I close but feel free to reopen if it does not.

PR #892, opened 2018-04-26
Fix numerical fulltext queries.

afoucret:
Fix for #885


afoucret:
Not a so good idea.
Abandoned.

PR #893, opened 2018-04-26
Add dependency to Magento_LayeredNavigationStaging for Smile_ElasticsuiteCatalog. 

### LINKED ISSUE #886: Filters in frontend disappear when you add a sequence to Smile_ElasticsuiteCatalog in a module
### Preconditions
Magento Version : _EE 2.2.2_
ElasticSuite Version : _2.5.8_
Environment : _Developer_

### Steps to reproduce
1. Create a new module
2. Add a sequence to **Smile_ElasticsuiteCatalog**

### Expected result
The new module works properly

### Actual result
The filters in the frontend, on the category page, disappear. 

### What is causing the issue
It is caused by the fact that **_Smile_ElasticsuiteCatalog_** gets loaded before **_Magento_LayeredNavigationStaging_**. This is due to the fact that **_Smile_ElasticsuiteCatalog_** has a sequence to **_Magento_LayeredNavigation_**, instead of  **_Magento_LayeredNavigationStaging_**.

**_Smile_ElasticsuiteCatalog_** has a virtualType that requires **_Magento_LayeredNavigationStaging_** to be loaded first for it to work :
```
<virtualType name="Magento\LayeredNavigationStaging\Block\Navigation\Search" type="Smile\ElasticsuiteCatalog\Block\Navigation">
    <arguments>
        <argument name="filterList" xsi:type="object">searchFilterList</argument>
    </arguments>
</virtualType>
```
You can find it in `frontend/etc/di.xml`

### How to solve the issue
Simply change the line `<module name="Magento_LayeredNavigation" />` in `Smile\ElasticsuiteCatalog\etc\module.xml` by `<module name="Magento_LayeredNavigationStaging" />`, so that **_Magento_LayeredNavigationStaging_**, which also loads **_Magento_LayeredNavigation_** is always loaded before **_Smile_ElasticsuiteCatalog_**.

afoucret:
Fix #886.

@romainruaud : can you test it please.

### DIFF
--- a/src/module-elasticsuite-catalog/etc/module.xml
+++ b/src/module-elasticsuite-catalog/etc/module.xml
@@ -24,6 +24,7 @@
             <module name="Magento_CatalogSearch" />
             <module name="Magento_LayeredNavigation" />
             <module name="Magento_CatalogInventory" />
+            <module name="Magento_LayeredNavigationStaging" />
         </sequence>
     </module>
 </config>


romainruaud:
Ok for me, I even tested a complete reinstall from scratch to ensure Magento allows to add & enable the module despite the hacky sequence.

It works like a charm.

Regards

PR #940, opened 2018-06-07
Ensure search weight is properly computed for SKU

### LINKED ISSUE #920: Search weight for SKU. Search Engin ignore weight of attribute

Please check an example:

romainruaud:
This one fix #920 .

It allow also merging for other configuration parameters from DB over the xml file. 

Let me know if we should only keep the search weight for now.


romainruaud:
I close this one, an ugly merge commit is trailing. Will open a new.

PR #941, opened 2018-06-07
Ensure search weight is properly computed for SKU

### LINKED ISSUE #920: Search weight for SKU. Search Engin ignore weight of attribute

romainruaud:
This one fix #920 .

It allow also merging for other configuration parameters from DB over the xml file. 

Let me know if we should only keep the search weight for now.


afoucret:
I am not an huge fan of how things are done.
I would prefer the Field to be created with accurate params than overriding them.

romainruaud:
Ok I will try to do it in another manner. Complex part is that SKU is not built by the `getFields()` of the attribute provider because it has a `static` backend.

Maybe enforcing the SKU to be indexed on this datasource could be enough.

I will try an let you know.

afoucret:
Closed because of the discussion above.

PR #962, opened 2018-06-19
Conditional appliance of optimizers.

romainruaud:
Finally I decided to rework what I initially had.

For now, the preview support in the BO is still missing, but I wanted to have your thoughts on the chosen implementation first : 

- previous caching of all optimizers rules remains intact
- I add an object which is able to filter optimizers if not applicable in current context.
- the cost is just an additional SQL query on runtime (no effect on category pages thanks to caching).

maybe the whole correlation map between optimizers and query_id/category_id could be loaded and cached only once to prevent this SQL query.

Let me know

### DIFF
--- a/src/module-elasticsuite-catalog-optimizer/Model/Optimizer/ApplierList.php
+++ b/src/module-elasticsuite-catalog-optimizer/Model/Optimizer/ApplierList.php
@@ -12,12 +12,9 @@
  */
 namespace Smile\ElasticsuiteCatalogOptimizer\Model\Optimizer;
 
-use Magento\Framework\App\CacheInterface;
 use Smile\ElasticsuiteCatalogOptimizer\Api\Data\OptimizerInterface;
-use Smile\ElasticsuiteCatalogOptimizer\Model\Optimizer\Collection\ProviderInterface;
 use Smile\ElasticsuiteCatalogOptimizer\Model\ResourceModel\Optimizer\Collection;
 use Smile\ElasticsuiteCore\Api\Search\Request\ContainerConfigurationInterface;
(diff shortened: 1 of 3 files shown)


src/module-elasticsuite-catalog-optimizer/Model/Optimizer/RelevantOptimizerList.php
```diff
@@ -0,0 +1,114 @@
...
+namespace Smile\ElasticsuiteCatalogOptimizer\Model\Optimizer;
+
+/**
+ * Return relevant optimizers for a given search context.
+ *
+ * @category Smile
+ * @package  Smile\ElasticsuiteCatalogOptimizer
+ * @author   Romain Ruaud <romain.ruaud@smile.fr>
+ */
+class RelevantOptimizerList
```
afoucret: The name of the class is not appropriate.
It suggest a that optimizer are selected using a relevance algorithm.

OptimizerList seems more accurate to me.

romainruaud: Ok for this one, just OptimizerList is fine for me too.

afoucret:
I did not audit the code completely but I think the caching mechanism should be reworked from the ground too. I will propose you something on this specific point.

romainruaud:
I renamed the class according to your thoughts. 

Let me know for the next steps.

afoucret:
I will merge this PR but rewrite the way you are filtering the optimizers.

Let you know when ready and will proceed to the final review of the feature on the accurate branch.

PR #963, opened 2018-06-20
Make the InventoryData resource model easier to extend/replace

nicobatty:
I wanted to rewrite parts of the `loadInventoryData` method but it seems like I can't do it easily without replacing the datasource because:

- I couldn't inherit the class or use a plugin because the class attributes and methods are all `private`.
- I couldn't switch the ResourceModel class to an other class because the construct parameter must be of InventoryData  resource model type, which I can't because of the first reason.

That is why I offer this solution to both of these problems if someone needs to inherit the current InventoryData resource model class or change the class entirely.

### DIFF
--- a/src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/InventoryData.php
+++ b/src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/InventoryData.php
@@ -13,7 +13,7 @@
 namespace Smile\ElasticsuiteCatalog\Model\Product\Indexer\Fulltext\Datasource;
 
 use Smile\ElasticsuiteCore\Api\Index\DatasourceInterface;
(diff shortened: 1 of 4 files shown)


romainruaud:
Actually I'm not sure about using `protected` variables since Magento's good practices are to use composition over inheritance.

I'll let @afoucret give you a more complete answer on this one.

But anyway, thank you for using our module and to contribute.

Regards

nicobatty:
Composition over inheritance does not mean composition only.

The idea here is that I can choose weither I prefer to use composition or inheritance depending on the situation. If I need to change most of the code, I can use composition. If I need to change just a tiny part of the code, I can inherit the class.

But I understand what you mean.

Here I cannot use any of the two without copy pasting the InventoryData Model and ResourceModel which is basically violating the DRY principle.

romainruaud:
Imho this is already covered in latest versions.

See https://github.com/Smile-SA/elasticsuite/blob/master/src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/InventoryData.php#L16

which is now using an interface.

Regards

PR #982, opened 2018-06-26
Fix special chars display in autocomplete.

afoucret:


romainruaud:
Tested OK for me.

I approve this PR.

PR #990, opened 2018-06-28
Fix relevance config in Magento 2.2.5.

afoucret:


afoucret:
@romainruaud : can you test this one against both Magento 2.2.4 and 2.2.5 (especially config and admin relevance edit) and report to me.

mageho:
Same here.

PR #992, opened 2018-06-28
Fix category indexing in Magento 2.2.5

afoucret:


mageho:
Thank you, it solved this issue in Magento 2.2.5

nicobatty:
Hi, I don't know if this is volontary or not but this update is only compatible with M2.2.5 since the `Magento\Catalog\Model\Indexer\Category\Product\TableMaintainer` only exists in M2.2.5.

See here :

M2.2.4 : https://github.com/magento/magento2/tree/2.2.4/app/code/Magento/Catalog/Model/Indexer/Category/Product
M2.2.5 : https://github.com/magento/magento2/tree/2.2.5/app/code/Magento/Catalog/Model/Indexer/Category/Product

Maybe it should be considered a version 2.6.x and specified in the documentation.

afoucret:
That is why I have used the ObjectManager.
But I have also used TrableMaintainer::class and it can't work.
My bad ...
I will fix it quickly.

nicobatty:
Oh yes I did not look at the entire code, I just looked at the "use" statement. ;)

afoucret:
Think we will try to release something Monday (Friday is not a good day for a release).

afoucret:
PR #993 should fix the backward compat.
Propose you to continue the discussion there.

Issue #571, opened 2017-10-18
Price scroll incorrect edge values


madonzy:
The values on the price scroll are incorrect on init.

### Preconditions

Magento Version : 2.1.9

ElasticSuite Version : 2.3.8

Environment : Developer

Third party modules : -

### Steps to reproduce

1. Open catalog/search
2. See price scroll edge values

### Expected result

1. Edge value to be limited to the lowest and highest prices of filtered products

### Actual result

1. The down edge price is smaller then the price of the cheapest product (see brutto value)
2. The up edge price is higher then price of the most expensive product (not seen in screen)

**The products on screen are sorted by price from small to high.**
![screenshot-2](https://user-images.githubusercontent.com/5281020/31715837-a0e0c6a6-b405-11e7-94ab-45732b4bac45.png)

afoucret:
Are you using tier pricing ?

madonzy:
@afoucret no, normal pricing. The value I put in admin panel is "brutto" (ex. 699,00 for the first product)

afoucret:
Are you applying some post-filtering after the collection is loaded (in order to hide some products by example) ?

madonzy:
@afoucret no, I load all products that apply values of filters. There's no extra functionality

romainruaud:
@madonzy if I understand well, you have products on the page which have a "brutto" price which is >993 ?

If yes, can you please post here the same elements as in #557: 

- what you have configured in the Back-office under **Sales > Tax > Price Display Settings** (for each website if this does differ)
- the price which is set for your product, for each website (if the price differs across websites).
- the entry of the ```catalog_product_index_price``` table for your products.

Regards

romainruaud:
Closed due to being inactive/no-answers for 3 weeks.

Feel free to reopen it if you are able to provide more details.

madonzy:
@romainruaud 
price display settings are: 
<img width="1012" alt="screen shot 2017-11-21 at 3 05 16 pm" src="https://user-images.githubusercontent.com/5281020/33077046-f468e068-cece-11e7-9322-9200772c2345.png">

For the first product (when listing is sorted from low to high price):
price is set the same for each store = 699

catalog_product_index_price for this product is:
<img width="1019" alt="screen shot 2017-11-21 at 3 14 56 pm" src="https://user-images.githubusercontent.com/5281020/33077166-54ed6a62-cecf-11e7-8a64-314e929efd87.png">

nicobatty:
[...]

PR #1003, opened 2018-07-09
Add default tax class rate to the price index

nicobatty:
Hi!

This PR is related to this issue: #571

This is an attempt at fixing the price index so that it includes the default store tax.

In order to recover this tax, I added the `tax_class_id` to the price index request to limit the performance impact.

I'm not yet sure about the `$this->taxCalculation->getCalculatedRate` call in terms of performance, though it seems like the rate recovery is cached:
https://github.com/magento/magento2/blob/50f974b0c8f3ada8d7a2c9b0f9924976dacd2088/app/code/Magento/Tax/Model/Calculation.php#L362-L378

### DIFF
--- a/src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/PriceData.php
+++ b/src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/PriceData.php
@@ -14,8 +14,11 @@
 
 namespace Smile\ElasticsuiteCatalog\Model\Product\Indexer\Fulltext\Datasource;
 
+use Magento\Framework\Pricing\PriceCurrencyInterface;
 use Smile\ElasticsuiteCore\Api\Index\DatasourceInterface;
 use Smile\ElasticsuiteCatalog\Model\ResourceModel\Product\Indexer\Fulltext\Datasource\PriceData as ResourceModel;
+use Magento\Tax\Api\TaxCalculationInterface;
+use Magento\Tax\Model\Config;
 
 /**
  * Datasource used to append prices data to product during indexing.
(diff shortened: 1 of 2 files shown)


nicobatty:
@romainruaud I submitted this PR a while ago and I am wondering what you think about it.

Is there anything I can fix?

romainruaud:
Hello @nicobatty 

to me it's quite unclear what you are trying to achieve here. 

I see that you calculate on-the-fly the product prices according to their tax class id, but I do not understand the reason behind this.

Maybe you can give me more details about it. Also, it may be due to a custom tax configuration by your side, so would you mind posting a screenshot of your tax configuration in the Magento Back-Office ? I suppose there is something related with the way to display product price (incl or excl tax), and things like this.

Best regards

nicobatty:
Hi @romainruaud,

Sure no problem I can explain this to you.

Basically, in Magento you have two configurations:

* There is a configuration to specify if the catalog price exclude or include the tax (the price set in the admin) : Stores > Configuration > Sales > Calculation Settings > Catalog Prices.
* There is a configuration to specify which country, state, post code to use as a default tax calculation for the store, so for example, in France we woud set it to `France, *, *` : Stores > Configuration > Sales > Default Tax Destination Calculation.

Magento by default is using the "Including Tax" for the "Catalog Prices" config, in this case the `catalog_product_index_price` table contains the prices including the tax but when is it set to "Excluding Tax", it will contains the price excluding the tax.

the issue is that the Price indexer for ElasticSuite is taking the raw price in the `catalog_product_index_price` which is not the actual displayed price on the frontend if you take the Default Tax Destination Calculation setting into consideration.

It might be easier to explain with an example:

1. I set the Catalog Prices tax config to "Excluding Tax".
2. I set the default destination config to `France, *, *`.
3. I create a new tax class **Apparel** with 20% tax for `France, *, *`
4. I create a product with a tax class of **Apparel** and a price of **20**.
5. We can see the product on our store appears with a price of 24€.
6. If we check on the `catalog_product_index_price` table, the price is set at 20 (untaxed).

This is what Magento does basically.

The issue is that the current implementation of the price indexer for ElasticSuite is taking the raw `catalog_product_index_price` information, so in this example, it will set the product price as 20 in the index, which is wrong since the price displayed on the store is actually 24.

cmuench:
@nicobatty @romainruaud We have the same problem here in Germany. Looks like a good idea to index the prices with tax if possible for France and Germany.

kevinvuillemin:
Hello,

@nicobatty  Did you find au solution please ? We have this problem too with multi taxes. 


https://github.com/Smile-SA/elasticsuite/issues/1489

nicobatty:
Hi @kevinvuillemin,

I no longer work on Magento projects but this PR should do the trick.

At least it used to work back when I did this change and it was used on a multistore / multitax website that was doing B2C and B2B.

All it really does is add the `tax_class_id` to the request for the regular price, recover the tax rate and calculate the taxed price.

I think you are probably better off patching these changes since it will probably never going to be merged.

The use case is specific to those using untaxed price by default and @romainruaud don't like the fact that the price is being recalculate here. The problem is, it is not calculate anywhere otherwise we wouldn't be using the price index as a reference.

I explained the issue in detail on this PR so if you want to patch it in an other way you are free to do it.

PR #1053, opened 2018-08-31
Add german translations

amenk:


afoucret:
Hi @amenk , @replacementchicken,

I see this PR is still WIP.
As we don't have any german skills, can you please notify us when you consider the job done ? 
Then I will be able to merge your PR.

Thanks for your contrib that should be very appreciated since we know we have a very strong german community.

afoucret:
Thank your for the contrib guys.
It will be part of release 2.6.1.

PR #1055, opened 2018-08-31
Refactoring mapping field building and merging with config files.

### LINKED ISSUE #920: Search weight for SKU. Search Engin ignore weight of attribute

romainruaud:
Several things here : 

- Mapping class refactored to allow only an array of fields in constructor.

- Fields are built by the Indices Config prior to creating the mapping.

- Fields that exist both in a "dynamic provider" and from the elasticsuite_indices.xml file are now properly merged.

I made the choice to give precedence to the file : only some fieldConfig is written to the file, meaning it will only overwrite some values of the dynamically computed Field.

For SKU, this allow to have the "defaultSearchAnalyzer" defined in the file, and to rely on the Back-Office for Search Weight of the attribute.

Fix #920


romainruaud:
I also refactored the Unit tests for Mapping since it does not rely anymore on Dynamic Fields Provider : easier to test now. But the Indices Config part remains untested about DynamicFieldProviders.

let me know

src/module-elasticsuite-core/Index/Mapping/Field.php
```diff
@@ -225,6 +225,14 @@ public function getDefaultSearchAnalyzer()
         return $this->config['default_search_analyzer'];
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    public function getConfig()
```
afoucret: Think it is a mistake to create a new getter for the config.
I would prefer a method mergeConfig(array $config) that would return a new instance of the field with updated config.

It is also possible to add a test on the mergeConfig method in FieldTest.

src/module-elasticsuite-core/Index/Indices/Config.php
```diff
@@ -154,6 +145,40 @@ private function initIndexConfig(array $indexConfigData)
...
+        $dynamicFieldProviders = array_filter($dataSources, [$this, 'isDynamicFieldsProvider']);
+
+        $fields = [];
+
+        foreach ($dynamicFieldProviders as $dynamicFieldProvider) {
+            $fields += $dynamicFieldProvider->getFields();
+        }
+
+        foreach ($typeConfigData['mapping']['staticFields'] as $fieldName => $fieldConfig) {
+            if (isset($fields[$fieldName])) {
```
afoucret: Use mergeConfig method instead of this one :

```php
$field = $this->mappingFieldFactory->create(['name' => $fieldName] + $fieldConfig);

if (isset($fields[$fieldName]) {
    $field = $fields[$fieldName]->mergeConfig($fieldConfig);
}

$fields[$fieldName] = $field;
```

romainruaud:
PR updated and rebased

PR #1063, opened 2018-09-06
Fix cache policy

romainruaud:
- Add a cache tag to Magento popular search results and allow cleaning it after full reindex.
- Purge all virtual categories parent categories tree after full reindex.

I can amend this PR if it's not enough, let me know


romainruaud:
I added cache policy for optimizers (fine-grained according to their categories / search terms).

Next step : 

- Thesaurus cache
- Manual merchandising on search term (if the term is among popular results).
- maybe more ?

romainruaud:
Thesaurus cache is working OK.

Issue #947, opened 2018-06-11
Manual sort order in category does not work


OvalMedia:
Configuring products' manual sortorder inside a category does work in frontend.

### Preconditions
Linux, Apache, MariaDB

Magento Version : 2.2.4 (CE)

ElasticSuite Version : 2.5 (composer says ^2.5 and no newer updates available) Otherwise I can't tell the version.

Environment : developer

Third party modules :

### Steps to reproduce

1. create a category
2. add products to the category
3. enable manual sort for those products

### Expected result

1. products in frontend show according to manual sort order

### Actual result

1. products are sorted by their ID instead

I see the desired sort orders are being saved to the table smile_virtualcategory_catalog_category_product_position. However they are NOT saved to the table catalog_category_product, I am not sure that this would be necessary though.

romainruaud:
Could you please use `composer info | grep smile` to see the exact version of elasticsuite ?

This bug is looking like this one : https://github.com/Smile-SA/elasticsuite/pull/936

And it has been fixed in latest Elasticsuite version (2.5.11)

Let us know,

Regards

OvalMedia:
That command returns "2.5.11"
I had seen that bug and checked the files for those changes (they are there).

romainruaud:
I am not able to reproduce your issue on elasticsuite 2.5.11 and Magento 2.2.4

Are you sure that : 

- the cronjob has been ran if you are in "update on schedule" for index management ?
- the sort order is set to "position" in frontend for your category ?

Regards

afoucret:
As @romainruaud explained we are not able to reproduce the issue.
Can you be more specific on detail that may allow us to reproduce it.

OvalMedia:
Sorry, I had to deinstall that module now. It may be possible that some other module was interfering with the functionality but I currently don't have the time to dig deeper. I might get back to you later.

afoucret:
As we were not able to reproduce and you can't give us more detail I close this one.
I suspect a third party module to cause the bug.

We are always sad to see people that can't use our module. So if you want to give it a try later and can provide us more detail, feel free to reopen this issue.

BR,

robmacecs:
Hey @romainruaud @afoucret 

I didn't want to open a new ticket because I suspect ours might be 3rd party module related also. But we have this exact behaviour, saving a category with manual sort, the updated order doesn't show up in the FE until you manually reindex via CLI. 

[...]

PR #1064, opened 2018-09-07
Fix manual sort position not reindexing on index by schedule

nicobatty:
Fix for: #947

This fix schedule an update on the `catalogsearch_fulltext` index for any products updated with manual sort since it seems to be depending on it.

I am not totally sure why the manual sort depends on this index but I guess someone could change that.


romainruaud:
@nicobatty please do it inside the virtual category module (since it's the one which is responsible for creating this table).

Regards

PR #1076, opened 2018-09-12
Fix manual sort position not reindexing on index by schedule

smuzza:
Update pull request #1064 by adding mview subscription to module-elasticsuite-virtual-category module instead.

Credit @nicobatty


romainruaud:
replaced by #1115 to be merged on 2.6.x instead of master.

PR #1077, opened 2018-09-12
Fix for: RelevanceConfig isn't saving to DB

pczerkas:
RelevanceConfig changes in admin aren't saving to DB. This pull request fixes this.


romainruaud:
Hello @pczerkas 

how are your saving the relevance config ? Just by clicking save on the back-office ? or you are doing it programmatically ?

pczerkas:
Hello @romainruaud 

I've tried to save relevance config from the backend (it didn't work). But the fix is rather straightforward: it adds missing parameter for child class constructor?

romainruaud:
Replaced by #1137 

Magento constructor differs **between two minor versions 2.2.5 and 2.2.6** and the order for additional parameters in constructor is inconsistent ...

See https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/Config/Model/Config.php

They introduced `$settingsChecker` before `$data` which makes us unable to have a proper backward-compatible `parent::__construct()` call with the `$data` parameter inside.

PR #1116, opened 2018-10-05
Add "To Top" button to the manual sort category listing

nicobatty:
**How it looks:**

to-top

**The reason why:**

I added this feature on our project because we had a lot of products in manual sort.

When you add a product as Manual sort, it gets to the end of the Manual sort product list and using the drag and drop to get a few products to the top is not really practical and can take some time, especially for a user that is not really good with a mouse.

**Technical Details:**

I did it and tested it in a few hours because I did not have too much time to spend on this so it might not be perfectly coded. I tried to reuse the API and logic that was currently available but the code could clearly be improved / refactored.

**What's missing:**

* Hide the button for the element at the top of the list (that is not really a problem though).

### DIFF
--- a/src/module-elasticsuite-catalog/view/adminhtml/web/css/source/_module.less
+++ b/src/module-elasticsuite-catalog/view/adminhtml/web/css/source/_module.less
@@ -141,6 +141,25 @@
                 }
             }
 
+            .totop-handle {
+                position: absolute;
+                top: 30px;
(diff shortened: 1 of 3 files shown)


romainruaud:
Seems great ! 

I'll try to have a look on this one and to test it deeply.

But it's definitely a big THANKS for contributing to our Open Source project :)

PR #1125, opened 2018-10-11
Display correct count in price slider if range ‚from value‘ = ‚to value‘

dmcanna:
f.e. we have products in shop with price 100$. If we make in price slider "from value" = "to value", we got "No products in the selected range." message. Also if we open direct url https://some_shop/some_url?price=100-100 we didn't see the products with price 100$

### DIFF
--- a/src/module-elasticsuite-catalog/Model/Layer/Filter/DecimalFilterTrait.php
+++ b/src/module-elasticsuite-catalog/Model/Layer/Filter/DecimalFilterTrait.php
@@ -52,7 +52,7 @@ trait DecimalFilterTrait
 
                 $this->getLayer()->getProductCollection()->addFieldToFilter(
                     $this->getAttributeModel()->getAttributeCode(),
-                    array_filter(['gte' => $fromValue, 'lt' => $toValue])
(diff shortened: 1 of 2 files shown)


dmcanna:
this is also related to the fix https://github.com/Smile-SA/elasticsuite/commit/b647b03191dc9c5c7721463ad429f1c35dd71cd7 and issue https://github.com/Smile-SA/elasticsuite/pull/381

rbayet:
While the issue does exist when min = max in the slider, changing the behavior of the filter would lead to other problems, because the slider is built around an histogram which takes the floor() of a price.
So, considering you have in the middle of your slider four products with actual prices 68.0, 68.30, 69.0 and 69.30, the corresponding histogram slots will be :
- 68 = 2
- 69 = 2

At the moment, if you put min on 68 and max on 69, the number of available products will be displayed as 2 (because of the lt).
And the resulting page, on a price=68-69 will also display 2 products because of the (lt).

Now, if we adopt your changes, and if you put min on 68 and max on 69, the number of available products displayed will be 4 (because of the lte).
But on the resulting page, you will only see 3 products, because ['gt' => 68, 'lte' => 69] *does not* include 69.30.

PR #1131, opened 2018-10-12
fix styles

Karlasa:
fix styles to fit more with M2 variables/syntax


romainruaud:
Hello @Karlasa,

nice job. 

I'm wondering if all the variables you use are properly defined on the "blank" theme ? It's mandatory to ensure they exist not only in Luma theme but also on blank, to ensure people are able to start integrating from blank theme if needed.

Regards

Karlasa:
Yes they do exist

romainruaud:
Ok, I double checked by myself. 

Approved. Thank you for contributing and helping us keeping the frontend styles cleaner :)

PR #1132, opened 2018-10-13
update category template

Karlasa:
added '>' element before last cat title
removed product qty element as it's not relevant for category name (is empty)


romainruaud:
Hello @Karlasa 

great job, and thank you for contributing. I do not understand the "added > element before last cat title". I do not think it was actually missing, or maybe I dot not understand well ?

Can you post a screenshot of the resulting design ? Regards

Karlasa:
sry was not paying atention on luma design anyways idea is that if anyone would like to show category results in one  line so it would be like breadcrubs > result cat name. It is kind of impossible as its added in template as ".join(' > ')" also it still generates html element breadcrubs even if result is root cat name without breadcrubs. Im sure there should be some way to make it more dynamic and not that hard coded but not sure what should be the right solution. 
Keep up the good work.

Cheers,

romainruaud:
Replaced by #1152

PR #1137, opened 2018-10-16
Fix relevance config save.

romainruaud:
Replace #1077


romainruaud:
@anakinsolo yes it's normal. It will be in 2.6.3.

Thank you.

PR #1139, opened 2018-10-17
Remove regular price span to show price label

DanieliMi:
Currently when searching it is not possible to show the price label. If you perform a search an the autocomplete shows products and their prices these prices are not labeled. Is this behavior intended?

In our case we have many products with tier prices and in the autocomplete will be shown the regular price and the tier price but you cannot distinguish them so we want the "As low as" label to be shown for the tier price. But due to the `<span class="regular-price">` it's not displayed by magento2 css.
Basically, prices contained in the template are not always regular prices, they can be other prices as well
so this class "regular-price" does not make sense at this point.

### DIFF
--- a/src/module-elasticsuite-catalog/view/frontend/web/template/autocomplete/product.html
+++ b/src/module-elasticsuite-catalog/view/frontend/web/template/autocomplete/product.html
@@ -7,7 +7,7 @@
             <div class="product-primary"><div class="product-name"><%= data.title %></div></div>
             <div class="product-secondary">
                 <div class="price-box">
-                    <span class="regular-price"><span class="price"><%= data.price %></span></span>
+                    <span class="price"><%= data.price %></span>
                 </div>
             </div>
         </div>


romainruaud:
Hello @DanieliMi 

for me it's actually unclear what is doing your PR.

Would you mind attaching a screen before/after your modification so that I'm able to understand better what you are trying to achieve here ? 

Anyway, thank you for contributing to our module

DanieliMi:
Hey @romainruaud 
of course I will explain it better.
Before the PR the tier price of an product is displayed like this:
smile_before
After the PR it's going to look like this:
smile_after

Basically the "As low as" label for tier prices will be displayed again. The problem why it's not displayed lays in the line of code I removed. The product price is wrapped in `<span class="regular-price">` even tho logically it's not a regular price it's a tier price and because of that the magento luma theme will not display the label. The styling that is responsible for this can be found [here](https://github.com/magento/magento2/blob/fcf9eed8a2456a88d36cfcca02f6a35ad81d6972/app/design/frontend/Magento/luma/Magento_Catalog/web/css/source/module/_listings.less#L131)

romainruaud:
Ok ! thank you for the explanations.

Seems legit to me, I'll double check on my dev env but I'll probably merge this one.

Thank you for contributing !

romainruaud:
OK for me. I just replaced it by #1154 to be based on 2.6.x instead of master.

Thank you for contributing.

PR #1141, opened 2018-10-17
Fix missing assignment

pczerkas:
It looks like there is a missing assignment.


romainruaud:
Replaced by #1155 to be merged on 2.6.x instead of master.

Thank you

PR #1150, opened 2018-10-24
Add Dutch language pack to catalog module

Tjitse-E:


romainruaud:
Hello @Tjitse-E 

Wow, such a great effort, many thanks for contributing it.

I really appreciate this contribution, but I'm not a huge fan of having a partially translated module so ... you may have already guessed my question : do you plan to translate the other modules as well ? (I think Catalog is the biggest)

Issue #667, opened 2017-12-21
Error creating category in admin: catalog_product index does not exist yet. Make sure everything is reindexed.


careys7:
### Preconditions

Magento Version : EE 2.2.1

ElasticSuite Version : ^2.4.0 + Retailer

Environment : Developer

Third party modules : None

### Steps to reproduce

1. Admin > Catalog > Categories
2. Click on Root Category
3. Click on "Add Subcategory"

### Expected result

1. Category create form appears 

### Actual result

1. The POST request to the following URL returns HTTP 500:  https://domain/admin/virtualcategory/category_virtual/preview/ajax/1/key/351315ec7a4bec47d0aad1927ddbc1d91a6567a06af43c9a45f4dd9788ad02ed/?isAjax=true

Stack Trace:

```
1 exception(s):
Exception #0 (LogicException): catalog_product index does not exist yet. Make sure everything is reindexed.

Exception #0 (LogicException): catalog_product index does not exist yet. Make sure everything is reindexed.
#0 /var/www/current/src/vendor/smile/elasticsuite/src/module-elasticsuite-core/Search/Request/ContainerConfiguration.php(161): Smile\ElasticsuiteCore\Index\IndexOperation->getIndexByName('catalog_product', 0)
#1 /var/www/current/src/vendor/smile/elasticsuite/src/module-elasticsuite-core/Search/Request/ContainerConfiguration.php(94): Smile\ElasticsuiteCore\Search\Request\ContainerConfiguration->getIndex()
#2 /var/www/current/src/vendor/smile/elasticsuite/src/module-elasticsuite-core/Search/Request/Builder.php(150): Smile\ElasticsuiteCore\Search\Request\ContainerConfiguration->getIndexName()
#3 /var/www/current/src/vendor/smile/elasticsuite/src/module-elasticsuite-catalog/Model/ResourceModel/Product/Fulltext/Collection.php(583): Smile\ElasticsuiteCore\Search\Request\Builder->create(0, 'catalog_view_co...', 0, 0, NULL, Array, Array, Array, Array)
#4 /var/www/current/src/vendor/smile/elasticsuite/src/module-elasticsuite-catalog/Model/ResourceModel/Product/Fulltext/Collection.php(189): Smile\ElasticsuiteCatalog\Model\ResourceModel\Product\Fulltext\Collection->loadProductCounts()
[...]

PR #1164, opened 2018-10-30
WIP: Fixes #1153 Reducing list of filterable attributes

### LINKED ISSUE #1153: Cant save Category in large environment (Big Post)
### Preconditions

Magento Version : 2.2.6 and 2.3

ElasticSuite Version : 2.6.x-dev

Environment : Developer

Third party modules :

### Steps to reproduce

1. have a lot of categories with attributes. 
2. try to save them
3. get error of invalid form key. but its actually an other error.  he isnt actually submitting the the form. its posting to `/admin/catalog/category/validate/` with alot of post entities [https://pastebin.com/XTjx9dAR]
4. i believe the error is in the php setting `max_input_vars` but increasing those values doesnt seem to be an option in my opinion. 

### Expected result

1. save the category.
2. post the form without all the `facet_config[][]` entities if they havent changed. (i dont think its nessecary to make such a big post for every time the category needs to be changed)

### Actual result

1. Post Request: https://pastebin.com/XTjx9dAR
2. 
screenshot from 2018-10-25 09-28-58
screenshot from 2018-10-25 09-29-16
screenshot from 2018-10-25 09-29-43
screenshot from 2018-10-25 09-38-56

rbayet:
to those relevant for the currently assigned products


rbayet:
To test fully on a multi-store environment AND in category creation mode.
Currently a resurgence of #667 in EE 2.2.5 + multi-store environment + category creation mode.

src/module-elasticsuite-catalog/Plugin/Ui/Category/Form/DataProviderPlugin.php
```diff
@@ -116,6 +137,66 @@ private function getFilterableAttributeList($currentCategory)
...
+     * @return QueryInterface
+     */
+    private function getQueryFilter(CategoryInterface $category)
+    {
+        $query = null;
+
+        $category->setIsActive(true);
+
+        if ($category->getIsVirtualCategory() || $category->getId()) {
+            $query = $category->getVirtualRule()->getCategorySearchQuery($category);
```
romainruaud: This will fail if VirtualCategory module is disabled. Can you provide a fallback to a legacy category query in this case ? Based on category.category_id field.

src/module-elasticsuite-catalog/Plugin/Ui/Category/Form/DataProviderPlugin.php
```diff
@@ -132,6 +213,25 @@ private function getAttributes($category)
             ->addStoreLabel($category->getStoreId())
             ->setOrder('position', 'ASC');
 
+        $storeId = $this->getStoreId($category);
+
+        if ($storeId) {
+            /** @var \Smile\ElasticsuiteCatalog\Model\ResourceModel\Product\Fulltext\Collection $fulltextCollection */
+            $fulltextCollection = $this->fulltextCollectionFactory->create();
+            $fulltextCollection->setStoreId($storeId)
+                ->setVisibility([Visibility::VISIBILITY_IN_CATALOG, Visibility::VISIBILITY_BOTH]);
```
romainruaud: Probably not needed since it's default filter. You may want to replace it by injection of `Smile\ElasticsuiteCore\Api\Search\ContextInterface` and set properly the current Category and Store. It should then process the default filtering according to selected store and category : visibility, stock status, etc ...

src/module-elasticsuite-catalog/Plugin/Ui/Category/Form/DataProviderPlugin.php
rbayet: I could, but the list of attributes would be inherently wrong in this case.
So far I haven't found any code in VirtualCategory to delegate the query building to.
I need to identify more properly the way things are done there because it's not that obvious.

romainruaud: Something like : https://github.com/Smile-SA/elasticsuite/blob/master/src/module-elasticsuite-catalog-optimizer/Model/Optimizer/Preview/CategoryQuery.php#L72

src/module-elasticsuite-catalog/Plugin/Ui/Category/Form/DataProviderPlugin.php
rbayet: Indeed, the correct Category and Store are assigned to the ContextInterface.

romainruaud:
Seems great, a way less code than first version !

PR #1169, opened 2018-11-02
Dropdown attribute values not displaying in adminhtml

justinharris1986:
I migrated from 1.9.2.2 to 2.2.6, and when I go to stores -> Attributes -> Product, and select a attribute with a dropdown type, I see no attribute values in there. For example, if I had an attribute color, I would not see any migrated Values of "Blue", "Red", etc under "Manage Options".
Creating a new attribute with dropdowns would yield the same behavior.
Any new or existing values would appear in the front end and the product pages, so I know they were saving.
If I add any new values, It would save, but upon refresh would disappear.

Creating a whole new attribute had the same issue.

I however, could not delete any values, because they were not populating.

Disabling Smile_ElasticsuiteCatalog fixes the issue.

Commit 1bcebbe did not appear to fix the issue, just cleaning up noise in the browser Javascript console.

Commit 489cfce ultimately fixed the issue, but it smells pretty bad. When doing a diff of the product attribute page with Smile_ElasticsuiteCatalog enabled and then disabled, I noticed this code missing, so I stuffed it in here. I am sure there is a better fix.

### DIFF
--- a/src/module-elasticsuite-catalog/view/adminhtml/templates/catalog/product/attribute/js.phtml
+++ b/src/module-elasticsuite-catalog/view/adminhtml/templates/catalog/product/attribute/js.phtml
@@ -66,7 +66,7 @@
                 ($('frontend_input').value == 'select'
                 || $('frontend_input').value == 'multiselect'
                 || $('frontend_input').value == 'price'
-                || $('frontend_input').value == 'text')
+                || $('frontend_input').value == 'text'
                 || $('frontend_input').value == 'boolean')
             ) {
                 if($('is_filterable') && !$('is_filterable').getAttribute('readonly')){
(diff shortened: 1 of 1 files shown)


romainruaud:
Hello,

thank you for reporting it.

However, it's more complex than the fix you provided here. In fact Magento moved a lot of code from the `js.phtml` file to a real javascript component. See : https://github.com/magento/magento2/commit/2f057f74a651db2ee5bf689fdb0c5b66ac61489f#diff-5ab37ddfe3a4d9011d97bb9f7e4375bf

We did not notice this point, so thank you for pointing us to this issue.

I keep your PR open as a reminder to fix it on a nice way, but will not merge your solution as is.

Regards

rbayet:
Hi @justinharris1986,

Your PR has been superceded by #1191 which has been merged into 2.6.x and slated for release in 2.6.x

Regards

Issue #157, opened 2016-10-04
RFC : a search inside the virtual category product preview to refine the selection


djenvert:
When you're previewing products when building a virtual category, it would be nice to have a search widget to be able to refine for more specific products inside the products list of the preview (instead of scrolling of using "show more"), so that manually sorting a very specific product is easier with big catalog.

afoucret:
Hi @djenvert,

Such a feature must be designed to stay usable together with the sorting.

It is quite uneasy to sort a filtered subset of products and figure out what will append when the full list will be present.

At this time, I would propose :
- adding the fulltext search
- disable the drag and drop sort feature when the list is filtered
- keep switch to select automatic / manual sorting

This way, the fulltext search will allow to select products that have to be pinned or not more easily.

What do you think about such a solution ?

southerncomputer:
Yeah it would be nice to have full-text with facet search on backend !

djenvert:
It's not totally what I'm asking for (or rather what a customer of ours with a very big catalogue is asking for), @southerncomputer ! But yes, that could another request!
@afoucret I understand it's quite difficult. I guess the "manually sorted" ones will always be the one at the top of the list, won't they? So perhaps would it be possible to "freeze" them and only filter the "automatically sorted"? But it's getting complicated... So your proposal suits me.

rikwillems:
@afoucret I'm working on a POC for this as we need it here too.

rbayet:
Implemented in PR #1255 (based on PR #1183)

djenvert:
Thanks!

PR #1183, opened 2018-11-14
POC search in category product preview

rikwillems:
Proof of concept for https://github.com/Smile-SA/elasticsuite/issues/157

Adds a search box above the product list that allows to search through products to set them to manual sorting.

### DIFF
--- a/src/module-elasticsuite-catalog/Model/ProductSorter/AbstractPreview.php
+++ b/src/module-elasticsuite-catalog/Model/ProductSorter/AbstractPreview.php
@@ -64,7 +64,7 @@ abstract class AbstractPreview implements PreviewInterface
      * @param ItemDataFactory          $itemFactory       Preview item factory.
      * @param QueryFactory             $queryFactory      ES query factory.
      * @param integer                  $storeId           Store id.
-     * @param number                   $size              Preview size.
+     * @param integer                  $size              Preview size.
      * @param string                   $search            Preview search.
      */
     public function __construct(
--- a/src/module-elasticsuite-catalog/Model/ProductSorter/AbstractPreview.php
+++ b/src/module-elasticsuite-catalog/Model/ProductSorter/AbstractPreview.php
@@ -43,6 +43,11 @@ abstract class AbstractPreview implements PreviewInterface
     private $size;
 
     /**
+     * @var string
+     */
+    private $search;
+
+    /**
      * @var integer
      */
     private $storeId;
@@ -60,19 +65,22 @@ abstract class AbstractPreview implements PreviewInterface
      * @param QueryFactory             $queryFactory      ES query factory.
      * @param integer                  $storeId           Store id.
      * @param number                   $size              Preview size.
+     * @param string                   $search            Preview search.
      */
     public function __construct(
         ProductCollectionFactory $collectionFactory,
         ItemDataFactory $itemFactory,
         QueryFactory $queryFactory,
         $storeId,
-        $size = 10
+        $size = 10,
+        $search = ''
     ) {
         $this->collectionFactory = $collectionFactory;
         $this->itemFactory       = $itemFactory;
         $this->queryFactory      = $queryFactory;
         $this->storeId           = $storeId;
         $this->size              = $size;
+        $this->search            = $search;
     }
 
     /**
@@ -147,7 +155,10 @@ abstract class AbstractPreview implements PreviewInterface
      */
     private function getUnsortedProductData()
     {
-        $productCollection = $this->getProductCollection()->setPageSize($this->size);
+        $productCollection = $this->getProductCollection()
+            ->setSearchQuery($this->search)
+            ->setPageSize($this->size)
+        ;
 
         return ['products' => $productCollection->getItems(), 'size' => $productCollection->getSize()];
     }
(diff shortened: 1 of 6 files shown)


romainruaud:
Nice ! Thank you @rikwillems 

@rbayet can you have a look on this one and see if it's working properly ? 

Especially : 

- how it's dealing with this kind of "mixed query" between a category filter and a fulltext query
- general UX concerns like "I search for a term, merchandize a product, and reset the search, try another query, merchandize a second product", etc ...

@rikwillems do you plan adding more commits to this one or was it just meant to be a kickstart that you'd like us to finish ?

Regards

rikwillems:
@romainruaud you're welcome!

It's a kickstart for our own need and it appears to be working well at first sight. Our business is testing it as we speak. I hope to hear from them by monday. 

As it's a POC it has rough edges. I'm looking forward to your findings. From there we can complete it together.

romainruaud:
Thanks @rbayet for reviewing this one.

I agree that : 

- hitting enter should be enough to trigger the search
- there should be a way to reset the search easily instead of emptying the field manually
- code point of view : maybe the `->setSearchQuery($this->search)` should only be called if query string if not `''`
- I see a commented out line which could be removed

@rikwillems can you handle this by your side ? If not the case, I think we'd be able to do this, but cannot engage on any priority since for now we are focused on the next 2.7.0 major release to handle compatibility with Magento 2.3

Regards

src/module-elasticsuite-catalog/Model/ProductSorter/AbstractPreview.php
```diff
@@ -147,7 +155,10 @@ private function getProductCollection()
      */
     private function getUnsortedProductData()
     {
-        $productCollection = $this->getProductCollection()->setPageSize($this->size);
+        $productCollection = $this->getProductCollection()
+            ->setSearchQuery($this->search)
```
romainruaud: Shouldn't we apply `setSearchQuery` only if `$this->search !== ''`` ?

src/module-elasticsuite-catalog/view/adminhtml/web/js/form/element/product-sorter.js
```diff
@@ -155,6 +157,11 @@ define([
             return this.products().length < this.countTotalProducts();
         },
 
+        searchProducts: function () {
+            // this.currentSize(this.pageSize);
```
romainruaud: Dead code :)

src/module-elasticsuite-catalog/view/adminhtml/web/template/form/element/product-sorter.html
```diff
@@ -17,6 +17,11 @@
     </div>
     
     <div data-bind="if: hasProducts()">
+
+        <p>
+            <input type="text" name="search" id="elasticsuite-preview-search" class="admin__control-text"/>
+            <a href="#" data-bind="click: searchProducts, text: messages.search" class="action-default"></a>
```
romainruaud: Should trigger when ENTER is hit without needing to click : maybe the button can be removed, and replace by a "reset" button in case a search is active.

src/module-elasticsuite-catalog/Model/ProductSorter/AbstractPreview.php
rbayet: Possibly, but further down the line, the fulltext query part is not generated if `$this->search !== ''``

src/module-elasticsuite-catalog/view/adminhtml/web/js/form/element/product-sorter.js
rbayet: @romainruaud If you only really mean unnecessary commented code (line 161), I agree.

romainruaud: yep, I don't want to have unnecessary commented code

src/module-elasticsuite-catalog/view/adminhtml/web/template/form/element/product-sorter.html
rikwillems: I added this. The search button is still there but a reset link is added.

src/module-elasticsuite-catalog/view/adminhtml/web/js/form/element/product-sorter.js
rikwillems: It was a POC... :) I removed this.

src/module-elasticsuite-catalog/Model/ProductSorter/AbstractPreview.php
rikwillems: The receiving method should check its input and act on that. Now you have fluent code. Otherwise you'd end up with checks on both side which isn't necessary.

rbayet:
Hello @rikwillems,

Just tested with the UX changes on the search form, good job.

**But** it seems I had overlooked something : if I type a search term that cannot be found in the current category **and** if no product has been manually positioned, then the "Your product selection is empty for the Selected Store View [...]" message appears and as the search form is not visible anymore, the user has no choice but to reload the category page 

search_no_results
search_no_results_2

Looking at `product-sorter.html`, I wonder if the search field should not (conditionally ?) also be available under the `elasticsuite-admin-product-sorter-empty` (or located above the two "!hasProducts / hasProducts" divs.
And even with a custom info message in the `elasticsuite-admin-product-sorter-empty > message message-info` div.

@romainruaud  Do you concur ?

Regards,

rbayet:
Replaced by #1255

Issue #1198, opened 2018-11-24
M2.3.x ES2.6.x Category position indexing


Swahjak:
### Preconditions

Magento Version : 2.3.x

ElasticSuite Version : 2.6.x

Environment : developer

Third party modules : -

### Steps to reproduce

1. Index product fulltext

### Expected result

1. Category position is supplied

### Actual result

1. Category position is not supplied

Side note; We are in the process of migrating from 1.9.3.10 to 2.3.x.

I'm not sure if this is a bug or a migration issue but we're having problem with the category.position index. After migrating the `smile_virtualcategory_catalog_category_product_position` table is empty (is this a migration issue?). In `Smile\ElasticsuiteVirtualCategory\Model\ResourceModel\Product\Indexer\Fulltext\Datasource\CategoryData` are two selects; a 'BaseSelect' and a 'VirtualSelect'. To me it does not make sense that the 'BaseSelect' uses `p.position` which is mapped to `smile_virtualcategory_catalog_category_product_position` instead of to `catalog_category_product_index_storex`. Wouldn't it make more sense to set this to `cpi.position` or even to `COALESCE(p.position, cpi.position)`? In our case this select causes the `category.position` to be null and therefore sorting doesn't work.

romainruaud:
Unfortunately, it's by design that products positions are saved into `smile_virtualcategory_catalog_category_product_position` instead of being stored into the legacy table.

So actually, you'd better copy/paste product position data into this table during your migration process if you want to rely on them.

And this table is probably here to stay because we plan to allow to merchandise product position per store view, which is actually not doable with the magento legacy table.

Regards

Swahjak:
> And this table is probably here to stay because we plan to allow to merchandise product position per store view, which is actually not doable with the magento legacy table.

Actually Magento 2.3.x has 'new' tables allowing positions per store (something that is already taken into account in Elasticsuite). As an example we have the following tables:
- catalog_category_product_index
- catalog_category_product_index_replica
- catalog_category_product_index_store1
- catalog_category_product_index_store1_replica
- catalog_category_product_index_store2
- catalog_category_product_index_store2_replica

The store tables are an exact copy of what `catalog_category_product_index` used to be (in terms of columns). `catalog_category_product_index` is empty in 2.3.x. Elastic suite already takes this into account:
```php
/**
     * Get category product index table name.
     *
     * @param integer $storeId Store id.
     *
     * @return string
     */
[...]

PR #1199, opened 2018-11-24
Possible fix for Smile-SA/elasticsuite#1198

Swahjak:
Issue Smile-SA/elasticsuite#1198 for details


rbayet:
See explanations at https://github.com/Smile-SA/elasticsuite/issues/1198#issuecomment-442498038

Issue #1196, opened 2018-11-23
Support for Multi-Source Inventory (MSI) coming in 2.3.x


rbayet:
2.3.x introduces MSI (https://github.com/magento-engcom/msi/wiki) which allows to define different stock (stock_id) for different websites and inventory sources in (but probably not limited to) multi-websites environment.
When it is the case, it is not longer possible to rely purely on *legacy* tables cataloginventory_stock_* for indexing into ES the stock status of products, but there is a need to rely on new tables inventory_stock* (and probably inventory_sources*).

### Preconditions

Magento Version : CE 2.3.0-beta32 + sample data

ElasticSuite Version : 2.6.x

Environment : Developer

Third party modules : N/A

### Steps to reproduce

1. On a multi-websites environment (named hereafter "website A" and "website B"), create a new Inventory source (Stores > Inventory > Sources) named "alternate" besides the "default" one, and create a new Inventory stock (Stores > Inventory > Stocks) using that newly created source and with "website B" as the stock Sales Channel
2. On the catalog product grid, select from edition a **simple** product assigned to both "website A" and "website B" 
3. In the Sources fieldset, click the "Assign Sources" to assign the "alternate" source to the product (besides the "default" source already assigned) and save the product
4. Edit the product again (if need be) and in the Sources fieldset 
- switch the "Source Item Status" select to "Out of stock" for the "alternate" source
- switch if need be the "Source Item Status" select to "In Stock" (and assign some stock qty) for the "default" source
- save the product
5. Wait for cron reindexing or perform a full reindex via CLI
6. Inspect the indexed documents in both ES indexes for the product

Step 2. and 3. can also be done directly from the catalog product grid with the mass action "Assign Inventory Source".

### Expected result

1. In the "Website A/Store A" document, "stock.is_in_stock" should be set to "true"
2. In the "Website B/Store B" document, "stock.is_in_stock" should be set to "false"

### Actual result

1. In the "Website A/Store A" document, "stock.is_in_stock" is set to "true" (**OK**)
2. In the "Website B/Store B" document, "stock.is_in_stock" is set to "true" (**KO**)

romainruaud:
Let me know if it's doable directly on ElasticsuiteCatalog module.

Maybe something on the Stock Datasource that would be able to : 

- for a given website/store : be able to deduce how the stock is managed : legacy (it's unclear for me how relevant are legacy stock data & tables on Magento 2.3) or MSI
- then compute stock information for the proper table
[...]

PR #1202, opened 2018-11-27
Feature #1196 Multi Source Inventory support

rbayet:
Created an additional module with a somewhat evocative name, but I'm not so sure about it anymore, since \Magento\CatalogInventory\Api\StockRegistryInterface and \Magento\CatalogInventory\Api\StockConfigurationInterface are cleared marked as deprecated (see full list in https://devdocs.magento.com/guides/v2.3/inventory/catalog-inventory-replacements.html).

Having the changes directly in elasticsuite-catalog would remove the need for the extra interface I had to introduce.

### DIFF
--- a/src/module-elasticsuite-catalog/etc/di.xml
+++ b/src/module-elasticsuite-catalog/etc/di.xml
@@ -243,4 +243,6 @@
         </arguments>
     </type>
 
+    <preference for="Smile\ElasticsuiteCatalog\Model\ResourceModel\Product\Indexer\Fulltext\Datasource\InventoryDataInterface"
+                type="Smile\ElasticsuiteCatalog\Model\ResourceModel\Product\Indexer\Fulltext\Datasource\InventoryData" />
 </config>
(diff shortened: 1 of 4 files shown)


src/module-elasticsuite-catalog/etc/di.xml
```diff
@@ -231,4 +231,10 @@
                 sortOrder="100"
         />
     </type>
+
+    <type name="Smile\ElasticsuiteCatalog\Model\Product\Indexer\Fulltext\Datasource\InventoryData">
```
rbayet: Actually don't know why I went for argument over preference...

romainruaud:
Impressive ! seems it was pretty easy after all :)

I'm not sure either about the additional module : 

- implementation looks quite transparent to me, and could probably be done directly on the ElasticsuiteCatalog module directly.

- in any cases, hoping for BC with such huge changes, by having an additional module which is meant to be disabled on 2.2.x is kinda risky (could double the maintenance time), in addition of being quite ugly and not easily understandable.

- the Interface for Stock data fetching suits me well, we can keep it, it will allow people to write their own and switch the implementation nicely, so we can keep it.

Let's talk about it together today, but we'll probably go for a direct implementation on ElasticsuiteCatalog, and having a 2.7.0 major version which requires Magento >= 2.3.0.

PR #1207, opened 2018-11-29
Fixes #1201 Allow elasticsearch/elasticsearch 6.x

### LINKED ISSUE #1201: Support for composer elasticsearch 6.x
In composer.json file version ~5.1 of elasticsearch/elasticsearch is required. Which means supported versions are 5.1.0 - 5.1.9. So package cannot be installed using composer for the project that use ES 6.x.

Magento Version :
2.2.6

ElasticSuite Version :
2.6.3

Environment :
Developer

### Steps to reproduce

1. Open CLI
2. Do `composer require smile/elasticsuite ^2.6.0`

### Expected result

1. Package is installed

### Actual result

1. http://prntscr.com/lnhu0c

rbayet:
which is required for an Elasticsearch server in 6.x

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
         "magento/module-catalog": ">=102.0.0",
         "magento/module-catalog-search": ">=100.2.0",
         "magento/magento-composer-installer": "*",
-        "elasticsearch/elasticsearch": "~5.1"
+        "elasticsearch/elasticsearch": "~5.1|~6.0"
     },
     "replace": {
         "smile/module-elasticsuite-core": "self.version",


rbayet:
Let's discuss about this : which release line, before or after 2.7, and if any change to docs is actually required.

romainruaud:
Does it work with Magento EE ?

I guess it does, because of the piped syntax.

rbayet:
Yes it does work with Magento EE due to the piped syntax.

As discussed, as long as the the 5.x version of elasticsearch/elasticsearch is sufficient to work with a 6.x ElasticSearch server, let's not embark in having the risk of all CE users have or move to the 6.x version of elasticsearch/elasticsearch while the EE users being stuck with the 5.x version.
Which works perfectly fine anyway.

romainruaud:
Exact, and since the piped syntax will always prefer 6.4 over 5.1, it could create bad situation were people using an Elasticsearch 5 *server* were given the Elasticsearch 6 *connector*.

So yes, as we discussed together, it's a wont fix for now.

PR #1210, opened 2018-11-30
Fix 1209 admin attribute options

rbayet:


rbayet:
To test on a clean 2.3.0

rbayet:
Tested on a clean CE 2.3.0 : 
* without the swatches modules and with or without elasticsuite, it's not possible to edit / move / delete attribute options (JS error due to a missing addPanel function)
* * that error has been backported "as is" in the overriden JS file for elasticsuite
* with the swatches modules and with or without elasticsuite, it's possible to edit / move / delete attribute options
* * and with elasticsuite the custom behavior of allowing layered navigation on text and boolean works

Issue #1223, opened 2018-12-08
[BUG] ES 2.7 M 2.3 InventoryData loads all products


Swahjak:
### Preconditions

Magento Version : 2.3

ElasticSuite Version : 2.7

Environment : develop / production

Third party modules :

### Steps to reproduce

1. Try reindexing after upgrading to ES 2.7
2.
3.

### Expected result

1. Products are loaded into the index

### Actual result

1. Products only return a stock status and only the final batch is actually complete. The entire index before the last batch is overwritten by a stock only document.

Commit https://github.com/Smile-SA/elasticsuite/commit/010f0a86a15990270711980a2beeeab032328261#diff-59575d408d4bffc1107134de0085acdaL86 addresses multi inventory support but removes the filter on product ids.

The variable $productIds is never used. This causes the index to overwrite itself because it loads ALL product.

`Smile\ElasticsuiteCatalog\Model\ResourceModel\Product\Indexer\Fulltext\Datasource\InventoryData`
```php
    public function loadInventoryData($storeId, $productIds)
    {
        $websiteId = $this->getWebsiteId($storeId);
        $stockId   = $this->getStockId($websiteId);
        $tableName = $this->stockIndexTableProvider->execute($stockId);

        $select = $this->getConnection()->select()
            ->from(['product' => $this->getTable('catalog_product_entity')], [])
            ->join(
                ['stock_index' => $tableName],
                'product.sku = stock_index.' . IndexStructure::SKU,
                [
                    'product_id'    => 'product.entity_id',
                    'stock_status'  => 'stock_index.' . IndexStructure::IS_SALABLE,
                    'qty'           => 'stock_index.' . IndexStructure::QUANTITY,
                ]
            );

        return $this->getConnection()->fetchAll($select);
    }
```

Swahjak:
Created two slightly different pull requests since there was some room for improvement that wasn't directly related to fixing the bug. Up to the maintainers to merge the one they want ;)

rbayet:
Hello @Swahjak,

I'll look into it. I don't know why the issue didn't popped up.
What is your catalog size ?

I'll close your second PR, it actually breaks MSI support.

Regards,

Swahjak:
Hi @rbayet our catalog size is roughly 8000 sku's. But anything > 1000 products (batch size) should have this problem.

[...]

PR #1224, opened 2018-12-08
Fixes Smile-SA/elasticsuite#1223

Swahjak:
Fixes Smile-SA/elasticsuite#1223

On a side note; is there a specific reason for joining on SKU? The index table (`inventory_stock_x`) does contain a `product_id` column, no need to join `catalog_product_entity`. Create a secondary pull request with the adjusted select statement https://github.com/Smile-SA/elasticsuite/pull/1225.

### DIFF
--- a/src/module-elasticsuite-catalog/Model/ResourceModel/Product/Indexer/Fulltext/Datasource/InventoryData.php
+++ b/src/module-elasticsuite-catalog/Model/ResourceModel/Product/Indexer/Fulltext/Datasource/InventoryData.php
@@ -93,7 +93,9 @@ class InventoryData extends Indexer implements InventoryDataInterface
                     'stock_status'  => 'stock_index.' . IndexStructure::IS_SALABLE,
                     'qty'           => 'stock_index.' . IndexStructure::QUANTITY,
                 ]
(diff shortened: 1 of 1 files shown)


src/module-elasticsuite-catalog/Model/ResourceModel/Product/Indexer/Fulltext/Datasource/InventoryData.php
```diff
@@ -93,7 +93,8 @@ public function loadInventoryData($storeId, $productIds)
                     'stock_status'  => 'stock_index.' . IndexStructure::IS_SALABLE,
                     'qty'           => 'stock_index.' . IndexStructure::QUANTITY,
                 ]
-            );
+            )
+            ->where('product.entity_id IN(?)', $productIds);
```
rbayet: Could you 

- add a space between "IN" and "(?)" (just to nitpick)
- add a `->group('product.entity_id');` clause after the where ?

On a EE version, there could be staged updates present in the `catalog_product_entity` and otherwise, several (identical, mind you) rows could be retrieved for a single product.

PR #1225, opened 2018-12-08
Fixes Smile-SA/elasticsuite#1223 and improves select statement

Swahjak:
Fixes Smile-SA/elasticsuite#1223 and improves select statement

### DIFF
--- a/src/module-elasticsuite-catalog/Model/ResourceModel/Product/Indexer/Fulltext/Datasource/InventoryData.php
+++ b/src/module-elasticsuite-catalog/Model/ResourceModel/Product/Indexer/Fulltext/Datasource/InventoryData.php
@@ -84,16 +84,12 @@ class InventoryData extends Indexer implements InventoryDataInterface
         $tableName = $this->stockIndexTableProvider->execute($stockId);
 
         $select = $this->getConnection()->select()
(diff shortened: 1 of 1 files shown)


src/module-elasticsuite-catalog/Model/ResourceModel/Product/Indexer/Fulltext/Datasource/InventoryData.php
```diff
@@ -84,16 +84,12 @@ public function loadInventoryData($storeId, $productIds)
...
-            ->join(
-                ['stock_index' => $tableName],
-                'product.sku = stock_index.' . IndexStructure::SKU,
-                [
-                    'product_id'    => 'product.entity_id',
-                    'stock_status'  => 'stock_index.' . IndexStructure::IS_SALABLE,
-                    'qty'           => 'stock_index.' . IndexStructure::QUANTITY,
-                ]
-            );
+            ->from($tableName, [
```
rbayet: It will only work if you're only using the default stock : 

- the table `inventory_stock_1` is actually a view based on a join of `cataloginventory_stock` and `catalog_product_entity` so it does provide the product_id
- any subsequent stock index table will only have columns `sku`, `quantity` and `is_salable`

Performance loss due to the extra join is negligible if not non-existent if only using the default stock.

Issue #1205, opened 2018-11-29
Category Pages serverd from ElasticSearch indexes


badev-ajay:
Hi ElasticSearch Team,

Can we be served fully category page from ElasticSearch currently product list is coming from the database?

Swahjak:
I think that is the question; can we retrieve the entire product list from Elasticsearch instead of from the database. Skipping the database could have some interesting benefits.

romainruaud:
I agree, it could probably be a nice performance gain on frontend.

Probably quite easy to do since we have the source of any document in the search response actually.

It would require replacing/skipping the `_loadAttributes` of the collection to use data coming from Elasticsearch (and fallback to the legacy SQL query based on many many JOINs if an attribute is not indexed/retrieven).

BUT : 

- data coming from ES can be outdated (probably not that much, but it's not real time indexing) 
- there is probably something special to deal with stock status and price

Anyway, that's something we already thought about. I'd be happy to review any PR on this one.

badev-ajay:
Hi Team,

When We can aspect this feature or please give some clue about that sho i can change myself

bjoern-flagbit:
If you want to have the full catalog from ES you should maybe check this open-source stuff here:
https://github.com/lizards-and-pumpkins/catalog

Swahjak:
@rbayet @romainruaud As insiders, could you make an estimation of the work it would take to make this happen? I would consider spending some time on this, but without an in depth knowledge of ES it's hard to make an estimation on the time this would cost. If size isn't a problem we could replace / modify the flat tables to use elasticsearch as a backend.

Swahjak:
@bjoern-flagbit That's an interesting options. But to me it seems they aim to have a 'broad' support for catalogs while I think a Magento 2 specific solution would allow for full Magento 2 capability.

badev-ajay:
Hi Swahjak,
i have enabled product listing from backend now I am getting bellow error

1 exception(s):
Exception #0 (Exception): Notice: Array to string conversion in /var/www/html/m23/vendor/magento/module-catalog/Model/View/Asset/Image.php on line 188

Exception #0 (Exception): Notice: Array to string conversion in /var/www/html/m23/vendor/magento/module-catalog/Model/View/Asset/Image.php on line 188
[...]

PR #1228, opened 2018-12-10
Proof of concept using elastic suite for product listing

Swahjak:
References Smile-SA/elasticsuite#1205

As stated in #1205 it could be useful to use Elasticsuite (elasticsearch) as a backend for the product listing. For me there are two points of interest;
- Possible performance gain; Limiting the number of queries to the database or bypassing the database as a whole
- This could to some extend replace the product flat index; When there are too many attributes used for product listing the flat tables can no longer be used.

This is not a pull request that should be reviewed as a feature request, but rather as a proof of concept. I'm trying to figure out the direction that is preferred if this would ever be considered as a serious feature. I tried to use a minimal amount of code to get to a point where the full collection is provided by Elastic Suite.

Some initial insights / questions;
- The index source should have a list of attributes that are used for product listing. The proposal is based on the `attributes_indexed` which is far from optimal
- Is there a specific reason for having the attributes values as an array in the source document?
- Should this be part of `ElasticsuiteCatalog` or should it be developed in a separate folder?

Please share your thoughts how, and if, this could become a serious feature.

### DIFF
--- a/src/module-elasticsuite-catalog/Data/Collection/Db/FetchStrategy/SearchQuery.php
+++ b/src/module-elasticsuite-catalog/Data/Collection/Db/FetchStrategy/SearchQuery.php
@@ -1,0 +1,124 @@
+<?php
+/**
+ * DISCLAIMER
+ * Do not edit or add to this file if you wish to upgrade Smile ElasticSuite to newer
+ * versions in the future.
+ *
+ * @category  Smile
+ * @package   Smile\ElasticsuiteCore
+ * @author    Romain Ruaud <romain.ruaud@smile.fr>
+ * @copyright 2018 Smile
+ * @license   Open Software License ("OSL") v. 3.0
+ */
+namespace Smile\ElasticsuiteCatalog\Data\Collection\Db\FetchStrategy;
+
+use Magento\Framework\Data\Collection\Db\FetchStrategyInterface;
+use Magento\Framework\DB\Select;
+use Magento\Search\Model\SearchEngine;
+use Smile\ElasticsuiteCatalog\Helper\ProductListing;
+use Magento\Framework\Data\Collection\Db\FetchStrategy\Query;
+use Smile\ElasticsuiteCore\Helper\Mapping;
+
+/**
+ * Class SearchQuery
+ */
+class SearchQuery extends Query implements FetchStrategyInterface
+{
+    /**
+     * Search Request
+     *
+     * @var \Smile\ElasticsuiteCore\Search\RequestInterface
+     */
+    private $request;
+
+    /**
+     *
+     *
+     * @var \Magento\Framework\Search\ResponseInterface
+     */
+    private $response;
+
+    /**
+     * Search Engine
+     *
+     * @var \Magento\Search\Model\SearchEngine
+     */
+    private $searchEngine;
+
+    /**
+     * Product Listing helper used to check if product listing is enabled
+     *
+     * @var \Smile\ElasticsuiteCatalog\Helper\ProductListing
+     */
+    private $productListingHelper;
+
+    /**
+     * SearchQuery constructor.
+     *
+     * @param \Magento\Search\Model\SearchEngine               $searchEngine         Search Engine
+     * @param \Smile\ElasticsuiteCatalog\Helper\ProductListing $productListingHelper Product Listing Helper
+     */
+    public function __construct(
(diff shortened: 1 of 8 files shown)


romainruaud:
Hello @Swahjak 

Wow, thank you for taking some time to write it. It's not that far from what I would have done myself.

So again, thank you.

Do you have any metrics with/without this one ? 

I see some things that will probably not work very well, or which remains to be discussed : 

- for composite products, attributes are merged inside ES, even those who "should not" support multiple values : Eg. `name` is an array for a bundle product. This way, I'm not sure about how this could work. Maybe we should have another "version" of fields containing the data to be used in product listing ?

- actually I'm not sure about proper price injection inside product data, maybe this should remain as a join on the price index table. We need to think about it imho.

- same thing for stock data, should we fetch it from ES or keep the join, I'm unsure.

- people may encounter strange issue when using the fulltext collection and trying to retrieve data for attributes that are not used in product listing. This can be confusing, and since data will not exist on the ES index in this case, there is no proper solution (or we could have an hybrid fetch from ES and fallback to the DB... but I'm not a huge fan of this solution).

- and last but not least : is it still relevant to do such things when the new standard is PWA/GraphQL and we will soon have headless magento directly triggering the Search API ?

So there are still many points to discuss on this one.

Regards, thank you again for sharing your point of view !

Swahjak:
@romainruaud 

You are welcome, thank you for your efforts in making this request possible :wink: 

> Do you have any metrics with/without this one ?
Not yet
> * for composite products, attributes are merged inside ES, even those who "should not" support multiple values : Eg. `name` is an array for a bundle product. This way, I'm not sure about how this could work. Maybe we should have another "version" of fields containing the data to be used in product listing ?

I see, I haven't gotten to the point with anything other than simple products.

> * actually I'm not sure about proper price injection inside product data, maybe this should remain as a join on the price index table. We need to think about it imho.
> * same thing for stock data, should we fetch it from ES or keep the join, I'm unsure.

I'm not sure. For starters I think we would be better of leaving it out. I think observers / plugins would go a long way, but maybe not far enough. We could hook to both the Magento index processes which should give us near realtime stock and pricing data.

> * people may encounter strange issue when using the fulltext collection and trying to retrieve data for attributes that are not used in product listing. This can be confusing, and since data will not exist on the ES index in this case, there is no proper solution (or we could have an hybrid fetch from ES and fallback to the DB... but I'm not a huge fan of this solution).

I agree that it might be hard to sell for a broader audience, maybe this could / should be more of a side project outside of this repo.

> * and last but not least : is it still relevant to do such things when the new standard is PWA/GraphQL and we will soon have headless magento directly triggering the Search API ?

API's will still heavily rely on the database and they might actually be better suited for use with ES since they are more predictable. Looking at https://github.com/DivanteLtd/vue-storefront for instance, which heavily relies on Elasticsearch. I'm not sure about GrapQL though. I'm quite sure about the fact that Magento's current API is hardly suited for any serious PWA (though I hope GraphQL will change that).The other question is how long it will take for the general public to start using PWA as their general storefront. This could easily take a (couple of) year(s).

romainruaud:
During christmas holidays I thought about this one.

Here is another approach that I do prefer : 

- create a new DataSource 
- these data source will compute indexing for all attributes that are marked as "used_in_product_listing" automatically
- it will wrap this data inside a meta field on the ES index. Something like `_listing`, allowing to have `_listing.name` and so on... this way, no collision with legacy search index data.
- `_listing` sub fields should have a lightweight mapping, something like `index : false` or `not_analyzed` or both (should check). Should allow proper indexing, and since they are used only for data loading, no need to mess them with the analyzed fields.
- and finally, same logic for the `_loadAttributes` part.

Let me know your opinion @Swahjak

romainruaud:
I close for now since it will not be merged as is. But I keep in mind there are a lot of good ideas in this one.

francescosalvi:
Hi @romainruaud 
I know the issue is closed, but stumbled upon it while looking for information about using the ES index to feed _all_ catalog data into a headless solution's frontend, where I don't require all catalog attributes to be searchable/filterable, or indexed at all.

> something like index : false

from my limited knowledge of ElasticSearch, this is exactly the approach I was able to gather I should follow... can you confirm that at the moment ElasticSuite does not offer any way (e.g. via configuration) to allow for such field mappings to be generated?

Thank you

Issue #1234, opened 2018-12-13
Search Suggestion doesnt work


wouter-toppershops:
### Preconditions

Magento Version : 2.3

ElasticSuite Version : 2.7.1

Environment : DEVELOP

Third party modules :

### Steps to reproduce

1. search in bar, and wait for search suggestion
2.
3.

### Expected result

1. error on the suggestion controller (https://pastebin.com/DTBfZwj9), our custom attribute does not have a `type` => `termBucket` but its `0` => `termBucket` caused by `Smile\ElasticsuiteCatalog\Model\Autocomplete\Product\Collection\AttributeAggregation:65` there is no `type` key set.

### Actual result

1. 
![screenshot from 2018-12-13 13-16-07](https://user-images.githubusercontent.com/40226924/49939436-22ceb600-fedd-11e8-8fc2-ce0b628449ac.png)

wouter-toppershops:
See https://github.com/Smile-SA/elasticsuite/pull/1235

rbayet:
Confirmed on M2 2.3.x and 2.7.x / developer mode

wouter-toppershops:
See New Pull: https://github.com/Smile-SA/elasticsuite/pull/1237

rbayet:
Hello @wouter-toppershops,

I've seen it, thanks !

Regards

PR #1237, opened 2018-12-14
Update AttributeAggregation.php

wouter-toppershops:
For `type` instead of `0` in the array


wouter-toppershops:
See Issue: https://github.com/Smile-SA/elasticsuite/issues/1234

PR #1255, opened 2019-01-09
Feature/search category products preview

rbayet:
Finalization of PR #1183 which was a proof of concept for #157 

Rebased on 2.6.x and squashed @rikwillems commits and treated https://github.com/Smile-SA/elasticsuite/pull/1183#issuecomment-442540605 
- search form is not visible in the search term merchandizer screen
- on the category products preview, the search form is only visible if there are products or there is an outstanding search
- added i18n for de_DE, fr_FR and en_US


romainruaud:
OK for me, I let you squash the commits and merge this one

PR #1268, opened 2019-01-17
WIP : Admin category loading optimization

### LINKED ISSUE #1244: M2.3.x ES2.7.1 | Performance admin category edit
We experiencing some 'extreme' performance issues on the category edit page in Magento 2.3.0 (but I assume this happens in other versions as well).

### Preconditions

* ~10k products
* ~3k categories
* 1 root category

Magento Version : 2.3.0

ElasticSuite Version : 2.7.1

Environment : Developer and production

Third party modules :

### Steps to reproduce

1. See preconditions

### Expected result

1. Category edit page loads within several seconds

### Actual result

1. It takes up to 2 minutes (or more) to load the category page
2. Loading the page costs a staggering ~3600 sql queries

A simple comment in `Smile\ElasticsuiteVirtualCategory\Model\Rule::154` reduces the number of queries from 3670 to 1269 and the load time from 83 to 23 seconds (on a development box, all cache enabled).

`Smile\ElasticsuiteVirtualCategory\Model\Rule::154` causes a recursive call for subcategories AND fires at least one count query. I'm not sure on a solution just yet. The main problem seems to be the amount of categories (~3000) in a single tree, since the loading gets easier the further you go down the tree.

Still doing some further investigation. Could some body elaborate on what the purpose of this function is?

```php
    public function getCategorySearchQuery($category, $excludedCategories = [])
    {
        $query         = null;

        if (!is_object($category)) {
            $category = $this->categoryFactory->create()->setStoreId($this->getStoreId())->load($category);
        }

        if (!in_array($category->getId(), $excludedCategories)) {

            if ((bool) $category->getIsVirtualCategory() && $category->getIsActive()) {
                $query = $this->getVirtualCategoryQuery($category, $excludedCategories);
            } elseif ($category->getId() && $category->getIsActive()) {
            }
                // $query = $this->addChildrenQueries($query, $category, $excludedCategories);
            }
        }

        return $query;
    }
```

romainruaud:
Will aggregate commits related to Fix #1244

### DIFF
--- a/src/module-elasticsuite-catalog/Plugin/Ui/Category/Form/DataProviderPlugin.php
+++ b/src/module-elasticsuite-catalog/Plugin/Ui/Category/Form/DataProviderPlugin.php
@@ -109,7 +109,9 @@ class DataProviderPlugin
 
         $currentCategory = $dataProvider->getCurrentCategory();
 
-        $data[$currentCategory->getId()]['facet_config'] = $this->getFilterableAttributeList($currentCategory);
+        if ($currentCategory->getId() !== null && $currentCategory->getLevel() > 2) {
+            $data[$currentCategory->getId()]['facet_config'] = $this->getFilterableAttributeList($currentCategory);
+        }
 
         return $data;
     }


romainruaud:
We won't go further on this one for now.

I merge

PR #1272, opened 2019-01-17
Ensure affected products are properly computed if resetting default v…

romainruaud:
…alue when editing a specific store view.


rbayet:
Cross-tested on EE 2.2.5, CE 2.2.6, EE 2.2.7

src/module-elasticsuite-virtual-category/Plugin/Catalog/Category/SaveProductsPositions.php
```diff
@@ -101,16 +101,36 @@ function () use ($category) {
      */
     private function getAffectedProductIds($category)
```
rbayet: The fix on getAffectedProductIds() will only apply for "index on save" behavior, though

PR #1273, opened 2019-01-18
head.additional declaration

mageho:
Good afternoon,

Just one typo error with the script declaration "_Smile_ElasticsuiteTracker::js/tracking.js_" (anyway, it include correctly this javascript file).

And some inconsistents with the declaration of head.additional which is a block not a container 
(Same problem on Magento_GoogleAnalytics, this is not solved in my version of _Magento 2.3_ - https://github.com/magento/magento2/issues/16497)

Personally, I would rather move the script tracking.js into _before.body.end_ before _smile.tracker.page.base_ block.
I think you recommends to place it in the <head>, so if someone aborts a load of the page, the script might get executed.
"It helps you to accurately record traffic stats because your header scripts are loaded earlier than any other scripts on your page."

Cheers,
Ilan PARMENTIER


romainruaud:
You are right, this was quite old code and Magento layout has changed.

This one will be part of next major version.

Regards

PR #1276, opened 2019-01-21
Force scheduled reindexing of affected products

rbayet:
after a reset of positions and blacklist

### DIFF
--- a/src/module-elasticsuite-virtual-category/Model/ResourceModel/Product/Indexer/Fulltext/Changelog/Writer.php
+++ b/src/module-elasticsuite-virtual-category/Model/ResourceModel/Product/Indexer/Fulltext/Changelog/Writer.php
@@ -1,0 +1,83 @@
+<?php
+/**
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade this module to newer
+ * versions in the future.
+ *
+ *
+ * @category  Smile
+ * @package   Smile\ElasticsuiteVirtualCategory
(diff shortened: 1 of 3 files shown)


romainruaud:
Looks clean to me (code point of view), I let you merge and cascade it if it's working.

rbayet:
> Looks clean to me (code point of view), I let you merge and cascade it if it's working.

Unfortunately, it covers the case where resetting everything after switching "Use default positions" from "Yes" to "No", but not the case immediately after that : falling back to "Use default positions" from "No" to "Yes" with positions and blacklist defined at the default level.
It seems the `dataHasChangedFor` does not work in this case (because of the default value of the `use_store_positions` at store_id 0 ?).

I'm trying to find a solution other than brute force which would consist in to unconditionnally write the affected product ids to the CL table if the indexer is in scheduled config.

rbayet:
> I'm trying to find a solution other than brute force which would consist in unconditionnally write the affected product ids to the CL table if the indexer is in scheduled config.

Went half-way, condition went from 
``` 
        if (!empty($category->getAffectedProductIds()) 
            && $category->dataHasChangedFor('use_store_positions')
        )
```
to
```
        if (!empty($category->getAffectedProductIds())
            && ($category->dataHasChangedFor('use_store_positions') || (false == $category->getUseStorePositions()))
        )
```

romainruaud:
Ok for me, as long it remains on a single responsibility plugin, I'm fine with it.

You can merge and cascade this one

PR #1290, opened 2019-01-30
Fix product image plugin - magento changed the visibility of getFilte…

delyriand:
…rArray method

### DIFF
--- a/src/module-elasticsuite-swatches/Model/Plugin/ProductImage.php
+++ b/src/module-elasticsuite-swatches/Model/Plugin/ProductImage.php
@@ -28,17 +28,46 @@ class ProductImage extends \Magento\Swatches\Model\Plugin\ProductImage
 {
     /**
      * {@inheritdoc}
-     * @deprecated since Magento 2.1.6
      */
-    protected function getFilterArray(array $request)
-    {
-        $filterArray = parent::getFilterArray($request);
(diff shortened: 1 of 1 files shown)


rbayet:
Hello @delyriand,

Our CI ensures compatibility of the 2.6.x branch with both 2.2.5 and 2.2.6 versions of Magento.

The problem with that is that the visibility of `getFilteredArray` is **not yet** changed in 2.2.5, so redeclaring it as private when it's protected there causes a PHP fatal error.
Could you alter your PR so the *local* `getFilteredArray` is named differently (even if it's a blatant copy) to avoid this issue ?

Regards,

delyriand:
Hello @rbayet,

it's done ;)

romainruaud:
so this one is correct now ? @rbayet can you check it against 2.2.5 and 2.2.6 ?

rbayet:
2.2.5 (EE production mode) in catalog navigation + search : OK
2.2.6 (developer mode) catalog navigation OK, search KO

delyriand:
@rbayet,

I confirm i have same problem with magento 2.2.7 in developer mode. It seems that issue is at the level of [Magento](https://github.com/magento/magento2/issues/17825).
Indeed, in search, we do not get block interceptor (ListProduct\Interceptor) so plugins are not called

PR #1296, opened 2019-02-06
Aggregations: support for pipeline aggs and filter

rbayet:
Both can be defined in a container.
Only bucket selector pipeline aggregation for now.


rbayet:
TODO : Unit testing.

PR #1305, opened 2019-02-15
allow zero values options indexation

sky-hub:
This Pull Request fixes indexing issue for zero value attribute options, including boolean attributes.

Changes:
1. src/module-elasticsuite-catalog/Helper/AbstractAttribute.php line 175
- use `array_map` to filter attribute values as `array_filter` removed zero values.

Fixed issue:
- allow indexing for "No" option for boolean attributes
- allow indexing for "0" option for numeric attributes

1. src/module-elasticsuite-catalog/Helper/AbstractAttribute.php line 274
- removed cast to bool for boolean type attributes to avoid `false` value being removed by `array_map`

1. src/module-elasticsuite-catalog/Model/Layer/Filter/Boolean.php line 91
- removed value condition to display correctly the label for both "Yes" and "No" options.


src/module-elasticsuite-catalog/Helper/AbstractAttribute.php
```diff
@@ -272,7 +272,7 @@ public function getFilterField(AttributeInterface $attribute)
     private function prepareSimpleIndexAttributeValue(AttributeInterface $attribute, $value)
     {
         if ($this->getFieldType($attribute) == FieldInterface::FIELD_TYPE_BOOLEAN) {
-            $value = boolval($value);
+            $value = intval($value);
```
romainruaud: Not sure Elasticsearch will not complain if we try indexing 0 or 1 into a boolean field.

src/module-elasticsuite-catalog/Helper/AbstractAttribute.php
```diff
@@ -172,7 +172,7 @@ public function prepareIndexValue(AttributeInterface $attribute, $storeId, $valu
         }
 
         $value = array_map($this->attributeMappers[$mapperKey], $value);
-        $value = array_filter($value);
+        $value = array_diff(array_map('trim', $value), ['', null, false]);
```
romainruaud: A bit ugly, I understand why you removed the `boolval` later, mostly because `trim(true)` would result in `"1"` and `trim(false)` would result in `''`.

romainruaud:
I'm actually not sure how to handle this one.

I see two differents things : 

- allow "zero values" for attributes : definitely worth it. Your use case with 0/1/2 based attribute speaks by itself.

- boolean "no" indexing : we could achieve it and index true/false and yes/no labels. BUT this would result in a loss of data.

Let me explain the second point : you might have an attribute, let's say "Performance Fabric", which can have yes/no values.

With our current implementation, "performance fabric" (the raw string) is indexed into search data for all product having it set to yes. Which is imho a good thing, allowing customers to find products by searching for "performance fabric".

sky-hub:
Never thought about the "performance fabric" example, but it does make sense. Not sure either how to handle it. Do you think it is possible to add a config option which allows the customer to choose if the values should be indexed as "Yes/No" or as "performance fabric"?

rbayet:
Hello @sky-hub and @romainruaud 

I can confirm that using intval does not cut it, Elasticsearch will reject the document.
I decided to treat the problem of false, 0 and 0.00 indexing overall.
I decided to keep the behaviour of indexing the name of the label when the boolean attribute value is true (instead of Yes) and not indexing anything (for the label) for false.

rbayet:
Hello @sky-hub and @romainruaud 

I can confirm that using intval does not cut it, Elasticsearch will reject the document.
I decided to treat the overall problem of false, 0 and 0.00 indexing overall.
I decided to keep the behaviour of indexing the name of the label when the boolean attribute value is true (instead of Yes) and not indexing anything (for the label) for false.

See https://github.com/Smile-SA/elasticsuite/pull/2482

Regards,

PR #1313, opened 2019-02-19
Allow resetting the sort order to an empty array

abbatis:
In a normal Magento collection you would do something like this:

`$collection->getSelect()->reset(Zend_Db_Select::ORDER);`

However, this obviously wouldn't work for elasticsuite.


rbayet:
Hello @abbatis,

Could you recreate/relocate your feature with 2.6.x-dev as the base branch ?
We will be able to integrate it in both 2.6.x and 2.7.x releases.

Regards,

abbatis:
Will create a new pull request.

PR #1318, opened 2019-02-20
Lowercase given sort direction for missing

ulftietze:
Lowercase that given direction to make sure that neither magento (which it does sometimes) nor a dev can input uppercase and you get some wrong results.

### DIFF
--- a/src/module-elasticsuite-core/Search/Request/SortOrder/Standard.php
+++ b/src/module-elasticsuite-core/Search/Request/SortOrder/Standard.php
@@ -60,7 +60,7 @@ class Standard implements SortOrderInterface
         $this->name      = $name;
         $this->missing   = $missing;
         if ($this->missing === null) {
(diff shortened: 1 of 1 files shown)


ulftietze:
@romainruaud May have a look here? :)

rbayet:
Hello @ulftietze,

Could you relocate/recreate your patch with 2.7-x as the base branch ?
It will be easier to integrate it in the next release.

Regards,

romainruaud:
PR is OK for me, except that it would be better rebased on 2.7.x branch as said by @rbayet

rbayet:
@romainruaud side issue : https://github.com/Smile-SA/elasticsuite/pull/1333#issuecomment-467033666

romainruaud:
Ok after some investigation, it's nearly useless since Elasticsearch does not complain anymore about sort orders in uppercase since 5.x.

Versions < 2.6.x of Elasticsuite are not part of our Open Source support anymore. So merging this on 2.6.x (which require ES >5) will have no effect.

I close this one, but thank you for proposing it.

PR #1323, opened 2019-02-21
Fixes #1321 Prevent js errors if tracker blocked (2.7.x)

### LINKED ISSUE #1321: JS errors and non working search with enabled ublock origin
Hi,

i use chrome with enabled ublock origin as adblocker with this configuration:
ublock

In this case is get js errors, because the js file `...Smile_ElasticsuiteTracker/js/tracking.js` is blocked:
jserrors

The bigger problem is, also the search function is not working in case of the js errors.

Fix:
Add a try catch to all `smileTracker` calls.

rbayet:
by browser plugin (ex: ublock origin)


romainruaud:
Fix #1321

Issue #1322, opened 2019-02-21
Prices are not being indexed and sorting by price is not working


Mathiu:
Hi,

I have a problem where sorting products inside category by price is not working at all. I have dig a little and can see that prices are not being indexed in elasticsearch which I think is related because as far as I digged into the code sorting is done inside the query to elasticsearch. 

Here is what I get when I request: http://localhost:9200/magento2_default_migrated_catalog_product_20190221_070838/product/_search?pretty&q=entity_id:101722

```
{
  "took" : 8,
  "timed_out" : false,
  "_shards" : {
    "total" : 1,
    "successful" : 1,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : 1,
    "max_score" : 1.0,
    "hits" : [
      {
        "_index" : "magento2_default_migrated_catalog_product_20190221_070838",
        "_type" : "product",
        "_id" : "101722",
        "_score" : 1.0,
        "_source" : {
          "entity_id" : "101722",
          "attribute_set_id" : "9",
          "type_id" : "configurable",
          "sku" : [
            "TestS30",
            "TestS30-one size"
          ],
          "has_options" : false,
          "required_options" : false,
          "created_at" : "2016-10-21 10:37:28",
          "updated_at" : "2018-11-20 06:39:06",
          "search_weight" : "0",
          "visibility" : "4",
          "category" : [
            {
              "category_id" : 3,
              "is_virtual" : "false"
            },
            {
              "category_id" : 8,
              "is_parent" : true,
              "is_virtual" : "false"
            },
            {
              "category_id" : 12,
              "is_parent" : true,
              "is_virtual" : "false",
              "name" : "Brands"
            },
            {
              "category_id" : 256,
              "is_parent" : true,
              "is_virtual" : "false",
              "name" : "Test "
            }
          ],
          "name" : [
            "Testing123"
          ],
          "indexed_attributes" : [
            "name",
            "image",
            "description",
            "status",
            "tax_class_id",
            "news_from_date"
          ],
          "image" : [
            "/m/e/metting123.jpg"
          ],
          "description" : [
            "testing123"
          ],
          "status" : [
            1
          ],
          "option_text_status" : [
            "Enabled"
          ],
          "tax_class_id" : [
            2
          ],
          "option_text_tax_class_id" : [
            "Taxable Goods"
          ],
          "news_from_date" : [
            "2017-01-01 00:00:00"
          ],
          "children_ids" : [
            101723
[...]

PR #1326, opened 2019-02-22
Add Magento\Framework\App\ResourceConnection table name altering plugins to di.xml

Mathiu:
Fixes my particular issue (#1322) on a migrated store from M1 to M2 where price indexes were being retrieved from a wrong database table.

The issue here is that Magento is using 2 plugins for Magento\Framework\App\ResourceConnection class but only on frontend area to alter the name of the "catalog_product_index_price" which is not the case when reindexing data, the scope is then adminhtml.

I think this might also be a Magento core issue, so if that's the case, please let me know.

### DIFF
--- a/src/module-elasticsuite-catalog/etc/di.xml
+++ b/src/module-elasticsuite-catalog/etc/di.xml
@@ -223,4 +223,9 @@
     <type name="Smile\ElasticsuiteCore\Model\Search\RequestMapper">
         <plugin name="catalogProductRequestMapper" type="Smile\ElasticsuiteCatalog\Plugin\Search\RequestMapperPlugin" sortOrder="10"/>
     </type>
+
+    <type name="Magento\Framework\App\ResourceConnection">
+        <plugin name="get_catalog_category_product_index_table_name" type="Magento\Catalog\Model\Indexer\Category\Product\Plugin\TableResolver"/>
+        <plugin name="get_catalog_product_price_index_table_name" type="Magento\Catalog\Model\Indexer\Product\Price\Plugin\TableResolver"/>
+    </type>
 </config>


Mathiu:
As far as I can see the "get_catalog_product_price_index_table_name" plugin is a part of indexer optimization of magento 2.2.6: https://github.com/magento/magento2/commit/a3de9ab0b5d0288c837b8e24f849354d1d95f681#diff-da4fdd9a985964091b2bbe0a3722ab77 and https://github.com/magento/magento2/commit/412650ccdbc0a046555c50f4152e304a75507bb8#diff-da4fdd9a985964091b2bbe0a3722ab77 and its apparently being enabled for almost every area but adminhtml.

I would appreciate if anyone could look at this and advise if its a Magento core issue or how to fix it for 2.2.5 as well.

Thanks!

rbayet:
Hello @Mathiu,

Yes, to me, it might be a core issue.
Without an access to MAGETWO-91820, MAGETWO-91873 and MAGETWO-64467, I don't really understand why the plugins are declared everywhere *but* in adminhtml.
So as it stands, @romainruaud and I have an issue about your fix, because if it is **the** fix, it should be present in `magento/module-catalog`.

On the other hand, looking at the rest of the changes in `module-catalog/etc/di.xml`, the 2 commits you pointed might be a BC in magento we need to address in our module.
We will have a look and may get back to you in the issue #1322.

Regards,

Mathiu:
@rbayet Awesome, thank you for reviewing this!

rbayet:
Closing this PR as it would not fully address the issue pointed in #1322 : as it stands, ElasticSuite does not support the price index segmentation mode other than `none` (ie "no segmentation").
See https://github.com/Smile-SA/elasticsuite/issues/1322#issuecomment-467399936 

This PR would only work with a `website` segmentation and would only index valid prices if they do not vary between websites.

PR #1327, opened 2019-02-22
Add an alt attribute to the tracking image.

paulpartington-cti:
The tracking pixel is flagged by WCAG accessibility scanners because it doesn't have an alt tag. This change adds an empty alt attribute which means it's treated as a decorative image.


rbayet:
Hello @paulpartington-cti, 

Can you relocate/recreate your fix with a 2.6-x branch base ?
So we will be able to merge (and then cherry-pick) it for both M2 2.2.x and 2.3.x compatible releases.

Regards,

paulpartington-cti:
@rbayet No problem. I've created https://github.com/Smile-SA/elasticsuite/pull/1329

PR #1330, opened 2019-02-25
Add an alt attribute to the tracking image.

rbayet:
Add an alt tag to the tracking pixel so that it is treated as a decorative element in WCAG scanners.


rbayet:
Cherry-pick of PR #1329 merged on 2.6.x

PR #1328, opened 2019-02-22
Allow resetting the sort order to an empty array

abbatis:
In a normal Magento collection you would do something like this:

`$collection->getSelect()->reset(Zend_Db_Select::ORDER);`

However, this obviously wouldn't work for elasticsuite.

PR #1331, opened 2019-02-25
Feature allow resetting sort order 2.6.x

rbayet:
Revamp of PR #1328 to address the phpmd issue.
(See possible origin/fix : https://github.com/phpmd/phpmd/issues/409 and https://github.com/phpmd/phpmd/pull/495)


rbayet:
Correct warning suppresson was PHPMD.TooManyPublicMethods and it had to be located at the end of the class' PHPDoc block.

PR #1336, opened 2019-02-27
Fixes #1321 Prevent js errors if tracker blocked (2.6.x)

### LINKED ISSUE #1321: JS errors and non working search with enabled ublock origin

rbayet:
by browser plugin (ex: ublock origin)

### DIFF
--- a/src/module-elasticsuite-tracker/view/frontend/templates/config.phtml
+++ b/src/module-elasticsuite-tracker/view/frontend/templates/config.phtml
@@ -19,6 +19,7 @@
 <?php if ($block->isEnabled()) : ?>
 <script type="text/javascript">
 //<![CDATA[
+try {
     smileTracker.setConfig({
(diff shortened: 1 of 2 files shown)


southerncomputer:
You might want to rename tracking.js - i did to fracking.js to avoid the ublock in the first place!

rbayet:
Hi @southerncomputer,

Yes, I figured out it was the name of the JS file that was the source of the problem, but I'll let @romainruaud decide.

Regards,

romainruaud:
Yes, we'll probably have to rename the tracker file.

But actually the try/catch is a proper workaround, and it should be here in any cases (we cannot break all pages due to tracking, whatever causes it to crash).

romainruaud:
Fix #1321

Issue #1086, opened 2018-09-19
M2.2.6 : BadRequest400Exception on a fresh install (index_already_exists_exception)


kanduvisla:
When installing Magento 2.2.6 using the CLI (it's a fresh install), I get a `BadRequest400Exception` when a `installData` script is trying to create stores.

It gave no problems in 2.2.5 and elasticsuite 2.5.x

### Preconditions

Magento Version: 2.2.6
ElasticSuite Version: 2.6.1
Environment: Developer
Elasticsearch: 5.6.8 with icu and phonetic plugins

### Steps to reproduce

Create an installData script that creates stores. Example script:

```
<?php

namespace Vendor\Module\Setup;

use Magento\Framework\Event\ManagerInterface;
use Magento\Framework\Setup\InstallDataInterface;
use Magento\Framework\Setup\ModuleContextInterface;
use Magento\Framework\Setup\ModuleDataSetupInterface;
use Magento\Store\Api\Data\StoreInterfaceFactory;
use Magento\Store\Api\GroupRepositoryInterface;
use Magento\Store\Api\StoreRepositoryInterface;
use Magento\Store\Api\WebsiteRepositoryInterface;
use Magento\Store\Model\ResourceModel\Group;
use Magento\Store\Model\ResourceModel\Store;
use Magento\Store\Model\ResourceModel\Website;

/**
 * Class InstallData
 */
class InstallData implements InstallDataInterface
{
    /**
     * @var StoreRepositoryInterface
     */
    protected $storeRepository;

    /**
     * @var Store
     */
    protected $storeResource;

    /**
     * @var StoreInterfaceFactory
     */
    protected $storeFactory;

    /**
     * @var ManagerInterface
     */
    protected $eventManager;

    /**
     * @var WebsiteRepositoryInterface
     */
    protected $websiteRepository;

    /**
     * @var Website
     */
    protected $websiteResource;

    /**
     * @var GroupRepositoryInterface
     */
    protected $groupRepository;

    /**
     * @var Group
     */
    protected $groupResource;

    /**
     * InstallData constructor.
     * @param StoreRepositoryInterface $storeRepository
     * @param Store $storeResource
     * @param StoreInterfaceFactory $storeFactory
     * @param ManagerInterface $eventManager
     * @param WebsiteRepositoryInterface $websiteRepository
     * @param Website $websiteResource
     * @param GroupRepositoryInterface $groupRepository
     * @param Group $groupResource
     */
    public function __construct(
[...]

PR #1352, opened 2019-03-14
only create index if it does not exists

philippsander:
fixes https://github.com/Smile-SA/elasticsuite/issues/1086


philippsander:
does not work

cooper85:
@philippsander anyway, thank you, works for me
Also please add same for Smile\ElasticsuiteThesaurus\Model\Indexer
to get it work in that place also
Not sure why it is closed without some verdict a-la "heh, lol, man, you need always to get empty solr db, to work with this extension while debugging Magento project / etc"

PR #1360, opened 2019-03-22
Fix issue with missing values on swatches attributes filters

jburel-clever-age:
Hi!

I found an issue with a particular swatch attribute type only: there is no value in the left sidebar on categories but all my products have a value for this attribute! It's weird!

### Preconditions
Magento Version : EE 2.3
ElasticSuite Version : 2.7.5 (latest)
Environment : Developer and Production

### Steps to reproduce
1. Add a swatch attribute into filters. Please verify that your products have a value for this!
2. Go to a category page
3. Check your availables filters on the left

### Expected result
1. All my products related attribute values are available for filtering

### Actual result
1. This attribute is not filterable, there is no value

### Fix
I found the bug for this particular case, there is a comparaison between value (= ID) and label !

NOTE : After checking all branches, it appears this bug concerns all ElasticSuite versions! So you can merge in all branches :)

Thanks!

### DIFF
--- a/src/module-elasticsuite-swatches/Block/Navigation/Renderer/Swatches/RenderLayered.php
+++ b/src/module-elasticsuite-swatches/Block/Navigation/Renderer/Swatches/RenderLayered.php
@@ -32,7 +32,7 @@ class RenderLayered extends \Magento\Swatches\Block\LayeredNavigation\RenderLaye
     protected function getFilterOption(array $filterItems, Option $swatchOption)
     {
         $resultOption = false;
-        $filterItem = $this->getFilterItemById($filterItems, $swatchOption->getLabel());
+        $filterItem = $this->getFilterItemById($filterItems, $swatchOption->getValue());
         if ($filterItem && $this->isOptionVisible($filterItem)) {
             $resultOption = $this->getOptionViewData($filterItem, $swatchOption);
         }


romainruaud:
Thank you for reporting this one, we'll have a look.

But can you give more details about the configuration of your particular attribute ? Since the "color" attribute is working out-of-the-box, I'm curious to check what is different for yours.

Regards

jburel-clever-age:
Hello,

Color is a select attribute, so it works successfully but if you create a "Text Swatch" attribute, you'll be able to reproduce the issue!
Your attribute must be a swatch attribute type :)

If you check the called method `getFilterItemById`, the second parameter must be an id, but actually it's the swatch option label (admin scope option label) !

Thanks for your response !

romainruaud:
I'm sorry but I'm actually unable to reproduce this issue.

Attribute I created : 

Sélection_667

Result in Front-Office (I have no products with value C) : 

Sélection_668

Maybe I missed something ?

romainruaud:
And if I apply your fix, I have this : 

image

So actually, either we are missing some evidences, or I'm a bit tired, but for me the legacy behavior is fine, whereas your fix is breaking my dev environment.

jburel-clever-age:
Hi Romain,

I don't know what you can't reproduce this issue :(
Are you in 2.3.1 version?

Maybe it's due to a custom layout directive or something else. If you don't have this issue with a stock Magento version and ElasticSuite module, you can close this issue!

Sorry for the time loss :(

tiagofssampaio:
Hello,

I think I'm into the same problem (kinda)!
Basically I have an attribute (attribute code => "size") with with the option label "XS" and corresponds to the option_id "34". What happens is that I have an option with the label "34" - see the problem? That's conflicting the option_id with the option label.

I fixed the issue overriding this function in "\Smile\ElasticsuiteSwatches\Block\Navigation\Renderer\Swatches\RenderLayered":

`    protected function getFilterOption(array $filterItems, Option $swatchOption)
    {
        $resultOption = false;
        $filterItem = $this->getFilterItemByLabel($filterItems, $swatchOption->getLabel());
        if ($filterItem && $this->isOptionVisible($filterItem)) {
            $resultOption = $this->getOptionViewData($filterItem, $swatchOption);
        }
        return $resultOption;
    }`

TO

`    protected function getFilterOption(array $filterItems, Option $swatchOption)
    {
        $resultOption = false;
        $filterItem = $this->getFilterItemById($filterItems, $swatchOption->getValue());
        if ($filterItem && $this->isOptionVisible($filterItem)) {
            $resultOption = $this->getOptionViewData($filterItem, $swatchOption);
        }
        return $resultOption;
    }`

The version I'm using is 2.8.4

PR #1384, opened 2019-04-18
Fix virtual category api

### LINKED ISSUE #1298: How to set rules for virtual category via API in Magento 2?
Hello, thank you for a great suite, that you guys have developed.
I do have a question though:

How do I create a virtual category and set rules for it via API?

PS To see what it might look like, I manually created a virtual category and set rules for it, but when I retrieve info about this category via API, 'virtual_rule' value is just an empty list. Why is that? Where do I find en example for setting rule for a virtual category via API?

Thanks.

### Preconditions

Magento Version :
Magento CE 2.3

ElasticSuite Version :
2.7.4

Environment :
Developer mode

Third party modules :
None

romainruaud:
Fix #1298

### DIFF
--- a/src/module-elasticsuite-catalog-rule/Api/Data/ConditionInterface.php
+++ b/src/module-elasticsuite-catalog-rule/Api/Data/ConditionInterface.php
@@ -1,0 +1,121 @@
+<?php
+/**
+ * DISCLAIMER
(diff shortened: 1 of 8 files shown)


romainruaud:
Merged despite Travis build error, it's false-positive.

PR #1392, opened 2019-04-25
Fixes #1369 virtual_category_root new default value 2.3.x

### LINKED ISSUE #1369: Virtual Categories Root select : default/root category not pre-selected anymore
In M2 2.2.8 and 2.3.1, when creating a **new** virtual category, the "Virtual Category Root" select does not pre-select anymore the store default/root category.
In 2.2.8, the select displays/has pre-selected a "Select..." option.
In 2.3.1, the select displays/has pre-selected a "Entity with ID: 0 doesn't exist" option.

### Preconditions

Magento Version : 2.2.8 and 2.3.1 (CE and EE)

ElasticSuite Version : 2.6.8 / 2.6.x-dev and 2.7.6 / 2.7x-dev

Environment : Developer and Production mode

Third party modules : N/A

### Steps to reproduce

1. In admin, create a new sub-category under an existing category, save it
2. In "Products in Category", switch "Virtual Category" to "Yes"

### Expected result

1. The "Virtual Category Root" select which appears should have pre-selected the current store default/root category option
virtual_category_root_expected

### Actual result

1. In 2.2.8, a "select" option is pre-selected instead, the actual default/root category option being below
virtual_category_root_select

2. In 2.3.1, a "Entity with ID: 0 doesn't exist" option is pre-selected, the actual default/root category option being below
virtual_category_root_entity_does_not_exist

rbayet:
test new install setup: *OK*

### DIFF
--- a/src/module-elasticsuite-virtual-category/Setup/VirtualCategorySetup.php
+++ b/src/module-elasticsuite-virtual-category/Setup/VirtualCategorySetup.php
@@ -118,7 +118,7 @@ class VirtualCategorySetup
                 'input'      => null,
                 'global'     => \Magento\Eav\Model\Entity\Attribute\ScopedAttributeInterface::SCOPE_GLOBAL,
                 'required'   => false,
-                'default'    => 0,
+                'default'    => null,
                 'visible'    => true,
                 'note'       => "Root display of the virtual category (usefull to display a facet category on virtual).",
                 'sort_order' => 200,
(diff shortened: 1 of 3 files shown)


src/module-elasticsuite-virtual-category/Setup/VirtualCategorySetup.php
```diff
@@ -205,6 +205,40 @@ public function updateVirtualCategoryRootTypeToInt(\Magento\Eav\Setup\EavSetup $
         return $this;
     }
 
+    /**
+     * Migration from 1.4.0 to 1.4.0 :
```
romainruaud: from 1.4.0 to 1.4.1 I think :)

PR #1397, opened 2019-04-26
Rationalize the join, and apply it only if the attribute is existing.

romainruaud:
Tested working on my multi store environment. Behavior of "use_store_positions" is still working.

Tested by returning directly "null" with `getUseStorePositionsAttribute`, indexation is not broken. Of course in this case, results are a bit messy because you can have several products on position 1 (a product is 1 on global settings, another is 1 for a specific store). But this would imply having product positions already set into the table without the attribute existing, which should rarely occurs.


rbayet:
Hello @romainruaud,

Can you duplicate your PR for 2.2.x / 2.6.x-dev branch ?

Regards,
Richard

PR #1405, opened 2019-05-11
Use correct method to add / remove class name

aapokiiso:
In admin, it seems the global `$` refers to PrototypeJS instead of jQuery. This `js.phtml` is otherwise aware of this, but it still assumes in one method that `$` refers to jQuery. This causes the following error:
```
Uncaught TypeError: panel.removeClass is not a function
    at checkOptionsPanelVisibility ((index):1371)
    at bindAttributeInputType ((index):1378)
    at HTMLDocument.<anonymous> ((index):1688)
    at fire (jquery.js:3232)
    at Object.add [as done] (jquery.js:3291)
    at jQuery.fn.init.jQuery.fn.ready (jquery.js:3542)
    at jQuery.fn.init (jquery.js:2967)
    at new jQuery.fn.init (jquery-migrate.js:231)
    at jQuery (jquery.js:75)
    at (index):1687
```

### DIFF
--- a/src/module-elasticsuite-catalog/view/adminhtml/templates/catalog/product/attribute/js.phtml
+++ b/src/module-elasticsuite-catalog/view/adminhtml/templates/catalog/product/attribute/js.phtml
@@ -51,11 +51,11 @@
 
                 if($('frontend_input') && ($('frontend_input').value=='select' || $('frontend_input').value=='multiselect')){
                     panel.show();
-                    panel.addClass(activePanelClass);
+                    panel.addClassName(activePanelClass);
                 }
                 else {
                     panel.hide();
-                    panel.removeClass(activePanelClass);
+                    panel.removeClassName(activePanelClass);
                 }
             }
         }


romainruaud:
Hello @aapokiiso 

Thank you for submitting this PR. @androshchuk can you review it and try to reproduce the error locally, and ensure that this PR is fixing it.

Regards

aapokiiso:
@androshchuk Sorry for not including the reproduction steps to the original PR! It happened when editing an attribute with the frontend input "Dropdown" in the admin, but I don't remember anymore exactly how. 

Even without steps to reproduce I think it's possible to understand what's wrong just from the template itself. It tries to use the jQuery methods `addClass` and `removeClass` on a PrototypeJS object on lines 54 and 58.

romainruaud:
Since our file is a copy-paste of the legacy one : https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Catalog/view/adminhtml/templates/catalog/product/attribute/js.phtml

We will not fix this issue since it's not fixed by Magento themselves.

Regards

Issue #866, opened 2018-04-13
Configurable products shown in virtual category as discounted products, when they are not


dverkade:
Configurable products shown in virtual category as discounted products, when they are not

### Preconditions
Magento Version : 2.2.3
ElasticSuite Version : 2.5.4
Environment : Production mode

### Steps to reproduce
1. Create a configurable product with different sizes. The price needs to be different per simple product.
2. Create a virtual category where you have all products which are in stock & are discounted.

### Expected result
1. Configurable products where the simple price is different and which do not have a special price should not be shown in the virtual category. 

### Actual result
1. Products are shown in the virtual category. 

![image](https://user-images.githubusercontent.com/1058480/38754023-8d804f46-3f60-11e8-8e06-48843497a495.png)
![image](https://user-images.githubusercontent.com/1058480/38754047-9a36701c-3f60-11e8-911c-4e850c1793f8.png)

afoucret:
Confirmed when simple price are not all equals.

romainruaud:
Not sure how we could fix this one easily. 

For now we rely on the `catalog_product_index_price` table and this does not seem to be enough to differentiate these 2 cases : 

- product has many children of different prices. These children have no special price. => should NOT have `is_discounted`

- one of the children product has a special price. => product should have `is_discounted`

any advice @afoucret  ?

moejoee91:
Hello, 
any fix planned for this? or any work around available for now ?

afoucret:
Hi @moejoee91, 

I am currently braimstorming on this one. The technical part is quite easy but this bug highlight a functional issue I will explained later. Without a clean functional solution it will be useless to solve this issue.

amenk:
We had this issues mentioned in #1140 and https://magento.stackexchange.com/questions/247097/price-filter-returns-wrong-results-price-ist-stored-as-0-insteal-of-null-and-i with the price index in 2.1.6 -- now we are on 2.1.8 and believe the price index is in a bit better shape

Simples work fine now, but configurables are wrongly tagged as is_discounted

* min_price and max_price seem to be right
* but price is now 0 in the index for configurables.

and  is_discounted becomes true, which it should not.

The question we are asking ourselves now is, what should be the price for configurables in the price index -- actually I think "0" is quite correct, because normally a configurable does not have a price. But how could the is_discounted indexing then work of elastic suite?

amenk:
ping @romainruaud  ;-)

amenk:
[...]

PR #1411, opened 2019-05-22
#866 Proof Of Concept: For configurables, determine is_discount proper

amenk:
We iterate in the price index over all simples and determine if any
is discounted, then set the configurable to discounted.

### DIFF
--- a/src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/PriceData.php
+++ b/src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/PriceData.php
@@ -68,7 +68,11 @@ class PriceData implements DatasourceInterface
             $indexData[$productId]['price'][] = [
                 'price'             => $price,
                 'original_price'    => $originalPrice,
-                'is_discount'       => $price < $originalPrice,
+                'is_discount'       => $priceModifier->getIsDiscount($priceDataRow,
+                    $productId,
+                    $storeId,
+                    $priceDataRow['customer_group_id']
+                ),
                 'customer_group_id' => $priceDataRow['customer_group_id'],
             ];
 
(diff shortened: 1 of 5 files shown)


amenk:
This pull request has one problem: If a simple is changed, the index is not triggered for the configurable. Any clue how I can trigger that?

amenk:
Se also https://magento.stackexchange.com/questions/275867/how-to-trigger-an-index-for-a-configurable-if-the-simple-price-index-was-changed

amenk:
I would be happy for comments

amenk:
@romainruaud Shall I proceed working on this or do you have any input?

romainruaud:
The concept looks fine for me, but imhp there is a huge potential performance issue with using `$this->productRepository->getById` and `getUsedProducts` on each configurable product that is being indexed.

It would be way better if we can do it once for each batch of products that are being indexed.

Regards

amenk:
Thanks for your comment. Unfortunately we decided to work with real attributes which are updated on the fly instead of using the elastic suite index here in the future.

I will probably not follow up that PR further.

Mbertu:
I could take care of the development of this patch. I need this feature for a project.

nkajic:
Hi @Mbertu 
I have tested @amenk's PR and it works. But there are huge performance issues.

Have you tried to solve performance issues with @romainruaud's guidance or do you have any new informations?

Thanks,
N

Mbertu:
Hi @nkajic thanks for the reply.
Yes, it was an easy fix because in `addData` just before the `foreach` on `$priceData` I already have all products ids and I can get them all in one single query.
I can organize a new PR so you con check the code. 
Thanks,
Michele

nkajic:
Hi @Mbertu 
I would appreciate if you can share your solution of even better create PR for it.
Thanks,
N

Issue #1420, opened 2019-06-05
Virtual category rule - "is not one of" condition not working


alexayg:
### Preconditions

Magento Version : Magento CE 2.2.5
ElasticSuite Version : 2.5
Elasticsearch version : 5.6.10
Magento Environment : dev

### Steps to reproduce

1. Create a new virtual category
2. Select a "Virtual Category Root" for it
3. Under "Virtual rule" add the following condition: "Category **is not one of** category_ids_list"

### Expected result

1. The virtual category page should not display the products from the "category_ids_list" categories.

### Actual result
1. The virtual category page is displaying all the products from the "Virtual Category Root" categories.

romainruaud:
Hello @alexayg 

ElasticSuite 2.5 is quite old now, and I'm afraid that we cannot continue supporting it as part of our Open Source support commitment here on github.

That being said, thank you for reporting this bug, we'll try to reproduce it on most recent release and to bring a fix for this one if this is still accurate.

@androshchuk can you try to reproduce this issue on 2.8 version of ElasticSuite, and if still occuring, work on a fix ?

Regards

romainruaud:
Fixed by associated PR

PR #1426, opened 2019-06-11
fix condition "category is not one of category ids list"

androshchuk:
fix for https://github.com/Smile-SA/elasticsuite/issues/1420

### DIFF
--- a/src/module-elasticsuite-virtual-category/Model/Rule/Condition/Product.php
+++ b/src/module-elasticsuite-virtual-category/Model/Rule/Condition/Product.php
@@ -95,12 +95,12 @@ class Product extends \Smile\ElasticsuiteCatalogRule\Model\Rule\Condition\Produc
     /**
      * Build a search query for the current rule.
      *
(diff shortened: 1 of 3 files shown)


src/module-elasticsuite-virtual-category/Model/Rule/Condition/Product.php
```diff
@@ -82,21 +89,23 @@ public function __construct(
...
      *
-     * @param array $excludedCategories Categories excluded of query building (avoid infinite recursion).
+     * @param array      $excludedCategories  Categories excluded of query building (avoid infinite recursion).
+     * @param int|string $virtualCategoryRoot Category root for Virtual Category.
      *
-     * @return QueryInterface
+     * @return QueryInterface|null
      */
-    public function getSearchQuery($excludedCategories = [])
+    public function getSearchQuery($excludedCategories = [], $virtualCategoryRoot = ''): ?QueryInterface
```
romainruaud: Why an empty string as default value ? Seems weird, `null` could be better, don't you think so ?

PR #1429, opened 2019-06-12
Updated to treat pageSize of false as intended in Magento core

kirkmadera:


romainruaud:
Hello @kirkmadera 

We'll not merge this one since we rollbacked the default case of `_pageSize = 0` : 

https://github.com/Smile-SA/elasticsuite/blob/2.8.x/src/module-elasticsuite-catalog/Model/ResourceModel/Product/Fulltext/Collection.php#L535

There's no point to cap the size to 10000 (especially if we hardcode it). Elasticsearch will handle automatically the number of results according to his configuration.

Regards

kirkmadera:
Looks like this should also resolve this. Thanks for the update

PR #1430, opened 2019-06-13
fix blacklist in products list preview and sorting

androshchuk:


romainruaud:
It was important for me to add this because it will ensure an exact preview, including the optimizers associated to the category, if any.

But yes, I made a mistake and this is also applying the blacklist, which should not be the case.

Instead of removing the line, I'd go for skipping these methods if we are on the admin store : 

https://github.com/Smile-SA/elasticsuite/blob/2.8.x/src/module-elasticsuite-catalog/Model/Product/Search/Request/Container/Filter/SearchBlacklist.php#L55

https://github.com/Smile-SA/elasticsuite/blob/2.8.x/src/module-elasticsuite-virtual-category/Model/Product/Search/Request/Container/Filter/CategoryBlacklist.php#L56

PR #1440, opened 2019-06-25
Solve array to string conversion issue

arnoudhgz:
When filtering on array values e.g. with color swatch you would get an
array to string conversion bug in Magento Swatches
loadFirstVariationWithImage().

By setting up the array differently and than looping trough each value
the `array_diff_assoc` check will not throw an error anymore.


romainruaud:
@arnoudhgz can you provide us a scenario when this array to string conversion issue is happening ?

regards

arnoudhgz:
@romainruaud when the configurable product does not have an image itself and you filter on multiple values e.g. `?color%5B0%5D=Blue&color%5B1%5D=Yellow`

niccifor:
We had the same issue, and this pull request solves it.

arnoudhgz:
@romainruaud any update on this?

romainruaud:
I close, this has been stalled for too much time and not mergeable as-is .

PR #1471, opened 2019-07-15
fix cyrillic characters in layered navigation

androshchuk:


src/module-elasticsuite-catalog/view/frontend/web/js/attribute-filter.js
```diff
@@ -167,7 +167,7 @@ define([
         slugify: function(text) {
           return text.toString().toLowerCase()
             .replace(/\s+/g, '-')           // Replace spaces with -
-            .replace(/[^\w\-]+/g, '')       // Remove all non-word chars
+            .replace(/^\p{L}\p{Nd}/g, '')   // Remove all non-word chars
```
rbayet: I'm not sure the character "-" is included in \p{L}, so you might have to add it.
Same thing for "_" (which is present in the meta-character \w.
Finally you changed the whole meaning of regex from a repeated character range to a "begins by".

PR #1475, opened 2019-07-19
fix nice product url for virtual category

androshchuk:


romainruaud:
Rendering of the URL is fine, but, when I click on it, instead of being redirected to the product page I'm facing a fatal error : 

```
Uncaught TypeError: Return value of Smile\ElasticsuiteVirtualCategory\Controller\Router::match() must be of the type boolean, object returned in /var/www/magento/src/vendor/smile/elasticsuite/src/module-elasticsuite-virtual-category/Controller/Router.php on line 98
```

PR #1481, opened 2019-07-22
Made autocompelete searches fuzzy

aario:


romainruaud:
What's exactly the purpose of this PR ?

Please describe what is your use-case and why you did choose to fix it this way ?

aario:
I "Made autocompelete searches fuzzy". The use case is to have fuzzy autocomplete. Currently they are not fuzzy. Because when autocomplete query is sent, it's not in form of an array and code currently chooses exact search mode for it. I changed it to choose fuzzy instead.

romainruaud:
I close, this has been stalled for too much time and not mergeable as-is .

PR #1487, opened 2019-07-26
add option 'ignore_above' to mapping and refactoring code

androshchuk:


src/module-elasticsuite-core/Index/Mapping/Field.php
```diff
@@ -78,7 +78,7 @@ public function __construct($name, $type = self::FIELD_TYPE_KEYWORD, $nestedPath
     {
         $this->name       = (string) $name;
         $this->type       = (string) $type;
-        $this->config     = $fieldConfig + $this->config;
+        $this->config     = array_merge($this->config, $fieldConfig);
```
romainruaud: Why replacing the + by an array_merge ?

src/module-elasticsuite-core/Index/Mapping/Field.php
```diff
@@ -200,14 +200,14 @@ public function getMappingProperty($analyzer = self::ANALYZER_UNTOUCHED)
         $fieldName    = $this->getName();
         $propertyName = $fieldName;
         $property     = $this->getMappingPropertyConfig();
-        $isDefaultAnalyzer = $analyzer == $this->getDefaultSearchAnalyzer();
+        $isDefaultAnalyzer = $analyzer === $this->getDefaultSearchAnalyzer();
 
-        if (isset($property['fields']) && !$isDefaultAnalyzer) {
+        if (!$isDefaultAnalyzer && isset($property['fields'])) {
```
romainruaud: why inverting the initial condition ?

src/module-elasticsuite-core/Index/Mapping/Field.php
```diff
@@ -320,21 +320,20 @@ private function getMultiFieldMappingPropertyConfig($analyzers)
...
 
         if ($this->isSearchable() || $this->isUsedForSortBy()) {
             // Default search analyzer.
             $analyzers = [$this->getDefaultSearchAnalyzer()];
-
-            if ($this->isSearchable() && $this->getSearchWeight() > 1) {
-                $analyzers[] = self::ANALYZER_WHITESPACE;
-                $analyzers[] = self::ANALYZER_SHINGLE;
-            }
+        }
```
romainruaud: Why closing the brace here ?

src/module-elasticsuite-core/Index/Mapping/Field.php
```diff
@@ -320,21 +320,20 @@ private function getMultiFieldMappingPropertyConfig($analyzers)
...
+        if ($this->isSearchable() && $this->getSearchWeight() > 1) {
+            $analyzers[] = self::ANALYZER_WHITESPACE;
+            $analyzers[] = self::ANALYZER_SHINGLE;
         }
 
-        if ($this->isFilterable() || empty($analyzers)) {
+        if (empty($analyzers) || $this->isFilterable()) {
```
romainruaud: Why inverting the condition ?

src/module-elasticsuite-core/Index/Mapping/Field.php
```diff
@@ -354,19 +353,26 @@ private function getFieldAnalyzers()
...
-        if ($this->getType() == self::FIELD_TYPE_TEXT && $analyzer == self::ANALYZER_UNTOUCHED) {
-            $fieldMapping['type']  = self::FIELD_TYPE_KEYWORD;
-        } elseif ($this->getType() == self::FIELD_TYPE_TEXT) {
-            $fieldMapping['analyzer'] = $analyzer;
-            if ($analyzer === self::ANALYZER_SORTABLE) {
-                $fieldMapping['fielddata'] = true;
-            }
-        } elseif ($this->getType() == self::FIELD_TYPE_DATE) {
-            $fieldMapping['format'] = implode('||', $this->dateFormats);
+        switch ($this->getType()) {
```
romainruaud: I'm not fan of switch/case, I prefer the original syntax with if/elseif (and it would also allow us to have a relevant git history).

src/module-elasticsuite-core/Index/Mapping/Field.php
```diff
@@ -354,19 +353,26 @@ private function getFieldAnalyzers()
...
+            case self::FIELD_TYPE_TEXT:
+                if ($analyzer === self::ANALYZER_UNTOUCHED) {
+                    $fieldMapping['type'] = self::FIELD_TYPE_KEYWORD;
+                    $fieldMapping['ignore_above'] = 256;
```
romainruaud: Use a constant instead of using hardcoded value

src/module-elasticsuite-core/Index/Mapping/Field.php
androshchuk: You're right, but I'm a fan of using array functions for array instead of addition operators and I followed phpstorm hints http://joxi.ru/Dr8OK11foqzWZA

src/module-elasticsuite-core/Index/Mapping/Field.php
androshchuk: Condition "!$isDefaultAnalyzer" execution costs less condition "isset($property['fields'])" .

src/module-elasticsuite-core/Index/Mapping/Field.php
androshchuk: Condition "$this->isSearchable()" was duplicated in previous logic construction http://joxi.ru/ZrJ0V44iwxM9P2 . When I changed it - the code became clean http://joxi.ru/12MN144ily0lMA

src/module-elasticsuite-core/Index/Mapping/Field.php
androshchuk: Condition "empty($analyzers)" execution costs less condition "$this->isFilterable()"

src/module-elasticsuite-core/Index/Mapping/Field.php
androshchuk: My mistake. I will correct

src/module-elasticsuite-core/Index/Mapping/Field.php
androshchuk: For me better use "switch/case" instance of many if/else. You're right about git history, but, in my opinion, we are losing in the code style and structure.

src/module-elasticsuite-core/Index/Mapping/Field.php
romainruaud: Yes but + and `array_merge` does not have exactly the same behavior.

Since this piece of code is legacy, and that changing it is not related to the issue you are fixing here, I'd recommend not modifying this line.

This can apply to everything else, I prefer that we don't change legacy code unless it's strictly needed for a bugfix.

src/module-elasticsuite-core/Index/Mapping/Field.php
romainruaud: That's what I had in mind, good spot. Ok for this one

romainruaud:
Ok, I marked some points as resolved.

What remains to be done : 

- the `+` which was replaced by an array_merge : I prefer to keep the old version.

- the `256` value which is hardcoded.

Thank you

PR #1499, opened 2019-08-06
fix display swatches attribute with manual sort

androshchuk:


romainruaud:
2 things on this one : 

- as we discussed, I'd prefer a fix without additional refactoring, for clearness of what this commit is really intended for.

- I notice that you did replace a `if !isset(xxx) or !isset(yyy)` by `if !isset(xxx,yyy)` which is different behavior because imho when you pass two parameters to isset, it creates an "AND" instead of an "OR"

romainruaud:
Looks good to me, no regression,

PR #1516, opened 2019-08-27
Check if index already exists

cmuench:
We added that code in our project to prevent this annoying issue. In our case we install Magento in a build pipeline. Sometimes up to ten times the build crashes.
The fix adds a check if the index already exists.
Please have a look on that. Seems that this fixes issue #1086


romainruaud:
@rbayet can you review this one ? Actually it's unclear for me if the "create index, populate it, switch the alias at the end, delete the old index" routine will still work properly in any cases with this fix, because it'll prevent from creating the new index.

But, probably it keep working because of this : https://github.com/Smile-SA/elasticsuite/blob/2.8.x/src/module-elasticsuite-core/Indexer/GenericIndexerHandler.php#L109

Catching this exception will force the creation of the index, that may explain why it's still getting created.

PR #1543, opened 2019-09-16
Bugfix: Ordering is not applied

thampe:
Hi,

I had a strange issue that the sorting was not applied to the collection. The reason for this is, that the collection has already rendered the filters, without an order. The ordering comes later through the Toolbar Block but it was not applied.

This tiny PR should fix it. 

Thanks for again these awesome module!

\- Thomas


jaxtheking:
This fix worked for me. Now filters are indeed applied before rendering.
Is there a plan to merged this in at some point?
Thanks, Luca.

PR #1545, opened 2019-09-18
Fixes #1488 Fixed indexing children_attributes for configurable produ…

### LINKED ISSUE #1488: Parent products excluding children_attributes
At the moment, this is a question, not a bug.

Tested on Magento 2.3.2 and ES 2.8.1

I noticed that in one site running elasticsuite, the results in the layered navigation didn't seem correct. For example, on a particular category, there should have been around 10 results for the colour `Orange`.

After investigation, I looked in to the data store in elasticsearch, and could see that for the configurable product in question, it included no records for `children_attributes`. 

```
"children_attributes" : [
            "indexed_attributes"
          ],
```

However, it did  have `children_ids`

```
          "children_ids" : [
            131036,
            131037,
            131038
          ],
```

So, I debugged this by moving the data to a vanilla install and testing again. Same result. 

I then created a new configurable product, assigning the same children, and this time, the new product included the `children_attributes`, but the original configurable product remained with these missing.

New product data:

```
          "children_ids" : [
            131036,
            131037,
            131038
          ],
          "children_attributes" : [
            "image",
            "description",
            "short_description",
            "brand",
            "option_text_brand",
            "guarantee",
            "option_text_guarantee",
            "lead_time",
            "option_text_lead_time",
            "search_colour",
            "option_text_search_colour",
            "height_adjustable",
            "option_text_height_adjustable",
            "seat_fabric",
            "option_text_seat_fabric",
            "indexed_attributes"
          ],
```

Old product data:

```
          "children_ids" : [
            131036,
            131037,
            131038
          ],
          "children_attributes" : [
            "indexed_attributes"
          ],
```

So, this confirms it's data related. I've checked for relationship differences, and there is nothing that I can find.

Just wondering if anyone has had similar experience, or any ideas what might be responsible

serg6854:
…ct while creating

### DIFF
--- a/src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/AttributeData.php
+++ b/src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/AttributeData.php
@@ -31,7 +31,13 @@ class AttributeData extends AbstractAttributeData implements DatasourceInterface
     /**
      * @var array
      */
-    private $forbidenChildrenAttributeCode = ['visibility', 'status', 'price', 'tax_class_id'];
+    private $forbidenChildrenAttributeCode = [
+        'visibility',
+        'status',
+        'price',
+        'tax_class_id',
+        'name',
+    ];
 
     /**
      * {@inheritdoc}
@@ -67,9 +73,7 @@ class AttributeData extends AbstractAttributeData implements DatasourceInterface
             }
         }
 
-        $indexData = $this->filterCompositeProducts($indexData);
-
-        return $indexData;
+        return $this->filterCompositeProducts($indexData);
     }
 
     /**
@@ -142,7 +146,6 @@ class AttributeData extends AbstractAttributeData implements DatasourceInterface
     private function addRelationData(&$parentData, $childAttributes, $relation)
     {
         $childAttributeCodes  = array_keys($childAttributes);
-        $parentAttributeCodes = array_keys($parentData);
 
         if (!isset($parentData['children_attributes'])) {
             $parentData['children_attributes'] = ['indexed_attributes'];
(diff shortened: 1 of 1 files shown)


src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/AttributeData.php
```diff
@@ -154,12 +153,7 @@ private function addRelationData(&$parentData, $childAttributes, $relation)
         );
 
         if (isset($relation['configurable_attributes']) && !empty($relation['configurable_attributes'])) {
-            $addedChildrenAttributes = array_diff(
-                $childAttributeCodes,
-                $this->forbidenChildrenAttributeCode,
```
romainruaud: Why removing the `forbidenChildrenAttributeCode` ?

serg6854: Because found that same (without $parentAttributeCodes) array_diff exists on 152 line.

romainruaud:
Before you take it @rbayet : for me there is something missing/broken : the "name" attribute get merged with this PR, but it was not the case before. It's better to have "name" unmerged especially for the sort by name.

rbayet:
> Before you take it @rbayet : for me there is something missing/broken : the "name" attribute get merged with this PR, but it was not the case before. It's better to have "name" unmerged especially for the sort by name.

I concur.

src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/AttributeData.php
rbayet: There are indeed two problems :
- if `$childrenAttributes` is no longer for a configurable only the attributes that exist solely in the children products, then 
- - it makes sense to remove the historic `array_diff` of line 157 to 161
- - but it would make also sense to remove the following `array_merge` of line 162 : I don't see the point of re-merging `$parentData['children_attributes']` into `$childrenAttributes`
- as stated by Romain in that comment https://github.com/Smile-SA/elasticsuite/pull/1545#issuecomment-541676432, now the `name` of children products are indexed into the parent, which might totally mess sorting products by name for configurables
- - so 'name' should be added to '$forbidenChildrenAttributeCode'
- - but what about bundle products ? don't they suffer **today** about that potentiel "sort by name" issue ?

I'm yet to fully understand the original issue (I haven't tried to reproduce it), but on a hunch, I think it's due to different expectations between (non-visible) children products made out of a configurable product vs children products existing in their own right handpicked to form a configurable product.

At this point, I would cover all bases having this new behavior being configurable via an admin option which is probably a bad idea.

src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/AttributeData.php
```diff
@@ -154,12 +153,7 @@ private function addRelationData(&$parentData, $childAttributes, $relation)
...
-                $parentAttributeCodes
-            );
-            $childrenAttributes = array_merge($addedChildrenAttributes, $parentData['children_attributes']);
+            $childrenAttributes = array_merge($childrenAttributes, $parentData['children_attributes']);
```
rbayet: No longer needed IMHO.

src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/AttributeData.php
```diff
@@ -31,7 +31,13 @@ class AttributeData extends AbstractAttributeData implements DatasourceInterface
     /**
      * @var array
      */
-    private $forbidenChildrenAttributeCode = ['visibility', 'status', 'price', 'tax_class_id'];
+    private $forbidenChildrenAttributeCode = [
+        'visibility',
+        'status',
+        'price',
+        'tax_class_id',
+        'name'
```
rbayet: Missing comma at the end of the line.

PR #1548, opened 2019-09-23
Fixes partially sorted product on frontend

z4kko:
Since elasticsearch is already delivering sorted result, querying mysql without sorting will return entities sorted by ascending entity_id; therefore, if one or more entity has an id too big to fit in the pagination chunk, it won't appear in the frontend results.

### DIFF
--- a/src/module-elasticsuite-catalog/Model/ResourceModel/Product/Fulltext/Collection.php
+++ b/src/module-elasticsuite-catalog/Model/ResourceModel/Product/Fulltext/Collection.php
@@ -414,6 +414,10 @@ class Collection extends \Magento\Catalog\Model\ResourceModel\Product\Collection
         }
 
         $this->getSelect()->where('e.entity_id IN (?)', ['in' => $docIds]);
+        if ($docIds) {
+            $ids = implode(',', $docIds);
+            $this->getSelect()->order(new \Zend_Db_Expr("FIELD(e.entity_id, $ids)"));
+        }
         $this->originalPageSize = $this->_pageSize;
         $this->_pageSize = false;
 


romainruaud:
@z4kko can you give me a clear scenario for reproducing the issue that this PR is supposed to fix ?

It's not clear for me actually.

Regards

z4kko:
Hello Romain,

sure i can.

A category has, say, 300 products, and it is paginated by 60 in frontend.
Most of his products have entity_id in mysql like 100,101,102..but some of
then are newly imported, and their entity_id are like 3000,3001,3002.

Now the admin decide that product 3000 and 3001 have to be displayed in the
first and second position of that category, ad he uses the drag-drop from
backend.
The positions are now saved in elasticsearch in the nested category.

When the category is displayed in frontend, the
method \Smile\ElasticsuiteCatalog\Model\ResourceModel\Product\Fulltext\Collection::_renderFiltersBefore
will collect a *sorted* response from elastic, and then invoke an *unsorted*
select against the product entity table in order to check if those ids
really exists.
The problem is, that of those 300 documents, the document 3000 and 3001 are
the bottom of the mysql response (which is the one that set variable _items
in the collection), despite being in the position 1 and 2 of the elastic
response, and therefore, not included in the first 60 products which the
collection will display.

Lately, the
method \Smile\ElasticsuiteCatalog\Model\ResourceModel\Product\Fulltext\Collection::_afterLoad
will iterate over the query response, checking if the (sorted) document ids
3000 and 3001 does exists in the variable originalItems, but they do not,
since the original items only contains the first 60 entity id out of mysql.

Hope i've explained the problem better :)

Regards,



Gianluca Zucco

*Contacts*
Mobile *(+49) 1525 5974650* | Email <gianluca.zucco3@gmail.com> | LinkedIn
<http://www.linkedin.com/pub/gianluca-zucco/5a/903/486> |


Il giorno mar 24 set 2019 alle ore 09:52 Romain Ruaud <
notifications@github.com> ha scritto:

romainruaud:
What seems strange to me in your case, is that we are not supposed to fetch the 300 products from MySQL if the pagination is said to display only 60.

When doing the "SELECT where entity_id IN (...)" we should have only 60 entity_id, including of course the two top-position products.

It's not the case on your environment ?

torreytsui:
Hey guys, I have raised a PR - with pretty much the same fix, but with a different perspective to the same problem.

https://github.com/Smile-SA/elasticsuite/pull/1632

I hope that it can convince that it is a real issue.

romainruaud:
Replaced by #1779

PR #1558, opened 2019-09-26
Supporting extra parameters/config for metrics

rbayet:
not available through the XML declaration though.

A case could be made that 
- "config" could be named "parameters", but "field" is already a parameter in itself
- not all metrics aggs do actually use a "field" parameter (see currently unsupported [weight avg aggregation](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-weight-avg-aggregation.html))

Maybe a broader rework of metrics aggs is necessary in the long run.

### DIFF
--- a/src/module-elasticsuite-core/Search/Request/Aggregation/Metric.php
+++ b/src/module-elasticsuite-core/Search/Request/Aggregation/Metric.php
@@ -36,17 +36,24 @@ class Metric extends \Magento\Framework\Search\Request\Aggregation\Metric implem
     private $name;
 
     /**
+     * @var array
+     */
+    private $config;
+
+    /**
      * Constructor.
      *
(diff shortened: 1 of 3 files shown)


src/module-elasticsuite-core/Search/Request/Aggregation/Metric.php
```diff
@@ -64,4 +71,12 @@ public function getName()
     {
         return $this->name;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getConfig()
```
romainruaud: This could be added into the MetricInterface, no ?

rbayet:
> Please add getConfig() to the interface

Indeed, had forgot to commit the interface because of another unrelated change in it.

PR #1594, opened 2019-10-27
Update InventoryData.php

southerncomputer:
Fix invalid return type when deprecated is actually used. Removing msi inventory modules causes this function to not return InventoryDataInterface.


rbayet:
Hello @southerncomputer,

Can you check the merge base branch ?
Or rebase your patch branch against 2.8.x if that is indeed the correct target branch ?

Regards,

southerncomputer:
to master?

romainruaud:
nope, fixes should end up in 2.8.x branch

Issue #1590, opened 2019-10-24
Replace jquery/ui with the new modular approach


Quazz:
**Describe the solution you'd like**
Replace each instance of jquery/ui dependency with the relevant modular approach.

The reason this is desirable is that it improves performance, especially on lower end devices. (less javascript to process is alway good!)

**Additional context**
Magento 2.3.3 changed jquery-ui to be split up into seperate modules. Extensions should now update their scripts to no longer depend on the bulky jquery-ui, but only the modules that it requires

Additional info can be found in this tweet: https://twitter.com/vova_yatsyuk/status/1182192753335324673?s=09

romainruaud:
Good catch @Quazz !

Any chance you provide some help on this one ? :)

Quazz:
@romainruaud Haven't fully tested it, but as far as I can tell, just replace jquery/ui with jquery-ui-modules/widget in the following locations:

https://github.com/Smile-SA/elasticsuite/blob/2.8.x/src/module-elasticsuite-catalog/view/frontend/web/js/range-slider-widget.js#L18

https://github.com/Smile-SA/elasticsuite/blob/2.8.x/src/module-elasticsuite-core/view/frontend/web/js/form-mini.js#L24


I don't believe anything other than Widget is used of jquery-ui, but I could be mistaken. Haven't logged any complaints in the browser console so far, in the instances I could think of. Looking over the code, it seems to be accurate, too.

romainruaud:
@androshchuk can you check that the proposed replacement is working ?

androshchuk:
@romainruaud it's working.

@Quazz thank you for your help.

Regards

rbayet:
BC with EE/CE M2 2.3.* to be tested.
I don't fancy introducing a bug for everyone but people using the 2.3.3.

Quazz:
Perhaps the 'delegated dependency' that the tweet mentions is the better approach then?

`Magento_Ui/js/modal/modal` instead of `jquery/ui`

androshchuk:
@rbayet I checked way use Magento_Ui/js/modal/modal instead of jquery/ui in  M2 CE 2.3.0 and M2 CE 2.3.3.

Everything works fine.

rbayet:
Solved by PR #1597

PR #1597, opened 2019-10-29
Feature #1590 replace jquery-ui with jquery-ui-module

androshchuk:


rbayet:
Build on Travis is green but the status has not been updated correctly, because reasons.

PR #1600, opened 2019-10-29
Update InventoryData.php to work with both return types

southerncomputer:
Update InventoryData.php to work with both return types: deprecated and inventorydatainterface...

### DIFF
--- a/src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/InventoryData.php
+++ b/src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/InventoryData.php
@@ -85,7 +85,7 @@ class InventoryData implements DatasourceInterface
      *
      * @return InventoryDataInterface
      */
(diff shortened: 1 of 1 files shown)


rbayet:
Self-note: if validated, squash and merge.

rbayet:
@romainruaud Any philosophical requirement to also change the PHPdoc return type ?

romainruaud:
I did not see anything non-working on this piece of code. Removing the return type should not break anything either.

But why do you want to remove it exactly @southerncomputer ? what's the use case ?

southerncomputer:
We were having extremely long reindex times with the buggy MSI implementation in ee 2.3.3 - so we removed that library and only have cataloginventory which results in Deprecated::Class which is not compatible with inventoryinterface return type.. Throws an error with php 7.3.11 :(

romainruaud:
Okay then it could be due to PHP 7.3 :)

We'll try to merge this one asap

PR #1602, opened 2019-10-30
Add reverse nested bucket and moving function pipeline.

romainruaud:


src/module-elasticsuite-core/Search/Request/Aggregation/Pipeline/MovingFunction.php
```diff
@@ -0,0 +1,102 @@
...
+     *
+     * @return string
+     */
+    public function getGapPolicy()
+    {
+        return $this->gapPolicy;
+    }
+
+    /**
+     * @return int
```
rbayet: Add function description ?

romainruaud: yes, something like "Get Pipeline Window" ? :D

rbayet:
Otherwise, behaves as expected.

romainruaud:
I updated the comments :)

Issue #1581, opened 2019-10-16
VirtualCategory with no rules doesn't load in backend (HTTP 500) w/ Medium size catalog


denis-zyk:
VirtualCategory with no specific rules (i.e. rule value serialised in DB table `catalog_category_entity_text` looks like `{"type":"Smile\\ElasticsuiteVirtualCategory\\Model\\Rule\\Condition\\Combine","attribute":null,"operator":null,"value":"1","is_value_processed":null,"aggregator":"all"}`) cannot be loaded in Catalog -> Categories admin backend causing HTTP 500 error.

### Preconditions
Magento 2.3.2 CE
ElasticSuite 2.8.1
70k+ products (~74k in our case)
PHP 7.1.32
Redis
MariaDB 10.1.37

Magento Version : 2.3.2 CE

ElasticSuite Version : 2.8.1

Environment : checked in both modes: Developer and Production

Third party modules :
Amasty_BannersLite
Amasty_Base
Amasty_Conditions
Amasty_CommonRules
Amasty_Preorder
Amasty_Rgrid
Amasty_Rules
Amasty_RulesPro
Amasty_ShippingArea
Amasty_Shiprules
Amazon_Core
Amazon_Login
Amazon_Payment
Anowave_Package
Anowave_Ec

### Steps to reproduce

1. Go to Catalog -> Categories
2. Click on "Add Subcategory"
3. Set "Virtual category" to "Yes"
4. DO NOT change / add any category rules
5. Save new category
6. Try loading newly created category in backend

### Expected result

1. Category loads in the backend (where you can change/configure the category)

### Actual result

1. Request fails with HTTP 500 error (reaching max allowed RAM limit)

After trying the same steps as above with smaller number of products (e.g. 500), everything loads with no issues.
Also, when creating VirtualCategory AND setting some category rules at the creation time yields no issues with loading such category afterwards.

rbayet:
Hello @denis-zyk,

I can somehow reproduce this issue on a CE 2.3.2 + Elasticsuite 2.8.1 with Luma + the medium performance toolkit profile (44k products in the catalog).

What is strange is that : 
- there are no errors when initially saving such a category
- there are no errors when the category edit form is displayed again and the switch to virtual category is performed and saved

But indeed, if changing back to the root category and coming back to the category, the error occurs.

But I've managed to display again such a category on and off.
We'll try to be looking for that.

Regards,

denis-zyk:
[...]

PR #1603, opened 2019-10-30
Fixes #1591 Explicit fulltext collection page size

### LINKED ISSUE #1591: 500 Error on show more filter results 1.6M skus
### Preconditions

Single instance (Apache2, Percona, Elasticsearch), 14 CPU's | 64GB RAM | SSD
Magento Version : 2.3.2 CE
PHP: 7.2
Elasticsearch min_heapsize: 16 GB
PHP memory_limit: 2048 MB
1.6M simple products

ElasticSuite Version : 2.8.2

Environment : Developer

We have around 1.6 M sku's with 4-6 active layered navigation filters with each between 100 - 600 entries. When your result set is large > ~ 50.000 skus then the show more button (and search) which does the ajax request to get the additional filters get's a 500 error.

First error was `Result window is too large, from + size must be less than or equal to: [100000] but was [1621521]. `
So we increased the shards to 20 as per #599 & reindexed.

Then we got a PHP Fatal error:
`Allowed memory size of 792723456 bytes exhausted (tried to allocate 1319170176 bytes) in /var/www/html/magento2/vendor/elasticsearch/elasticsearch/src/Elasticsearch/Connections/Connection.php on line 289`

So we increased the php memory_limit to 2048 MB.

Now we have the exhaustion of the database:
`Allowed memory size of 2147483648 bytes exhausted (tried to allocate 20480 bytes) in /var/www/html/magento2/vendor/magento/framework/DB/Statement/Pdo/Mysql.php on line 91`

It is obviously the size of the result set combined with the amount of navigation filters. On our 1.9.4 setup we limited the resultset of the filters to 200 for performance reasons but we would obviously prefer not to do that. Any ideas on what might cause the exhaustion? I figured as you guys have experience with larger magento 2 instances you might know the pitfalls. 

Thanks!

rbayet:
to avoid a pageSize of getSize() set in the collection ie all matching products.
Same issue as #1581 and same ugly fix for the time being.

### DIFF
--- a/src/module-elasticsuite-catalog/Controller/Navigation/Filter/Ajax.php
+++ b/src/module-elasticsuite-catalog/Controller/Navigation/Filter/Ajax.php
@@ -116,6 +116,8 @@ class Ajax extends \Magento\Framework\App\Action\Action
 
         $this->applyFilters();
 
+        $this->layerResolver->get()->getProductCollection()->setPageSize(1);
+
         return $this;
     }
 


rbayet:
@romainruaud see alternative : #1604

PR #1608, opened 2019-11-05
Backport attributes administration script from upstream

mzanibelli:
Hi :smile: ,

~I stumbled upon the error this morning and although I noticed [this PR](https://github.com/Smile-SA/elasticsuite/pull/1405), the fix seems to be official now and is a bit different.~

> Since our file is a copy-paste of the legacy one 

As I read this again, might be time for a full copy-paste again. Here is the force-push!

Thanks for this awesome module!

### DIFF
--- a/src/module-elasticsuite-catalog/view/adminhtml/templates/catalog/product/attribute/js.phtml
+++ b/src/module-elasticsuite-catalog/view/adminhtml/templates/catalog/product/attribute/js.phtml
@@ -14,354 +14,354 @@
 
 ?>
 <script>
-    require([
-        "jquery",
-        'Magento_Ui/js/modal/alert',
-        'Magento_Ui/js/modal/prompt',
-        "collapsable",
-        "prototype"
-    ], function(jQuery, alert, prompt){
+require([
+    "jquery",
+    'Magento_Ui/js/modal/alert',
+    'Magento_Ui/js/modal/prompt',
+    'uiRegistry',
+    "collapsable",
+    "prototype"
+], function(jQuery, alert, prompt, registry){
 
-        function toggleApplyVisibility(select) {
-            if ($(select).value == 1) {
-                $(select).next('select').removeClassName('no-display');
-                $(select).next('select').removeClassName('ignore-validate');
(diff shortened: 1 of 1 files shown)


rbayet:
Hello @mzanibelli,

Infortunately, you missed a beat in the first `if` in `bindAttributeInputType` for `$('frontend_input').value == 'text'`.

Could you also target 2.8.x as the merging branch instead of master ?
I guess you would have to cherry-pick that commit on a 2.8.x derived branch, so you might have to create a new PR ?

Regards,

src/module-elasticsuite-catalog/view/adminhtml/templates/catalog/product/attribute/js.phtml
```diff
@@ -14,354 +14,354 @@
...
-            } else {
-                return $('front_fieldset-wrapper');
+    }
+}
+
+function bindAttributeInputType()
+{
+    checkOptionsPanelVisibility();
+    switchDefaultValueField();
+    if($('frontend_input') && ($('frontend_input').value=='boolean' || $('frontend_input').value=='select' || $('frontend_input').value=='multiselect' || $('frontend_input').value=='price')){
```
rbayet: The (added by our module) condition for 'text' is missing.

mzanibelli: Indeed, what's the point to copy upstream otherwise :man_facepalming: 
Sorry, I was in a hurry on that day. Good catch.

mzanibelli:
After changing the PR's base on GH web UI and rebasing over 2.8.x, it seems to be clean and ready for merging. Let me know if anything else needs to be done. Cheers!

rbayet:
(Self-)Reminder: it's the template override
- which adds support for 'text' attributes as layered navigation filters
   - historically, it was to allow support for both 'text' and 'boolean' attributes but native M2 ended up supporting 'boolean' attr. on its own
- **when the swatches modules are disabled**
   - because Magento_Swatches overrides that file

PR #1625, opened 2019-11-19
Fixes #1619 fix price mapping for optimizer

### LINKED ISSUE #1619: Issues with optimizers "Based on attribute value"
I tried to used the optimizer "Based on attribute value", new feature in ES 2.6.3, and I faced several issues. Maybe some of them are bugs ?

### Preconditions
Magento Version : 2.3.2
ElasticSuite Version : 2.8
Environment : public demo http://demo.magento-elastic-suite.io/

### 1. Configuration of the attribute

When I create an optimizer "based on attribute value" the only attributes I can select (attributes displayed in the list) are attributes with the following config:
- "Type" = "Price"
- "Use in search Results Layered Navigation" = "Yes"

The fact that it's mandatory to display the attribute as a filter is an issue. For example it prevent the merchant to create a boost based on the margin, because we don't want to display the margin as a filter.

And also it seems a bit strange to force to display the attribute as a filter on "search result page" but not on "category page". On a business point of view, the boost per attribute value is useful on both pages, I don't see a reason to do a difference between them.

### 2. Preview of the boost

On the demo platform http://demo.magento-elastic-suite.io/ I tried to create a boost based on the "price" of the product, "scale function = linear", "boost multiply = 1", but in the "preview" window, nothing changed in the search results.
image

With "square root", I also have the same result.
And with "Logarithmic" the score is changed to the value 0 for all products.

### 3. What is the formula ?

This last one is not an issue, it's a question for ElasticSuite documentation: I would be interested to have some explanations about the formula used to calculate the score.
For example with a "linear" boost, multiplicator 1, do we do a simple multiplication of the score by the value of the attribute ? (in that case it would be very important to have no "zero" value)

androshchuk:


src/module-elasticsuite-catalog-optimizer/Model/Optimizer/Applier/AttributeValue.php
```diff
@@ -58,6 +58,10 @@ public function getFunction(ContainerConfigurationInterface $containerConfigurat
         $field       = $this->getField($containerConfiguration, $optimizer);
         $scaleFactor = (float) $optimizer->getConfig('scale_factor');
 
+        if ($field === 'price') {
+            $field = 'price_optimizer';
```
rbayet: Why not "price.price" ?

src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/PriceData.php
```diff
@@ -65,6 +65,7 @@ public function addData($storeId, array $indexData)
             $originalPrice = $priceModifier->getOriginalPrice($priceDataRow);
             $price         = $priceModifier->getPrice($priceDataRow);
 
+            $indexData[$productId]['price_optimizer'] = $price;
```
rbayet: But then the last price index row for the store has the last say ...

romainruaud:
was replaced by #1666

Issue #1629, opened 2019-11-26
Infinite recursion on category edit page


yutv:
The infinite recursion is caused in the in the Smile\ElasticsuiteVirtualCategory\Model\Rule::[getCategorySearchQuery](https://github.com/Smile-SA/elasticsuite/blob/2.8.3/src/module-elasticsuite-virtual-category/Model/Rule.php#L139) method.

### Preconditions

Magento Version : 2.3.3

ElasticSuite Version : 2.8.3

Environment : Production

Third party modules : all custom modules (app/code), theme (app/design) and other 3-rd party modules from composer.json are removed but the issue still remains.

### Steps to reproduce

Unfortunately it is not easy to provide an exact test case because it somehow relates on a specific category tree. There is a lot of virtual categories on the website, e.g.:

    Default Category
    - Some Category
    - Virtual Category 1
    - Virtual Category 2 (it has about 180 virtual subcategories)
    - Virtual Category 3
     ... etc

The Default Category is set as root for Virtual Categories.

When you try to open the "Virtual Category 2" edit page the website runs into infinite recursion (CPU is loaded on 100% and doesn't stop, you have to reload php-fpm).

### Expected result

1. category edit page should be loaded

### Actual result

1. **504 Gateway Time-out** error from nginx
2. 100% CPU load (I reloaded php-fpm manually after a couple of minutes)

### Some debug results

It looks like the issue in the Smile\ElasticsuiteVirtualCategory\Model\Rule::[getCategorySearchQuery](https://github.com/Smile-SA/elasticsuite/blob/2.8.3/src/module-elasticsuite-virtual-category/Model/Rule.php#L139) method.

I put the following code in the line 149 and got the following log https://pastebin.com/27XAMsd2. 

            if ($handle = fopen('/tmp/test.log', 'a+')) {
                $message = "category: id={$category->getId()}, path={$category->getData('path')}, "
                    . 'active: ' . ($category->getIsActive() ? 'Yes' : 'No') . ', '
                    . 'virtual: ' . ($category->getIsVirtualCategory() ? 'Yes' : 'No') . ', '
                    . 'hasChildren: ' . ($category->hasChildren() ? 'Yes' : 'No') . ', '
                    . 'excludedCategories=' . implode(',', $excludedCategories);
                fwrite($handle, $message . PHP_EOL);
                fclose($handle);
            }

[...]

PR #1654, opened 2020-01-02
1629 fix infinite recursion

yutv:
This PR fixes the Issue #1629 by passing the `$excludedCategories` variable by reference.

### DIFF
--- a/src/module-elasticsuite-virtual-category/Model/Rule.php
+++ b/src/module-elasticsuite-virtual-category/Model/Rule.php
@@ -136,7 +136,7 @@ class Rule extends \Smile\ElasticsuiteCatalogRule\Model\Rule implements VirtualR
      *
      * @return QueryInterface|null
      */
-    public function getCategorySearchQuery($category, $excludedCategories = []): ?QueryInterface
+    public function getCategorySearchQuery($category, &$excludedCategories): ?QueryInterface
     {
         $query         = null;
 
@@ -181,9 +181,9 @@ class Rule extends \Smile\ElasticsuiteCatalogRule\Model\Rule implements VirtualR
                 ->addIsActiveFilter()
                 ->addIdFilter($childrenIds)
                 ->addAttributeToSelect(['virtual_category_root', 'is_virtual_category', 'virtual_rule']);
-
+            $excludedCategories = [];
             foreach ($categoryCollection as $category) {
-                $childQuery = $this->getCategorySearchQuery($category);
+                $childQuery = $this->getCategorySearchQuery($category, $excludedCategories);
                 if ($childQuery !== null) {
                     $queries[$category->getId()] = $childQuery;
                 }
@@ -247,7 +247,7 @@ class Rule extends \Smile\ElasticsuiteCatalogRule\Model\Rule implements VirtualR
      *
      * @return QueryInterface
      */
-    private function getStandardCategoryQuery(CategoryInterface $category, $excludedCategories = []): QueryInterface
+    private function getStandardCategoryQuery(CategoryInterface $category, &$excludedCategories): QueryInterface
     {
         return $this->getStandardCategoriesQuery([$category->getId()], $excludedCategories);
     }
@@ -279,7 +279,7 @@ class Rule extends \Smile\ElasticsuiteCatalogRule\Model\Rule implements VirtualR
      */
     private function getVirtualCategoryQuery(
         CategoryInterface $category,
-        $excludedCategories = [],
+        &$excludedCategories,
         $virtualCategoryRoot = null
     ): ?QueryInterface {
         $query          = $category->getVirtualRule()->getConditions()->getSearchQuery($excludedCategories, $virtualCategoryRoot);
@@ -309,7 +309,7 @@ class Rule extends \Smile\ElasticsuiteCatalogRule\Model\Rule implements VirtualR
      *
      * @return \Smile\ElasticsuiteCore\Search\Request\QueryInterface
      */
-    private function addChildrenQueries($query, CategoryInterface $category, $excludedCategories = []): QueryInterface
+    private function addChildrenQueries($query, CategoryInterface $category, &$excludedCategories): QueryInterface
     {
         $childrenCategories    = $this->getChildrenCategories($category, $excludedCategories);
         $childrenCategoriesIds = [];
--- a/src/module-elasticsuite-catalog/Plugin/Rss/CategoryPlugin.php
+++ b/src/module-elasticsuite-catalog/Plugin/Rss/CategoryPlugin.php
@@ -59,8 +59,8 @@ class CategoryPlugin
 
         if ($category->getVirtualRule()) { // Implicit dependency to Virtual Categories module.
             $category->setIsActive(true);
-
-            $filterParam = $category->getVirtualRule()->getCategorySearchQuery($category);
+            $excludedCategories = [];
+            $filterParam = $category->getVirtualRule()->getCategorySearchQuery($category, $excludedCategories);
         }
 
         return $filterParam;
--- a/src/module-elasticsuite-virtual-category/Api/Data/VirtualRuleInterface.php
+++ b/src/module-elasticsuite-virtual-category/Api/Data/VirtualRuleInterface.php
@@ -33,7 +33,7 @@ interface VirtualRuleInterface
      *
      * @return \Smile\ElasticsuiteCore\Search\Request\QueryInterface
      */
-    public function getCategorySearchQuery($category, $excludedCategories = []);
+    public function getCategorySearchQuery($category, &$excludedCategories);
 
     /**
      * Retrieve search queries of children categories.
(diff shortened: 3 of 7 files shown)


yutv:
FYI: The Travis CI build failed because of issue https://github.com/phpmd/phpmd/issues/720.

rbayet:
@yutv We'll have a look at the PR but I'm not a fan of passing by reference.
I'd probably go a local memory cache way if possible.

rbayet:
Well, it does the job of limiting the size of the recursion, and it has not side effect in frontend / on the merchandiser preview query, which is good.

It has, though, a side effect on the query generated by the "Filterable attributes" UI component, at least, on the "categories" aggregation.
On the other hand : 
- the "categories" aggregation result is not used in that UI component
- there is a legacy "glitch" which **should impact** that UI component content but does not seem to (I'm being cautious here)

Here is the **glitch** : 
- when the merchandiser preview query is built, it behaves the same way as in frontend : the rule has a concrete store id context, and a virtual category root which is the same as the store category root (in a default Luma, that would the category "2") is ignored (see  \Smile\ElasticsuiteVirtualCategory\Model\Rule::getVirtualRootCategory)
```
    /**
     * Load the root category used for a virtual category.
     *
     * @param CategoryInterface $category Virtual category.
     *
     * @return CategoryInterface|null
     */
    private function getVirtualRootCategory(CategoryInterface $category): ?CategoryInterface
    {
        $storeId      = $this->getStoreId();
        $rootCategory = $this->categoryFactory->create()->setStoreId($storeId);

        if ($category->getVirtualCategoryRoot() !== null && !empty($category->getVirtualCategoryRoot())) {
            $rootCategoryId = $category->getVirtualCategoryRoot();
            $rootCategory->load($rootCategoryId);
        }

        if ($rootCategory && $rootCategory->getId()
            && ($rootCategory->getLevel() < 1 || (int) $rootCategory->getId() === (int) $this->getRootCategoryId($storeId))
        ) {
            $rootCategory = null;
        }

        return $rootCategory;
    }
```
- but when the query is built for the "Filterable attributes" UI component is built, the store id associated is most likely "0" (due to the "All store views" scope), and the `\Smile\ElasticsuiteVirtualCategory\Model\Rule::getRootCategoryId` method returns 0 as the store root category id
- - hence, 0 being different than 2, the Luma default root category id, the virtual category root "2" **is not ignored** and is being browsed

So one needs to be sure that the changes are not that detrimental to the glitchy behavior, and also test what happens when the virtual category root shared by most of virtual categories is **not** the root category id of the store.
For instance, a situation where the category tree would look like 
```
- 1/2 
- - 1/2/3 "All products"
- - 1/2/4 "All brands"
```
with sub-categories of "All brands" being virtual categories with "brand rules" having 1/2/3 as virtual category root.

Regards,

src/module-elasticsuite-virtual-category/Model/Rule.php
```diff
@@ -181,9 +181,9 @@ public function getSearchQueriesByChildren(CategoryInterface $rootCategory): arr
                 ->addIsActiveFilter()
                 ->addIdFilter($childrenIds)
                 ->addAttributeToSelect(['virtual_category_root', 'is_virtual_category', 'virtual_rule']);
-
+            $excludedCategories = [];
             foreach ($categoryCollection as $category) {
-                $childQuery = $this->getCategorySearchQuery($category);
+                $childQuery = $this->getCategorySearchQuery($category, $excludedCategories);
```
rbayet: Initializing/resetting the `$excludedCategories` should be done here.

=> This method is called to build the category facet/aggregation : each \Smile\ElasticsuiteVirtualCategory\Model\Rule::getCategorySearchQuery should therefore be independent, otherwise children virtual categories sharing the same virtual root category could simply "disappear" from the "categories" aggregation.

rbayet: Initializing/resetting the $excludedCategories should be done inside the loop.
=> This method is called to build the category facet/aggregation : each \Smile\ElasticsuiteVirtualCategory\Model\Rule::getCategorySearchQuery should therefore be independent, otherwise children virtual categories sharing the same virtual root category could simply "disappear" from the "categories" aggregation.

yutv: Unfortunately moving the definition of the `$excludedCategories` variable inside the loop doesn't resolve the issue.
 - https://filebin.ca/58onD2YI91Rh/test_2020-01-14_inside-loop-incomplete.txt - incomplete log with `$excludedCategories` inside the loop. I reloaded the php-fpm after got the 504 Gateway Time-out error from nginx
 - https://filebin.ca/58on9IGIW7Kz/test_2020-01-14_before-loop.txt - full log with `$excludedCategories` outside loop

rbayet:
And actually, it has an adverse side effect on the main query filter. The optimization alters the boolean clauses that are built.

Typically, in the example of the originating issue (#1629), all (virtual) children categories of the current category will be ignored because their common virtual category root has already been checked and added to the excluded list.
Given : 
```
Default Category
- Some Category
- Virtual Category 1
- Virtual Category 2 (it has about 180 virtual subcategories)
- Virtual Category 3
```
and the fact we are on "Virtual Category 2" which has a virtual category root of "Default Category", as do all its virtual subcategories, the sequence is 
- **\Smile\ElasticsuiteVirtualCategory\Model\Rule::getCategorySearchQuery("Virtual Category 2")**
- "Virtual Category 2" is added to the exclusion list
- **\Smile\ElasticsuiteVirtualCategory\Model\Rule::getVirtualCategoryQuery("Virtual Category 2")**
- "Default Category" is not yet in the exclusion list, so the query is [rule] + getCategorySearchQuery("Default Category")
   - **\Smile\ElasticsuiteVirtualCategory\Model\Rule::getCategorySearchQuery("Default Category")**
   - "Default Category" is added to the exclusion list
   - **\Smile\ElasticsuiteVirtualCategory\Model\Rule::getStandardCategoryQuery("Default Category")**
   - **\Smile\ElasticsuiteVirtualCategory\Model\Rule::addChildrenQueries("Default Category")**
      - let's ignore "Virtual Category 1" and say we reach "Virtual Category 2 / Sub-virtual Cat A" which points to ("Default Category")
      - **\Smile\ElasticsuiteVirtualCategory\Model\Rule::getCategorySearchQuery("Virtual Category 2 / Sub-virtual Cat A")**
      - "Virtual Category 2 / Sub-virtual Cat A" is added to the exclusion list
      - **\Smile\ElasticsuiteVirtualCategory\Model\Rule::getVirtualCategoryQuery("Virtual Category 2 / Sub-virtual Cat A")**
      - "Default Category" is in the exclusion list, so query = ([rule] = null) + getCategorySearchQuery("Default Category")
      - **\Smile\ElasticsuiteVirtualCategory\Model\Rule::getCategorySearchQuery("Default Category")**
           - "Default Category" is in the exclusion list, null is returned
      - the query for "Virtual Category 2 / Sub-virtual Cat A" is thus query = ([rule] = null) + ([parent] = null)
[...]
- at some point, we reach **\Smile\ElasticsuiteVirtualCategory\Model\Rule::addChildrenQueries("Virtual Category 2")**
- but all its children have already been checked (and excluded) by the same behavior as for "Virtual Category 2 / Sub-virtual Cat A"
- - ie their query has not been collected because "Default Category" had already been checked and excluded


You can check that behavior on Luma, using either "2 / Default Category" or "3 / Gear" as virtual root category for categories arranged like this : 
- [attribute_set is "Bags" + style is "Tote"]
  - [attribute_set is "Bags" + style is "Laptop"]
  - [attribute_set is "Bags" + style is "Messenger"]
  - [attribute_set is "Bags" + style is "Backpack"]
  - [attribute_set is "Bags" + style is "Duffel"]

The expected results (because Elasticsuite makes sure all categories are "anchor") are all products validating "[attribute_set is "Bags" + style is "Tote"] OR [attribute_set is "Bags" + style is "Laptop"] OR [attribute_set is "Bags" + style is "Messenger"] OR [attribute_set is "Bags" + style is "Backpack"] OR [attribute_set is "Bags" + style is "Duffel"]" with the extra limitation of the virtual category root : 
- if the virtual category root is "2 / Default category" = condition ignored = any matching product of the store
- if the virtual category root is "3 / Gear" = condition applied = any matching product also located in and below the "Gear" category

Without the patch :
- this is the behavior show in frontend and in the preview
- the 0 size query generated for the filterable attribute list has no adverse effect on the query

With the patch 
- if the virtual category root is "2 / Default category"
  - for the filterable attributes list, the condition becomes "[attribute_set is "Bags" + style is "Tote"] below Default Category"
  - in the category preview and in frontend, as "2 / Default category" is ignored, conditions stays "[attribute_set is "Bags" + style is "Tote"] OR [attribute_set is "Bags" + style is "Laptop"] OR [attribute_set is "Bags" + style is "Messenger"] OR [attribute_set is "Bags" + style is "Backpack"] OR [attribute_set is "Bags" + style is "Duffel"]"
- if the virtual category root is "3 / Gear", it is ignored neither in admin nor in preview/frontend
  - condition becomes "[attribute_set is "Bags" + style is "Tote"] below Gear"

*Conclusion* : this fix only works as intended without side effects on a very specific case.

Issue #1562, opened 2019-10-01
Filters in custom page with custom layer doens't respect product collection


sedax90:
I've created a custom controller with a custom layer following this guide:
https://webkul.com/blog/how-to-create-layered-navigation-on-custom-page-in-magento-2/

I've added to my frontend/di.xml some modifications following elasticsuite-catalog module:

```
<type name="Dexanet\Dealer\Block\Navigation">
        <arguments>
            <argument name="filterList" xsi:type="object">categoryFilterList</argument>
        </arguments>
    </type>

   <type name="Dexanet\Dealer\Model\Layer">
        <arguments>
            <argument name="context" xsi:type="object">Magento\CatalogSearch\Model\Layer\Category\Context</argument>
            <argument name="productCollectionFactory" xsi:type="object">Magento\CatalogSearch\Model\ResourceModel\Fulltext\CollectionFactory</argument>
        </arguments>
    </type>

    <type name="Dexanet\Dealer\Model\Layer">
        <plugin name="createSortOrdersOnPrepareFrontCollection" type="Smile\ElasticsuiteCatalog\Plugin\LayerPlugin"/>
    </type>
```

The page now show correct products besed on the filtered collection defined in layer but products count and filters are wrong.
Do I still have to include something?

ionpan:
+1

tschirmer:
The Collection provider is replaced with a virtual type:
"Magento\Catalog\Model\Layer\Search\ItemCollectionProvider"

This won't act normally, 

It should be loading:
"\Magento\Catalog\Model\Layer\Category\ItemCollectionProvider" when no search options are enabled for that category. I'm having the same problem, but I haven't solved this yet, it'll need a pull request and a module update.

Birjitsinh:
I have same issue in magento 2.3 custom product collection

`<virtualType name="MagentoCoders\Vendor\Model\ResourceModel\Vendor\ProductCollection" type="Smile\ElasticsuiteCatalog\Model\ResourceModel\Product\Fulltext\Collection">
        <arguments>
            <argument name="searchRequestName" xsi:type="string">vendor_search_container</argument>
        </arguments>
    </virtualType>
    <virtualType name="MagentoCoders\Vendor\Model\ResourceModel\Vendor\ProductCollectionFactory" type="Magento\Catalog\Model\ResourceModel\Product\CollectionFactory">
        <arguments>
[...]

Issue #706, opened 2018-01-24
All products are displayed in each category


VincentMarmiesse:
Hello,

I have a strange issue, all products are displayed in each category, even if they are not in it.

If I disable all the `Smile*` modules and reindex, it's working fine but when I enable them and reindex, all products are displayed.

FYI, I am in multi-store configuration, I don't use virtual categories and the preview displays the right products in admin.

Any idea on what can cause this issue?

romainruaud:
Hello Vincent,

your issue looks like #681 . Unfortunately, we were quite unable to reproduce this issue or to obtain more details about it.

Could you please precise exactly each version of any components like it's askes in the issue template ?

Also, how are you importing the data ? Is it Magento2 sample data or real ones ? 

Regards

VincentMarmiesse:
Hello Romain,

I'll try to answer with the most details possible.

Magento Version : Open Source 2.2.2 
ElasticSuite Version : 2.4.0
Environment : Not sure about what you want, but I'm using PHP 7.0.25, MariaDB 10.2.9, Elasticsearch 2.4.6
Third party modules : Some third party modules and custom ones but none of them is supposed to touch this part.

I'm using real data, products are imported with FastSimpleImport2, categories are imported with the default Magento importer.

You can contact me directly if you want some access to our staging.

Thanks!

romainruaud:
Could you please try upgrading to latest Elasticsuite 2.4 version ? (the 2.4.4).

Little chances that it fixes your issue, but I prefer to ensure you are running the most recent version.

I will try to dig on this one since you are not the first user who is noticing it. However, we have many projects at Smile running Elasticsuite, also many "sandboxes installations" with just Elasticsuite and the sample data, and none has encountered this issue.

Could you check the following points : 

- catalog_category_product table is properly populated
- catalog_category_product_index is also built properly
- what do you have on the "category" field for products into the ES index

Regards

VincentMarmiesse:
Ok so here are the steps I followed:

- Disable all `Smile*` modules in `app/etc/config.php`
- Clear cache, setup upgrade, DI compilation, full reindex
- My products are correctly displayed
- Enable all `Smile*` modules in `app/etc/config.php`
- Set Smile ElasticSuite as search engine
- Clear cache, setup upgrade, DI compilation, full reindex
- All products are displayed in each category
- Run `composer update`
- `smile/elasticsuite` has been updated from 2.4.3 to 2.4.4
[...]

Issue #1084, opened 2018-09-17
Wrong product count for configurable products when filtering


simonrl:
When filtering a category page, I sometimes have the wrong items count when the list contains configurable products with children. Will provide more details below.

I understand that the problem I notice _may be_ intended behaviour. If this is the case, it would be nice if you could show me where in the code I could tackle this problem. I have looked into `$item->getCount()` in some places but I think this is too late in the process.

### Preconditions
Have a configurable product with some children that have different values for a filterable attribute. This attribute is not the configurable attribute. The configurable product does not have a value for this attribute.
Example:
attribute_a_filter
-> child-1: 30
-> child-2: 30
-> child-3: 45
The attribute is displayed as a range slider.
All children are "not visible individually", the configurable product has "catalog,search".
A screenshot of how the relevant attribute looks in Kibana is attached.
![bildschirmfoto 2018-09-17 um 17 35 05](https://user-images.githubusercontent.com/3898987/45633477-269a9a00-baa0-11e8-8370-0d95dbfaf977.png)

Magento Version : Opensource 2.1.14

ElasticSuite Version : 2.3.11

Environment : Elasticsearch 2.4.6, Developer Mode

Third party modules :
none relevant

### Steps to reproduce

1. have products configured as above
2. assign all products to a category
3. go to that category page
4. set range slider to a range for example: 30-35

### Expected result

1. I expect that the "products" number in the range slider shows "1", because the configurable product is the only visible product, and the toolbar also shows "1 result"

### Actual result

1. the "products" count in the range slider shows 2

It does not matter, BTW, if the product is the only one in this category or if there are more products. For example if the category are other configurable products, the calculation is also wrong for every child relevant to the filter.

Thanks for your time!

romainruaud:
@simonrl I don't understand how you are able to have several configurable products with the same SKU ? all are displayed as conf-180677 which seems a bit weird to me ...

simonrl:
Hi, 

sorry for the bad screenshot.

We have several store views, and the screenshot shows the product in each store view, which means that these are several indexes. It's the same product. The attribute values are "translated" (metric / imperial units), so it looks as if it's more than one product.

Thanks
Simon

rbayet:
Hi @simonrl,

Purely out of curiosity, and outside of the scope of your issue, if the children simple products are not visible individually and if that attribute is not the configurable attribute, what is the purpose of the attribute ? Is it inherently dependent on another configurable attribute ?

[...]

PR #1667, opened 2020-01-17
Category Results Missing Current Category Filter

PivitParkour94:
Currently the layer responsible for loading the category products doesn't take the category into consideration. Rather than re-inventing the wheel, the Magento/CatalogSearch module has already created an item provider that filters the data by the category. Changing the class the `Smile\ElasticsuiteCatalog\Model\Layer\Category\ItemCollectionProvider` is based on means a custom class that simply calls `addCategoryFilter` on the collection doesn't need to be made.

Possibly Relates to https://github.com/Smile-SA/elasticsuite/issues/706, https://github.com/Smile-SA/elasticsuite/issues/1562

### DIFF
--- a/src/module-elasticsuite-catalog/etc/di.xml
+++ b/src/module-elasticsuite-catalog/etc/di.xml
@@ -112,7 +112,7 @@
         </arguments>
     </virtualType>
 
-    <virtualType name="Smile\ElasticsuiteCatalog\Model\Layer\Category\ItemCollectionProvider" type="Magento\Catalog\Model\Layer\Search\ItemCollectionProvider">
+    <virtualType name="Smile\ElasticsuiteCatalog\Model\Layer\Category\ItemCollectionProvider" type="Magento\CatalogSearch\Model\Layer\Category\ItemCollectionProvider">
         <arguments>
             <argument name="collectionFactory" xsi:type="object">Magento\CatalogSearch\Model\ResourceModel\Fulltext\CollectionFactory</argument>
         </arguments>


PivitParkour94:
Also solves this https://github.com/Smile-SA/elasticsuite/issues/1084

romainruaud:
Hi,

all these layers virtualType are a bit messy (our part, but also the legacy Magento part).

It's very easy to introduce side effects without noticing it when changing things related to this, even small changes like this PR.

I understand the logic in this PR, and it seems better, but can you provide a (detailed) test scenario where Elasticsuite is not working properly for now, and that would be fixed by this PR ? Because for now, it's unclear for me.

Regards

PivitParkour94:
@romainruaud, I totally understand the concern, the issue that we have is we've taken an existing Magento 1 store, that is used heavily in production, migrated it to Magento 2 using the official data-migration tool and viewed the category page and it showed every product in the system, the category filtering was missing. It could have something to do with having no virtual categories set, or the specific `Products List Preview and Sorting` not being changed at all.

There are many moving parts to this equation and it's going to take a long time to isolate exactly what the issue is. It would be nice to give you the exact conditions, but unfortunately that's a little unrealistic on our side. We can continue to patch your code on our end and if any unexpected side effects occur we can make you aware of them, but ideally we would have this integrated into the core module so we don't have to continue maintaining the code we have from you guys.

romainruaud:
Well, most probably you are importing all your categories without the `is_anchor=1` attribute which is mandatory for Elasticsuite to work properly.

Please try to add this to all of your categories and have a look if it's running properly.

PivitParkour94:
Not all categories are anchor categories.  Our clients don't necessarily want to do their merchandising with anchored categories. 
That seems like a strange requirement.
Maybe forcing the `Display Settings` to all be on or off but making it clear either way might be a way to resolve this issue, if the intended use is to have an unused attribute `is_anchor` turned on but replaced with other functionality..

PivitParkour94:
Setting all categories `is_anchor` to 1 is not an acceptable solution. We will just use our patch unmaintained until further discussion can be had.

romainruaud:
We will not merge this, at least for now.

I'm open to re-open the debate about is_anchor for our next major release, but that would take a lot of work regarding how indexing category/products relationship is done.

Regards

PR #1669, opened 2020-01-17
Index all price data.

romainruaud:


rbayet:
@romainruaud Final price is taken "as is" from the price index table (unlike price), so you end up on bundles with a final price at 0 (but a price > 0) 

See for Luma's "Sprite Yoga Companion Kit" : 
```
mysql> select * from catalog_product_index_price where entity_id = 45;
+-----------+-------------------+------------+--------------+--------+-------------+-----------+-----------+------------+
| entity_id | customer_group_id | website_id | tax_class_id | price  | final_price | min_price | max_price | tier_price |
+-----------+-------------------+------------+--------------+--------+-------------+-----------+-----------+------------+
|        45 |                 0 |          1 |            0 | 0.0000 |      0.0000 |   61.0000 |   77.0000 |       NULL |
|        45 |                 1 |          1 |            0 | 0.0000 |      0.0000 |   61.0000 |   77.0000 |       NULL |
|        45 |                 2 |          1 |            0 | 0.0000 |      0.0000 |   61.0000 |   77.0000 |       NULL |
|        45 |                 3 |          1 |            0 | 0.0000 |      0.0000 |   61.0000 |   77.0000 |       NULL |
+-----------+-------------------+------------+--------------+--------+-------------+-----------+-----------+------------+
4 rows in set (0.00 sec)
```

Sélection_815

I don't know if that's expected.

romainruaud:
Yes I've seen it, there also other strange price computations for configurable but this PR will be enough for my purpose, since we will not use this data in Elasticsuite core implementation (at least for now), there is no risk in indexing such erratic values for now.

PR #1694, opened 2020-02-05
Update dynamic-rows.html

tinik:
add pagination for Layered Navigation Filters


romainruaud:
Wow, great one. It's that easy ?

I mean, does it work ok if I change some values on page 1, go to page 2, and then hit the Save button ?

tinik:
found a few bugs, need to fix and fill)

romainruaud:
@tinik any update ?

Issue #1037, opened 2018-08-15
Product Attribute Admin Sort Not Working With Options that have "


mdroidian:
If a Product Attribute has Options with quotes in them (IE - 3", 2-1/2") then Admin Sort does not work.

I've tested this on our Magento installation, as well as a separate default Luma.
Same issue applies on Category or Search with same product results.

I've re-indexed, cleared cache, and double-checked the db 'facet_sort_order' is indeed '_manual'

### Preconditions

Magento Version : 2.2.3 EE

ElasticSuite Version : 2.5.10
Elasticsearch Version : 2.4.6

Environment : Developer

Third party modules :

### Steps to reproduce

1. Create Product Attribute, Dropdown

2. Storefront Properties
Use in Layered Navigation -> Filterable(with results)
Use In Search Results Layered Navigation -> Yes
Facet sort order -> Admin sort

3. Add options
5, 1/32", 1/16", 1/4", 1/2", 1", T40, Universal, 1', 2', 2-1/2'

4. Add attribute to products.

5. Search or View Category with said products/attribute.

### Expected result

1. Sort Order is same as Admin

### Actual result

1. Sort Order seems random.  Does not follow sort order in admin/db.
Screenshots show what looks like 'Count', but with multiple tests it has varied.
![sc2](https://user-images.githubusercontent.com/3792666/44171848-a1832600-a098-11e8-99d1-0d35389616aa.JPG)
![sc1](https://user-images.githubusercontent.com/3792666/44171849-a1832600-a098-11e8-88d1-cfbb69fd2784.JPG)

sc1.jpg (backend)
sc2.jpg (frontend)

romainruaud:
Confirmed. I reproduce this one on latest Elasticsuite.

Attributes without any quotes in option values are still sorted properly when using "Admin Sort".

Will dig more into this one.

Thank you for reporting it.

afoucret:
PR #1039 have been merged and will be part of the next bugfix release.

PR #1705, opened 2020-02-11
Fixes #1703 remove htmlspecialchars

### LINKED ISSUE #1703: Encoding issue breaks admin sorting
### Preconditions

Magento Version : 2.3.3 / 2.3.4

ElasticSuite Version : 2.8.3 / 2.8.4

Environment : Both

Third party modules :

### Steps to reproduce

1. Create a dropdown attribute
2. Set sorting to `_manual`
3. Add attribute option labels with special characters (example: "> 1")

### Expected result

1. Attributes are sorted the same way they are in the admin 'grid'

### Actual result

1. Attributes with special characters are not sorted correctly

This is related to the attribute option label being encoded at https://github.com/Smile-SA/elasticsuite/blob/2.8.3/src/module-elasticsuite-catalog/Model/Layer/Filter/Attribute.php#L217 after that the `isset($items[$optionLabel])` won't pass because the encoded value is `&gt; 1` while the $items value will be `> 1`

Swahjak:
See #1703

### DIFF
--- a/src/module-elasticsuite-catalog/Model/Layer/Filter/Attribute.php
+++ b/src/module-elasticsuite-catalog/Model/Layer/Filter/Attribute.php
@@ -212,14 +212,13 @@ class Attribute extends \Magento\CatalogSearch\Model\Layer\Filter\Attribute
 
             if (!empty($options)) {
                 foreach ($options as $option) {
-                    if (isset($option['label'])) {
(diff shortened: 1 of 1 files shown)


romainruaud:
Thanks for this PR,

@rbayet can you do a final review on this one based on the initial scenario in #1037 ?

Issue #1503, opened 2019-08-12
Visual swatches values not displayed in the right order in the filters


LilyBergonzat:
When a filter is a visual swatch, its values won't respect the order defined in the attribute administration (for example: number of results).

### Preconditions
Have an attribute being a visual swatch, with multiple values.
Have each of those values return a different amount of products.

Magento Version : 2.3.2
ElasticSuite Version : 2.8.1
Environment : Production
Third party modules : None (tested in a vanilla Magento, with just ElasticSuite enabled, and Luma theme)

### Steps to reproduce
1. Set a visual swatch attribute values sort order (facet sort order) to "Result count"
2. Load a category page that display the previously configured attribute as a filter

### Expected result
1. The values are displayed in the set order, starting with the one that has the most results, ending with the one that has the least results.

### Actual result
1. The values are displayed following the admin manually set order.

YannBergonzat:
Actually there seems to be the same problem on the category filter. Don't know if it's only "visual swatches and category" or just all filters though. Some of them are working just great!

rbayet:
Hello @YannBergonzat,

I've looked into it : the elasticsearch query is OK (aggregation uses the correct sort order) and the list of items come out ok in `\Smile\ElasticsuiteCatalog\Model\Layer\Filter\Attribute` whatever the attribute type.

The issue with the visual swatches attributes is that the custom template `vendor/magento/module-swatches/view/frontend/templates/product/layered/renderer.phtml` and renderer `\Magento\Swatches\Block\LayeredNavigation\RenderLayered` take over and list options in their admin sort order.

It does not seem limited to Magento 2.3.x or even 2.2.x, I'll look into but it might have been there since the earliest Magento2/Elasticsuite versions, I'll check with @romainruaud.

Regards,

romainruaud:
@androshchuk @serg6854 can you handle this one ?

romainruaud:
I'll review PR #1550 but in the meantime @YannBergonzat can you check that PR #1550 is also fixing the issue on your environment ?

Regards

YannBergonzat:
It seems to be fixed, indeed 🙌:tada: ! Thank you!

LilyBergonzat:
Oops, my bad, I only checked the swatches!

The category filter doesn't seem to be fixed. It has the same bug, [as my comment states](https://github.com/Smile-SA/elasticsuite/issues/1503#issuecomment-520554086).

[...]

PR #1718, opened 2020-02-19
Fix swatches manual sorting

Tjitse-E:
Related PR: #1503 

I noticed that PR #1503 broke the manual sorting of the swatches. My solution would be to just fall back to the core method `getSwatchData()` if the attribute is set to manual sorting. The core method works just fine in this case.

## Preconditions
- Create a text swatch attribute, set sorting to 'manual'
- Version used 2.8.4 + Magento 2.3.4

## Swatches sorting in backend
Admin screenshot

## Before (with #1503 applied)
Before screenshot

## After (with this PR applied)
After screenshot


romainruaud:
This looks good to me and I tested it.

Thank you for contributing !

PR #1722, opened 2020-02-26
Move Catalog GraphQL to a dedicated module.

romainruaud:


wouter-toppy:
Why is this merged into `master`, and not in the `2.8.x` branch?

In our case we don't use GraphQl at all. so we havent installed all those modules, now the compile of magento will break because elasticsuite depends on it. 

please provide a tag on `2.8.x` with this changes.

kpitn:
Totally agree, can't use 2.8.5 with Magento without GraphQl

Issue #1606, opened 2019-11-03
Indexes doesn't delete


capm4:
Old indexes doesn't delete
--
### Preconditions

As we no. When full reindex start for catalogsearch_fulltext. Elastic stay to listing old index, and create new. When new index is created elastic begin to listing new and delete old. But while new index is create and some problem happened and we got exeption, indexer that was created doesn't delete. So if you have huge number with product , and have to do reindex once 1-2 week, after 3-4 month you will have many unused index.  
I thing there must be some cron job that will clean old index.

Magento Version :
2.2.6

ElasticSuite Version :
2.4.0

Environment :
Developer and Production

Third party modules :

### Steps to reproduce

1.
2.
3.

### Expected result

1.

### Actual result

1. [Screenshot, logs]

rbayet:
Hello @capm4,

I don't think it's a good idea to add a cronjob that would in essence validate inaction towards indexing or architecture problems by removing traces of the errors.
I understand that having extra unused and opened indices might put a load on the architecture and accelerate/aggravate the situation.

I would favour a "warning" cronjob that would periodically scan for indices without an alias (hence inactive) and a name indicative of a creation in the past (so as not to warn about a new index being created).
Not automatically removing those indices, it would possibly let clues in place for a sysadmin to investigate.

I'll let @romainruaud voice his opinion, but feel free to push us a PR anyway.

Regards,

romainruaud:
I do agree, but I'm not sure such a warning signal should be put inside the Magento Back-Office : people reading it might not be able to proceed proper actions.

Maybe we could add a screen into the Back-Office, something like "Indices status" listing all the "managed by ElasticSuite" indices with status : 

- used
- rebuilding
- ghost index (in case of failure during previous indexing)

And a button to allow deleting ghost indices from the Back Office.

romainruaud:
@androshchuk can you try to work on this new feature under support of @rbayet ? 

It will be a new Back-Office screen, with a proper menu entry. 

This should display a list of actually existing Elasticsearch indices.

For each index, we should be able to display : 

- index name
- alias
- number of documents
- index size
- (and if possible, to be able to expand the index mapping, this one is a nice to have).
- Index state (see below for details).
- and a Delete button but only for "Ghost" indices.

Each index can have several states : 

- **Live** if the index is actually used by the application (if it has an alias, it's supposed to be live)
[...]

PR #1727, opened 2020-02-27
Feature #1606 module indices

androshchuk:


src/module-elasticsuite-indices/Controller/Adminhtml/Index/View.php
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function execute()
+    {
+        $indexName = $this->getRequest()->getParam('name');
+        try {
+            $index = $this->esClient->getMapping($indexName);
```
romainruaud: Can you encapsulate this into a wrapper class, something like "IndexMappingProvider", that will contain the Es Client and process the call, so that we don't need to inject directly a low-level object like the client directly here ?

src/module-elasticsuite-indices/Helper/Index.php
```diff
@@ -0,0 +1,306 @@
...
+
+    /**
+     * Store number of seconds in a year.
+     */
+    private const SECONDS_IN_YEAR = 31536000;
+
+    /**
+     * @var array
+     */
+    private $elasticSuiteIndices = [
```
romainruaud: This cannot be hardcoded and should be fetched from the elasticsuite_indices.xml files.

Eg if you add this module for CMS Pages, you will have a new index "cms_page" : https://github.com/Smile-SA/magento2-module-elasticsuite-cms-search

You can fetch the indices like this : https://github.com/Smile-SA/elasticsuite/blob/2.8.x/src/module-elasticsuite-core/Index/IndexSettings.php#L166

The names will then be the keys of the array.

**=> The same logic is written elsewhere, it should be refactored.**

src/module-elasticsuite-indices/Helper/Index.php
```diff
@@ -0,0 +1,306 @@
...
+
+    /**
+     * Get a index status.
+     *
+     * @param string $indexName Index name.
+     * @param string $alias     Index alias.
+     *
+     * @return string
+     */
+    public function getIndexStatus($indexName, $alias): string
```
romainruaud: `Helper` containing many method is a bit old-school and should be refactored.

Most probably this method could be into a single-purpose object called something like "IndexStatusProvider".

src/module-elasticsuite-indices/Model/ResourceModel/Index/Collection.php
```diff
@@ -0,0 +1,97 @@
...
+    {
+        foreach ($this->indexHelper->getElasticSuiteIndices() as $indexName => $alias) {
+            $item = $this->getNewEmptyItem();
+            $data = [
+                'index_name'  => $indexName,
+                'index_alias' => $alias,
+            ];
+
+            try {
+                $indexStatsResponse = $this->esClient->indexStats($indexName);
```
romainruaud: This could be done by a wrapper object, something like IndexStatsProvider, that will embed the client and process the call, so that we do not need to inject the low-level object $esClient directly here.

src/module-elasticsuite-indices/Model/ResourceModel/Index/Collection.php
```diff
@@ -0,0 +1,97 @@
...
+            } catch (Missing404Exception $e) {
+                $data['index_status'] = IndexStatus::REBUILDING_STATUS;
+                $this->addItem($item->setData($data));
+                continue;
+            }
+
+            $indexStats = current($indexStatsResponse['indices']);
```
romainruaud: The $indexStats can be built from the $indexStatsResponse directly in the IndexStatsProvider so that it will keep it easy to use in this Collection.

src/module-elasticsuite-indices/Model/ResourceModel/Index/Collection.php
```diff
@@ -0,0 +1,97 @@
...
+            $data['number_of_documents'] = $indexStats['total']['docs']['count'];
+            $data['size'] = $this->indexHelper->sizeFormatted($indexStats['total']['store']['size_in_bytes']);
```
romainruaud: And the piece of code responsible of computing the index size can be also moved into the IndexStatsProvider class :)

src/module-elasticsuite-indices/Model/ResourceModel/IndexMapping/Collection.php
```diff
@@ -0,0 +1,106 @@
...
+     * @param bool $printQuery Is print query.
+     * @param bool $logQuery   Is log query.
+     * @return Collection
+     * @SuppressWarnings(PHPMD.BooleanArgumentFlag)
+     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
+     */
+    public function loadData($printQuery = false, $logQuery = false): Collection
+    {
+        $data = [];
+        $mapping = $this->esClient->getMapping($this->name);
```
romainruaud: Same as previously, an intermediate object `IndexMappingProvider` should be use to avoid injecting the esClient directly here.

src/module-elasticsuite-indices/etc/acl.xml
```diff
@@ -0,0 +1,28 @@
...
+ * @copyright 2020 Smile
+ * @license   Open Software License ("OSL") v. 3.0
+ */
+ -->
+<config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:Acl/etc/acl.xsd">
+    <acl>
+        <resources>
+            <resource id="Magento_Backend::admin">
+                <resource id="Smile_ElasticsuiteCore::menu">
+                    <resource id="Smile_ElasticsuiteIndices::manage" title="Indices Status"/>
```
romainruaud: Add a new ACL "Delete Index" so that we can allow some people to read, and some other to read + delete

src/module-elasticsuite-indices/i18n/fr_FR.csv
```diff
@@ -0,0 +1,19 @@
...
+"You deleted the index %1.","Vous avez supprimé l'index %1."
+"B","B"
+"KB","KB"
+"MB","MB"
+"GB","GB"
+"Index can't be removed.","L'index ne peut pas être supprimé."
+"Mapping for index:","Mappage pour l'index:"
+"Name","Nom"
+"Type","Type"
+"Mapping is empty.","La cartographie est vide."
```
romainruaud: Thanks, we'll modify some things here at the end but that's nice to provide translations :)

romainruaud:
This has been replaced by #1775 and merged into master, good job Dima !

PR #1737, opened 2020-03-05
Preventing indexing cron failure on bad json data

rbayet:
usefull if DB storage table was populated with external data.

### DIFF
--- a/src/module-elasticsuite-tracker/Model/ResourceModel/EventQueue.php
+++ b/src/module-elasticsuite-tracker/Model/ResourceModel/EventQueue.php
@@ -79,7 +79,12 @@ class EventQueue extends AbstractDb
         $eventData = $this->getConnection()->fetchAll($select);
 
         foreach ($eventData as &$currentEvent) {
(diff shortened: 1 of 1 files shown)


rbayet:
In some situation where external tracking data (for instance, from production server logs, or from a environment with our [M1 module](https://github.com/Smile-SA/smile-magento-elasticsearch)) has been injected directly into the storage DB, there is a decent risk of json deserialization failure.
This should not stop the whole indexing process.

At the moment, an "empty" event will be indexed, but this could be changed so the event is completely skipped ?

src/module-elasticsuite-tracker/Model/ResourceModel/EventQueue.php
```diff
@@ -79,7 +79,12 @@ public function getEvents($limit = null)
         $eventData = $this->getConnection()->fetchAll($select);
 
         foreach ($eventData as &$currentEvent) {
```
rbayet: Due to the foreach syntax here, it's not possible to simply "unset" and bypass the whole event in the catch.

Issue #1668, opened 2020-01-17
Make "-" searchable - backport


simonrl:
In reference to this issue: https://github.com/Smile-SA/elasticsuite/issues/1011

Any idea how I implement only the changes from #1022 without updating to the latest version?

I tried defining the analyzers with new params in my own elasticsuite_analysis.xml file in a new module, but get the error message that multiple nodes have been found for /analysis/analyzers/analyzer[@name='standard' and @language='default']/filters/filter

If overriding analyzer definitions is not possible, would you merge a backport to the 2.3.x branch if I create one, even if this version is no longer supported?

Thanks
Simon

@rbayet

rbayet:
Hello @simonrl,

Sorry I'm only coming back to this subject.
Re-reading with fresh eyes your problem :
> I tried defining the analyzers with new params in my own `elasticsuite_analysis.xml` file in a new module, but get the error message that multiple nodes have been found for /analysis/analyzers/analyzer[@name='standard' and @language='default']/filters/filter

I'm realising that what you're missing is probably just the piece of code which merges different `elasticsuite_analysis.xml` files.
Could your provide me with your custom `elasticsuite_analysis.xml` section ?
I'll try to see if it I'm correct.

Regards,

simonrl:
Hey @rbayet 

sorry for the delayed reply.

I basically just copied `src/module-elasticsuite-core/etc/elasticsuite_analysis.xml` to my custom module and re-defined the analyzers in there, hoping they would be preferred over the core-analyzers. (I also added a `<depends>` node in my `module.xml`).

```
<?xml version="1.0"?>
<analysis xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:module:Smile_ElasticsuiteCore:etc/elasticsuite_analysis.xsd">
<analyzers>
        <analyzer name="standard" tokenizer="whitespace" language="default">
            <filters>
                <filter ref="lowercase" />
                <filter ref="ascii_folding" />
                <filter ref="trim" />
                <filter ref="elision" />
                <filter ref="word_delimiter" />
                <filter ref="standard" />
            </filters>
            <char_filters>
                <char_filter ref="html_strip" />
            </char_filters>
        </analyzer>
        <!-- more analyzers I changed -->
</analyzers>
</analysis>
```

I assume this is not correct?
[...]

PR #1744, opened 2020-03-06
Resolves #1742 Per language stemmer override feature

### LINKED ISSUE #1742: Introduce a stemmer override step in the analysers
**Is your feature request related to a problem? Please describe.**
Sometimes the automatically selected language specific stemmer has some edge issues, either failing to detect singular/plural form (seen with italian) or assigning the same root to words that, although sharing an etymologic ancestry, are now quite distinct (seen in french on an old Solr project).

Using the thesaurus to fix those issues is a bad idea, because it denatures its intended usage which **is not to fix analysis issues**, but apply **business specific synonyms and expansions**.

**Describe the solution you'd like**
Introduce a basic support for language/locale dependant [stemmer_override token filter](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/analysis-stemmer-override-tokenfilter.html) in the _standard_ analyser chain.
At the moment, supporting the simpler "_rules_" parameters should be enough.

**Additional context**
Problem historically seen with old versions (Solr 3.x based project) of the _french_ and _light_french_ stemmer which would reduce both _collectivités_ (collectivities) and _collections_ to _collect_.
Which was annoying considering the user was mainly selling books and almost all of them had a "collection" attribute in the format "Collection [XYZ]" : searching for "droit des collectivités" ("Collectivities law") would return all books.

Problem also seen recently with an italian user selling among other things power tools : the _italian_ stemmer would reduce _trapano_ and _trapani_ (drill and drills) to different stem.
=> Searching for "drills" would not return fewer products than search from "drill".

rbayet:
either to fix or customise stemmer output.

### DIFF
--- a/src/module-elasticsuite-core/etc/elasticsuite_analysis.xml
+++ b/src/module-elasticsuite-core/etc/elasticsuite_analysis.xml
@@ -139,6 +139,18 @@
         <filter name="standard" type="stemmer" language="tr">
             <language>turkish</language>
         </filter>
+        <filter name="standard_override" type="stemmer_override" language="fr">
+            <!-- No known issues with the "french" or "light_french" stemmers. But this is to show that having an empty list of rules is valid.
+                 This means existing rules can be both re-defined or voided in a custom module. -->
+            <rules>[]</rules>
+        </filter>
+        <filter name="standard_override" type="stemmer_override" language="it">
+            <!--
+                Known issues with the "italian" stemmer. Please note that we do not claim 100% accuracy in the selected stems.
+                For instance, the stem for "trapani" is originally "trapan", the same as the verb "trapanare".
+                So the actual valid rule might be "trapano => trapan" ... -->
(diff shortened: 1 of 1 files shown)


src/module-elasticsuite-core/etc/elasticsuite_analysis.xml
```diff
@@ -165,6 +177,7 @@
                 <filter ref="word_delimiter" />
                 <filter ref="lowercase" />
                 <filter ref="elision" />
+                <filter ref="standard_override" />
```
rbayet: Change name to "standard_before" ? Or something that does not directly reference the fact that it will be a stemmer_override (as "standard" does not say that the token filter will be a stemmer) ?

rbayet:
**TODO** check that those filters can indeed by overridden in a custom module (see #1668)

romainruaud:
Sounds good, I agree that we have to ensure inheritance and overriding between multiple xml files is possible.

If that's the case, you can rebase and merge.

PR #1787, opened 2020-04-30
Replace Zend with Laminas

rvitaliy:
Zend is replaced with Laminas in magento 2.3.5-p1

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,8 @@
         "magento/module-inventory-sales": ">=1.0.3",
         "magento/module-inventory-indexer": ">=1.0.3",
         "magento/magento-composer-installer": "*",
-        "elasticsearch/elasticsearch": "~6.7|~7.2"
+        "elasticsearch/elasticsearch": "~6.7|~7.2",
+        "laminas/laminas-stdlib": "^3.2"
     },
     "replace": {
         "smile/module-elasticsuite-core": "self.version",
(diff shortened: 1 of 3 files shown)


rvitaliy:
see https://github.com/magento/magento2/issues/28061

romainruaud:
Hi @rvitaliy 

It's (unfortunately) not the only place where Zend is used in Elasticsuite.

However, what's unclear for me is that it does not seem to break anything even if these classes are non-existing on my local environment.

I guess they have some sort of autoload or composer hack somewhere to handle this ?

Do you have any insight ?

rvitaliy:
Now in magento 2.3.5 they finally migrate to Laminas project instead of Zend so our code is incompatible with new magento version.
see https://github.com/zendframework/zend-stdlib & https://github.com/laminas/laminas-stdlib

This PR is not fix at all this problem because if our vendor require this dependency we must specify it in our composer.
now i fix this PR to require the laminas vendor and fix this bug

romainruaud:
If that's breaking something on your environment can you give me a reproduction scenario ?

romainruaud:
Okay I found out, it's not breaking anything because Magento does include this package : https://github.com/laminas/laminas-zendframework-bridge

So that's fine for now, but we'll have to migrate to Laminas namespace to be safer.

Thanks for reporting it, I'll probably merge this PR soon

romainruaud:
Done in master branch.

Thanks

PR #1809, opened 2020-05-13
Fix #1806 relevance configuration for containers scope.

### LINKED ISSUE #1806: Something went wrong while saving this configuration: Invalid scope type 'containers'
We are using smile elasticsuite version 2.8.0 with magento version 2.3.5.
We are not able to save value on Elasticsuite > Search Relevance with any view.

Ex :
I reach at Elasticsuite > Search Relevance, select view as "Catalog Category Search", uncheck the "use default" checkbox, change value on "Minimum Should Match" and try to save and I got error like "Something went wrong while saving this configuration: Invalid scope type 'containers'"

Please check attached screenshot for same.
Screenshot from 2020-05-12 12-56-08

Can you please help to sort out the issue.

Thanks in advance,
David Andersson

romainruaud:


romainruaud:
replaced by #1810

PR #1850, opened 2020-06-15
Fixes #1846 Preview Optimizers

### LINKED ISSUE #1846: In optimizers preview in admin panel optimizers are not filtered properly
During new optimizer creation (or editing) in admin panel, if I click the "Preview" button, the wrong search result appears in both "Base Results" and "Optimized Results" columns. The product list differs from the products list on the search result page on the Frontend.
I debugged it and found this: 
\Smile\ElasticsuiteCatalogOptimizer\Model\Optimizer\Search\OptimizerFilter::getOptimizerIds()
\Smile\ElasticsuiteCatalogOptimizer\Model\Optimizer\Category\OptimizerFilter::getOptimizerIds()
There are conditions like this:
`if ($this->searchContext->getCurrentSearchQuery()) {`
and
`if ($this->searchContext->getCurrentCategory()) {`

They never return true, so getOptimizerIds() always return null. Then here 
\Smile\ElasticsuiteCatalogOptimizer\Model\Optimizer\ApplierList::applyOptimizers
at this line optimizers are never filtered:
`$functions = array_intersect_key($functions, array_flip($optimizerIds));`
Elasticsuite doesn't respect the query text and selected category during the Preview from admin panel. When I search from the Frontend, everything is fine.
For me, it looks like a bug. Could you advise, please? Maybe it's a feature I do not get ;)

### Preconditions

Magento Version : 2.3.3.

ElasticSuite Version :
I faced the issue on 2.8.4, but I've reproduced it on latest 2.8.7 as well.

Environment :
Dev and Prod.

Third party modules :
Amasty_Base
Amasty_Feed
Amasty_ProductAttachment
FireGento_FastSimpleImport
CtiDigital_Configurator
MSP_ReCaptcha
MSP_TwoFactorAuth
Magehelper_Menusearch
SchumacherFM_Mocks
Smile_DebugToolbar
Smile_ElasticsuiteAdminNotification
Smile_ElasticsuiteCore
Smile_ElasticsuiteCatalog
Smile_ElasticsuiteCatalogRule
Smile_ElasticsuiteCatalogOptimizer
Smile_ElasticsuiteAnalytics
Smile_ElasticsuiteQuickOrder
Smile_ElasticsuiteSwatches
Smile_ElasticsuiteThesaurus
Smile_ElasticsuiteVirtualCategory
Tigren_Ajaxsuite
Tigren_Ajaxcompare

### Steps to reproduce

1. Create 'Catalog Product Search' optimizer with Apply to = 'Selected search terms' and select some search terms.
2. Start creating another new one optimizer (do not test on optimizer from p.1, because current optimizer is not applied in "Base Results" column)
3. Try to search for some query that is not selected as an optimizer search term in p.1.
4. Better to debug it, because it really depends on products how they will be ordered

### Expected result

1. Optimizer from p.1 should not be applied, because query text is not in the list of 'Apply to' search terms. Search result in optimizer preview is the same as on the search result on the Frontend.

### Actual result

1. Optimizer from p1 is applied. 
2. Search result in optimizer preview differs from the search result on the Frontend.

P.S. The same with a category. If you have some optimizer A with selected categories and then in another optimizer B search for some another category in Preview, optimizer A is applied anyway.

androshchuk:


romainruaud:
replaced by #1908 and #1909

PR #1882, opened 2020-07-21
Added 'jquery/ui' for correct work of price slider

androshchuk:


src/module-elasticsuite-catalog/view/frontend/web/js/range-slider-widget.js
```diff
@@ -15,7 +15,8 @@
 /*jshint browser:true jquery:true*/
 /*global alert*/
 
-define(['jquery', 'Magento_Catalog/js/price-utils', 'mage/template', 'Magento_Ui/js/modal/modal'], function ($, priceUtil, mageTemplate) {
+define(['jquery', 'Magento_Catalog/js/price-utils', 'mage/template', 'jquery/ui', 'Magento_Ui/js/modal/modal'],
```
Quazz, line 18: Hey, it's bad practice to load the entire jquery/ui library. 

I'm guessing the slider isn't working due to this PR https://github.com/magento/magento2/pull/27616

If so, use jquery-ui-modules/slider instead to keep the javascript loading to a minimum.

romainruaud: good catch, can you fill an issue with this ? so that we don't forget to fix it (of course, if you can also do a PR that would be awesome).

regards

PR #1884, opened 2020-07-22
Fix category filter

subbu1993:


romainruaud:
I close this, it's too old and does not give any details about what it does.

PR #1914, opened 2020-08-12
Fix - page size and pagination issue.

Bashev:
Fix issue which return all products in category page and autocomplete results and results page.


amenk:
Thanks @Bashev Are you able to add tests for this, because that constantly breaks :) Seems to be related to #1862

groomershop-mt:
We can confirm that this problem occurs in 2.9.x branch at Magento 2.3.5-p2 and after applying this changes the problem is being solved.

choleaoum:
This fixes the issue with autocomplete returning all products everytime and category pagination. Although, there is an issue when executing an actual search query. It shows all products for that search query on a single page along with a selection of page numbers. Each page shows the exact same products.

romainruaud:
I'll have a look at this one asap, thank you for submitting it.

romainruaud:
Merged in 2.8.x, 2.9.x and 2.10.x branch.

To be tagged and released soon.

Many many many thanks @Bashev :)

dawidudziela:
> choleaoum

Did you fixed your issue? I see the same problem.

amenk:
@dawidudziela I think it is better, if you create a separate question or issue with more details.

PR #1927, opened 2020-08-24
Fixes #1926 include jquery slider submodule instead of whole jquery-ui

### LINKED ISSUE #1926: Include only jquery slider submodule instead of whole jquery/ui
### Preconditions

Magento Version : 2.4.x

ElasticSuite Version : 2.10.x

Environment : n/a

Third party modules : n/a

As per @Quazz comment on #1882 we should not include directly `jquery/ui` for the slider to work properly : https://github.com/Smile-SA/elasticsuite/pull/1882#discussion_r466296453

vahonc:


jasperzeinstra:
@androshchuk this is indeed the fix, why is it not merged and tagged yet? :+1:

PR #1928, opened 2020-08-26
improve composer configuration

rvitaliy:
- added allowed php versions (^7.3 | ^7.4) (see "magento/framework": ">=103.0.0 requirements)
- added ext-json and ext-mbstring as required extensions
- remove useless replaces

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,6 +35,7 @@
         }
     ],
     "require": {
+        "php": "^7.3 | ^7.4",
         "magento/framework": ">=103.0.0",
         "magento/module-store": ">=101.1.0",
         "magento/module-backend": ">=102.0.0",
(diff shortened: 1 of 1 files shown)


romainruaud:
Nope we will not add PHP versions requirements, we are a Magento extension, so Magento himself is enough to handle PHP versions.

Not sure about the rest for now. Maybe ext-json and ext-mbstring should be specified anyway.

rvitaliy:
The best practice for a vendor is to specify self requirements for PHP version and extentions.

It's allow an external contributor instantly understand what syntax and extensions he can use for this project.
Another advantage is that the modern IDE like PhpStorm analyze composer.json to auto-configure the inspection of code.

I think both things should be approved.

romainruaud:
We do not have self requirements for PHP versions : we are forced to adapt to the versions required by magento/framework.

PR #1963, opened 2020-10-07
Pass excluded categories by reference

fredden:
We have a website where loading a category page in the admin is so slow it's practically unusable. Upon investigation it was discovered that this code was taking the vast majority of time. This change reduced page load time from ~9 minutes to ~15 seconds.

`Smile\ElasticsuiteVirtualCategory\Model\Rule::getCategorySearchQuery()` takes an array of category IDs that have already been processed, and therefore should be excluded next time. It adds any categories that have now been processed to this array to avoid infinite loops/recursion. However, because this parameter was not being passed by reference the additions were being thrown away and the same categories were being processed again the next time this is called.


fredden:
Hi @rbayet. What can I do to help get this merged in?

PR #1965, opened 2020-10-10
Fix graphql pagination

bheptinh:
GraphQL pagination not working (Magento 2.4.0)

Based on [query request](https://github.com/Smile-SA/elasticsuite/blob/2.10.x/Resources/tests/graphql/category/query.gql) I've got an empty items array if `currentPage` is more than 1. So I updated Smile\ElasticsuiteCatalogGraphQl\Model\Resolver\Products\Query\Search class to be consistent with Magento\CatalogGraphQl\Model\Resolver\Products\Query\Search

Looks like the problem is [because of limitations of sort and pagination on search API](https://github.com/magento/magento2/blob/5daadaee54d523c9f6bb3939bcb2df70539d8795/app/code/Magento/CatalogGraphQl/Model/Resolver/Products/Query/Search.php#L99)

### DIFF
--- a/src/module-elasticsuite-catalog-graph-ql/Model/Resolver/Products/Query/Search.php
+++ b/src/module-elasticsuite-catalog-graph-ql/Model/Resolver/Products/Query/Search.php
@@ -67,9 +67,9 @@ class Search implements ProductQueryInterface
     /**
      * @param SearchInterface       $search                Search Engine
      * @param SearchResultFactory   $searchResultFactory   Search Results Factory
-     * @param PageSizeProvider      $pageSize
+     * @param PageSizeProvider      $pageSize              PageSize Provider
      * @param FieldSelection        $fieldSelection        Field Selection
-     * @param ProductSearch         $productsProvider       Product Provider
+     * @param ProductSearch         $productsProvider      Product Provider
      * @param SearchCriteriaBuilder $searchCriteriaBuilder Search Criteria Builder
      */
     public function __construct(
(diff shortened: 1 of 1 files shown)


mercs600:
@bheptinh We're facing with the same issue so we'll test your patch on our instance.

romainruaud:
@bheptinh please fill an issue with a detailed scenario to reproduce what is occuring.

That being said, we will not merge this "as is" because we definitely do not want to introduce the legacy magento behavior that is doing a query to all the ids : it's not acceptable in term of performances.

I let @rbayet give you more details

rbayet:
@bheptinh as Romain, please open an issue and describe the exact problem that you are trying to fix by that pull request.

I am not aware of an issue on 2.4.1 (at least in terms of response to basic pagination requests), so it could be 2.4.0 specific.

Regards,

bheptinh:
@rbayet, @romainruaud sorry for long response, 
yeah, it looks like 2.4.0 specific problem, don't see it now on 2.4.1. 

Probably I'll just close this MR. Thank you.

jahvi:
I got this same problem on 2.4.2, this PR fixed it for me. Just to give more information:

1. Fresh 2.4.2 with sample data.
2. Run GraphQL query:
```graphql
query {
  products(search: "yoga", currentPage: 1, pageSize: 5) {
    items {
      name
    }
    page_info {
      current_page
      page_size
      total_pages
    }
    total_count
  }
}
```

**Expected result**

```graphql
{
  "data": {
    "products": {
      "items": [
        ...
      ],
      "page_info": {
        "current_page": 1,
        "page_size": 5,
        "total_pages": 10
      },
      "total_count": 50
    }
  }
}
```

**Actual result**

```graphql
{
  "data": {
    "products": {
      "items": [
        ...
      ],
      "page_info": {
        "current_page": 1,
        "page_size": 1, // <-- This is always 1
        "total_pages": 10
      },
      "total_count": 50
    }
  }
}
```

bmxmale:
Same here as @jahvi, need to re-open this.

Issue #1693, opened 2020-02-05
Category pagination with page selected and 'show all' shows empty result


lcampanella-wh:
### Preconditions
PHP 7.2
Magento 2.3.3 Enterprise (developer mode)
Magento patch applied: Catalog pagination issue on Elasticsearch 6.x patch for Magento 2.3.3 
ElasticSuite Version: 2.8.4
Elasticsearch 6.5.4

Third party modules : none

### Steps to reproduce

1. Add some products to a category to use pagination in the frontend
2. Refresh indexes
3. Access category in the frontend
4. Change to any page (3 for instance) on product listing
5. Click in "All" in dropdown to show all products
6. Url should look like: <base_url>/category.html?p=3&product_list_limit=all

### Expected result

All products for given category should be shown

### Actual result

No products are found, screen with this text is shown:
We can't find products matching the selection.

This was tested on a newly created Magento instance, no other 3rd party code is present besides ElasticSuite.

Thanks.

rbayet:
Hello @lcampanella-wh, 

Concerning
> Magento patch applied: Catalog pagination issue on Elasticsearch 6.x patch for Magento 2.3.3

Are your referring to 

- https://support.magento.com/hc/en-us/articles/360035142371-Catalog-pagination-doesn-t-work-when-Elasticsearch-6-x-is-used 
- https://github.com/magento/magento2/issues/25378

Have you installed this patch before moving to Elasticsuite ?
Or had you already a pagination issue with Elasticsuite and decided to apply this patch ?

Regards,

tina-trenkner:
Hi there, 

Bug is evident on the ElasticSuite demo store. I can replicate @lcampanella-wh 's exact steps to replicate the issue on the demo store. 

PRECONDITIONS:
Whatever is used on the demo store. I have also seen this on builds with M2 Open Source 2.3.2, 2.3.4; ElasticSuite ~2.8.0; ^2.8

STEPS TO REPRODUCE:
1. Visit http://demo.magento-elastic-suite.io/index.php/women/tops-women.html
2. Click on page 2 on Women's Tops category page
3. On the product qty to show on page dropdown, choose "All"
4. URL appears as http://demo.magento-elastic-suite.io/index.php/women/tops-women.html?p=2&product_list_limit=all

EXPECTED RESULT:
User should be able to see all products in the category

ACTUAL RESULT: 
No products are found, screen with this text is shown:
We can't find products matching the selection.
[...]

PR #1994, opened 2020-11-16
Fixes #1986 #1693 Handling out of bonds page.

### LINKED ISSUE #1986: No results found when visiting the last page and changing 'show how many'
### Preconditions

Clean install with Elasticsearch 6.8.11

Magento Version : 2.3.5-p1

ElasticSuite Version : 2.9.3

Environment : Production

Third party modules : Nope, basically clean install.

### Steps to reproduce

1. Open Magento advanced search and search for products that match a product attribute
2. Go to the last page of results
3. Click back any page
4. Up the "how many results" from 10 to 25 for example

### Expected result

1. See the results of the search

### Actual result

image

rbayet:
Magento 2.3.x toolbar does not reset the current page var or re-compute it accurately when changing page size or using 'Show all'.


rbayet:
Will be merged in 2.9.x in #1993

PR #2006, opened 2020-11-25
Fixes #1931 trying to access array offset on value of type int

### LINKED ISSUE #1931: Notice: Trying to access array offset on value of type int - Layer Navigation - Swatches
### Preconditions

PHP7.4 & Mysql 8

Magento Version : 2.4.0

ElasticSuite Version: 2.10.0

Environment: Production mode

### Steps to reproduce

1. Configure a product attribute of type "Visual Swatch" 
2. Add labels, one label should start with a space
3. Make the attribute filterable and set it to show up in layer navigation
4. Create products with the values setup on the created attribute
5. Navigate to Catalog Product Listing page 

### Expected result
1.The layer navigation should be shown and work properly

### Actual result

1. In production mode the layer navigation block is not rendered at all
2. In production mode the following row is written in logs: 

==> var/log/system.log <==
[2020-08-27 10:19:01] main.CRITICAL: Notice: Trying to access array offset on value of type int in /var/www/html/releases/62/vendor/smile/elasticsuite/src/module-elasticsuite-swatches/Block/Navigation/Renderer/Swatches/RenderLayered.php on line 70 [] []

3. In developer mode, the execution is stoped with the following error trace
image

vahonc:


src/module-elasticsuite-swatches/Block/Navigation/Renderer/Swatches/RenderLayered.php
```diff
@@ -67,7 +67,9 @@ public function getSwatchData(): array
         }
 
         foreach (array_merge(array_flip($sortingArr), $attributeOptions) as $item) {
-            $attributeOptionsSort[$item['id']] = $item;
+            if (isset($item) && is_array($item)) {
```
rbayet, line 70: Could you remove the `isset($item)` test ? It's not an accurate test to know if a key exists in an array and besides, we know that $item exists, since it is the byproduct of the loop (ie the current item).
I would go for `is_array($item) && array_key_exists('id', $item)`.

The original issue due to the fact that the label is " XXX" and the indexed label is "XXX" should be taken care of, but considering we have another issue in the area (#2029), we'll maybe look at it then.

rbayet (changes requested):

rbayet (approved):

PR #2007, opened 2020-11-27
Fixes #1937 preventing displaying all products from catalog on deleted virtual category

### LINKED ISSUE #1937: When a virtual category root is deleted, all products are shown
This is just for the record, don't really expect a fix, but want to have it documented in case others stumble upon it.

### Preconditions

* Create a category A and put no products into it
* Create a subcategory "B", parent A, and put some products into it
* Create a virtual category "V" below A with a the category "B" as root
* Delete the referenced category "B"

image

In the backend it shows "entity with ID doesn't exist"

In the virtual category (another their parents, because of is_anchor) now unrelated products appear

* smile/elasticsuite 2.8.6 
* magento/product-community-edition                      2.3.4-p2   

## Expected Result

The virtual category V is empty, because it links to an invalid category

## Current Result

The virtual category V and also the parent categories seem to include all products from the catalog.

vahonc:


src/module-elasticsuite-virtual-category/Model/Rule.php
```diff
@@ -139,15 +139,21 @@ public function __toString(): string
     public function getCategorySearchQuery($category, $excludedCategories = []): ?QueryInterface
     {
         $query         = null;
+        $rootCategory  = null;
```
rbayet, line 142: I don't a reason to initialize that variable since it is systematically overwritten in your additional condition.

src/module-elasticsuite-virtual-category/Model/Rule.php
```diff
@@ -139,15 +139,21 @@ public function __toString(): string
...
             $excludedCategories[] = $category->getId();
-
-            if ((bool) $category->getIsVirtualCategory() && $category->getIsActive()) {
+            if ($category->getIsVirtualCategory() && $category->getVirtualCategoryRoot() !== null
+                && !empty($category->getVirtualCategoryRoot())) {
+                $rootCategoryId = $category->getVirtualCategoryRoot();
+                $rootCategory = $this->categoryFactory->create()->setStoreId($this->getStoreId())->load($rootCategoryId);
+                $catId = $rootCategory->getId();
+            }
+            if ((bool) $category->getIsVirtualCategory() && $category->getIsActive() && isset($catId)) {
```
rbayet, line 156: Not a fan of the approach since it integrates part of the logic of checking the existence of the root category in \Smile\ElasticsuiteVirtualCategory\Model\Rule::getVirtualRootCategory.
I would change the logic of \Smile\ElasticsuiteVirtualCategory\Model\Rule::getVirtualCategoryQuery so that it returns a null query if \Smile\ElasticsuiteVirtualCategory\Model\Rule::getVirtualRootCategory returns null and change this section of code in \Smile\ElasticsuiteVirtualCategory\Model\Rule::getCategorySearchQuery 
with something like this : 
```
            if ($category->getIsActive()) {
                if ((bool) $category->getIsVirtualCategory()) {
                  $query = $this->getVirtualCategoryQuery($category, $excludedCategories, $category->getData('virtual_category_root'));
                }
                if (!$query) {
                  if ($category->getId() {
                  }
                }
            }
                $query = $this->addChildrenQueries($query, $category, $excludedCategories);
            }
```
But it's not perfect, and still a bit more complex.

On an additional note, your fix will apply when visiting the virtual category, but won't have any effect on parent categories, so we might want to address the problem as a whole by looking at what would happen in \Smile\ElasticsuiteVirtualCategory\Model\Rule::addChildrenQueries

rbayet (changes requested):

src/module-elasticsuite-virtual-category/Model/Rule.php
rbayet: Conclusion: we need to discuss this.

PR #2017, opened 2020-12-08
REFACTOR: `around` plugins should be avoided at any price

lbajsarowicz:
[Around plugins should be avoided](https://devdocs.magento.com/guides/v2.4/coding-standards/technical-guidelines.html#4-interception) so this PR removes redundant one.


romainruaud:
Merged as part of #2341

Issue #2011, opened 2020-12-02
Crashing server if ES not responding.


alin1918:
### Preconditions

Set in the database 2-3 es servers:
smile_elasticsuite_core_base_settings/es_client/servers --- elasticsearch:9400,elasticsearch:9200,elasticsearch:9500

On first have have a non responding ES server (but active). Can be reproduced by starting a netcat listen on that server:  nc -l 9400 
Try to load a page that is using the elasticsearch.
It will remain in loading until you go and kill the process : "nc -l 9400".

A fix can be done by setting a timeout on the connection. ex vendor/elasticsearch/elasticsearch/src/Elasticsearch/Endpoints/AbstractEndpoint.php : extractOptions
![Screen Shot 2020-12-02 at 10 08 31 PM](https://user-images.githubusercontent.com/792685/100925952-5f8d8080-34eb-11eb-8e38-9ec0564284da.png)

![Screen Shot 2020-12-02 at 10 09 43 PM](https://user-images.githubusercontent.com/792685/100925750-19d0b800-34eb-11eb-93c9-4271516cf07c.png)
or 
vendor/elasticsearch/elasticsearch/src/Elasticsearch/Transport.php : performRequest 
or 
vendor/smile/elasticsuite/src/module-elasticsuite-core/Client/Client.php:search
vendor/smile/elasticsuite/src/module-elasticsuite-core/Client/Client.php:info
vendor/smile/elasticsuite/src/module-elasticsuite-core/Client/Client.php: indexExists ( A fix can be done just with this one, because will mark the two problematic servers as dead and will get a connection on the third one)

![Screen Shot 2020-12-02 at 10 20 25 PM](https://user-images.githubusercontent.com/792685/100926828-b051a900-34ec-11eb-9370-4b586cd74a42.png)

![Screen Shot 2020-12-02 at 10 20 18 PM](https://user-images.githubusercontent.com/792685/100926816-adef4f00-34ec-11eb-8005-9685177d6f42.png)

Magento Version :
Magento EE 2.3.5-p2

ElasticSuite Version :
"smile/elasticsuite": "^2.8"

Environment :
Developer

Third party modules :
A lot.

### Expected result

1. Page should Load

### Actual result

1. [Screenshot, logs]
Page is not loading
![Screen Shot 2020-12-02 at 10 10 53 PM](https://user-images.githubusercontent.com/792685/100925888-4b498380-34eb-11eb-8e05-5308fd89515c.png)

[...]

PR #2018, opened 2020-12-09
Issue-2011 - Set configuration for readtimeout and set timeout on ES …

alin1918:
…client https://github.com/Smile-SA/elasticsuite/issues/2011

### DIFF
--- a/src/module-elasticsuite-core/Client/ClientBuilder.php
+++ b/src/module-elasticsuite-core/Client/ClientBuilder.php
@@ -109,6 +109,14 @@ class ClientBuilder
             $clientBuilder->setSelector($selector);
         }
 
+        $connectionParams = ['client' => []];
+        if (isset($options['timeout'])) {
+            $connectionParams['client']['timeout'] = $options['timeout'];
+        }
+        if (count($connectionParams['client'])) {
+            $clientBuilder->setConnectionParams($connectionParams);
+        }
+
         return $clientBuilder->build();
     }
 
--- a/src/module-elasticsuite-core/Client/ClientConfiguration.php
+++ b/src/module-elasticsuite-core/Client/ClientConfiguration.php
@@ -113,6 +113,14 @@ class ClientConfiguration implements ClientConfigurationInterface
     }
 
     /**
+     * {@inheritdoc}
+     */
+    public function getReadTimeout()
+    {
+        return (int) $this->getElasticsearchClientConfigParam('timeout');
+    }
+
+    /**
      * {@inheritDoc}
      */
     public function getOptions()
@@ -125,6 +133,7 @@ class ClientConfiguration implements ClientConfigurationInterface
             'http_auth_pwd'         => $this->getHttpAuthPassword(),
             'is_debug_mode_enabled' => $this->isDebugModeEnabled(),
             'max_parallel_handles'  => $this->getMaxParallelHandles(),
+            'timeout'               => $this->getReadTimeout(),
         ];
 
         return $options;
--- a/src/module-elasticsuite-core/etc/adminhtml/system.xml
+++ b/src/module-elasticsuite-core/etc/adminhtml/system.xml
@@ -66,6 +66,11 @@
                     <comment>In seconds.</comment>
                     <frontend_class>validate-number</frontend_class>
                 </field>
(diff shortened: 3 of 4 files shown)


romainruaud:
Sounds great !

src/module-elasticsuite-core/Client/ClientConfiguration.php
```diff
@@ -112,6 +112,14 @@ public function getMaxParallelHandles()
         return (int) $this->getElasticsearchClientConfigParam('max_parallel_handles');
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function getReadTimeout()
+    {
+        return (int) $this->getElasticsearchClientConfigParam('timeout');
```
rbayet, line 120: Change param name to 'read_timeout' here if we change it in the system.xml ?

src/module-elasticsuite-core/etc/adminhtml/system.xml
```diff
@@ -66,6 +66,11 @@
                     <comment>In seconds.</comment>
                     <frontend_class>validate-number</frontend_class>
                 </field>
+                <field id="timeout" translate="label comment" type="text" sortOrder="57" showInDefault="1" showInWebsite="0" showInStore="0">
```
rbayet, line 69: Should we change the param name to 'read_timeout' here to understand to clearly differentiate with the connection one ?

src/module-elasticsuite-core/Client/ClientBuilder.php
```diff
@@ -109,6 +109,17 @@ public function build($options = [])
             $clientBuilder->setSelector($selector);
         }
 
+        $connectionParams = ['client' => []];
+        if (isset($options['timeout'])) {
+            $connectionParams['client']['timeout'] = $options['timeout'];
```
rbayet, line 114: Change $options['timeout'] to $options['read_timeout'] here also to be consistent ?

rbayet:
Hello @alin1918,

Thanks for the PR !
What do you think of the proposed change to use 'read_timeout' as far as possible from the name of the system.xml variable up to the point it is passed as a connection param ?
Also, if we have a new system.xml variable, please add a default value in config.xml.

Feel free also to squash your commits and use the prefix "Fixes #2011 " in your commit message.

Regards,

alin1918:
> Hello @alin1918,
> 
> Thanks for the PR !
> What do you think of the proposed change to use 'read_timeout' as far as possible from the name of the system.xml variable up to the point it is passed as a connection param ?
> Also, if we have a new system.xml variable, please add a default value in config.xml.
> 
> Feel free also to squash your commits and use the prefix "Fixes #2011 " in your commit message.
> 
> Regards,

Hi, the reason why I let it as "timeout" instead of "read_timeout", is because is already defined in etc/config.xml as "timeout" (but not used), the fact that the "timeout" name is being use in the library that creates the curl object, and also that the curl option is called being defined is CURLOPT_TIMEOUT. In etc/adminhtml/system.xml the label states that is "Read Timeout". 
I have no problem changing to read_timeout, just wanted to point out my reasoning in case you missed any of those. 

Just let me know and I can change it. Thanks.

rbayet:
Hello @alin1918,

Thanks for your explanations. Indeed you used the correct naming conventions.
Could you simply squash your two commits with a commit message along the way of "Fixes #2011 ES client support for connection and read timeout" ?

Regards,
Richard

alinalexandru:
@rbayet 
There is one thing that might be done here.
To spit that timeout into 2 fields (frontend and admin). During reindex you might need to have a higher value for read timeout. (because it takes longer to process), and when it's used in frontend to mark the server as down much faster.

For example for frontend use: 2-3 seconds should be ok.
For admin usage (reindex): 30-60 seconds.

We are using this patch in production, already

rbayet:
@alin1918,

I was thinking about that test case of reindexing on a huge catalog and also for some particular operations.
If that cannot be nicely done, maybe we'll go with the connection_timeout (can't seem to find where the support for it disappeared) for now and rethink about the read_timeout

Regards,

alin1918:
@rbayet I will do some tests to see how will this work with a big reindexing and see if there is an easy fix for that.

PR #2019, opened 2020-12-09
Bugfix: Magento does not support GTE, it should be GTEQ

lbajsarowicz:
`GTE` condition in filters does not work correctly with Magento collections, although it works fine with raw Elasticsearch.
Magento collections should be filtered by `GTEQ`.


torhoehn:
It's also mentioned correctly here: https://github.com/Smile-SA/elasticsuite/wiki/Querying#filters

rbayet:
Hello @lbajsarowicz, @torhoehn,

Considering that filter will be applied at the Elasticsearch level and not at the Magento/DB and that indeed, the Magento/GTEQ operator will be transformed as an Elasticsearch/GTE, I'm a bit puzzled by the "Bugfix" prefix in the PR title/commit message.

I would say it is more the case of a documentation consistency or code esthetics enhancement, wouldn't you ?

Please provide more context by opening an issue if I'm missing something.

Regards,

PR #2030, opened 2021-01-05
Added check on fetch order from checkout session

mirkocesaro:
This pull request contains a check to avoid errors like:
Error: Call to a member function getMethod() on null in /var/www/html/vendor/smile/elasticsuite/src/module-elasticsuite-tracker/Block/Variables/Page/Order.php:74

This could happen when the object returned from the checkout session by [getLastRealOrder](https://github.com/magento/magento2/blob/2.4.1/app/code/Magento/Checkout/Model/Session.php#L531) method is an empty Order class.

### DIFF
--- a/src/module-elasticsuite-tracker/Block/Variables/Page/Order.php
+++ b/src/module-elasticsuite-tracker/Block/Variables/Page/Order.php
@@ -65,7 +65,7 @@ class Order extends \Smile\ElasticsuiteTracker\Block\Variables\Page\AbstractBloc
 
         $order = $order = $this->checkoutSession->getLastRealOrder();
 
-        if ($order) {
+        if ($order && $order->getIncrementId()) {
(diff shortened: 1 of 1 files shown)


rbayet:
Hello @mirkocesaro,

Just out of curiosity, do you have a functional example of such a situation ?

Regards,

mirkocesaro:
Hi @rbayet ,
unfortunately with the sample data and luma theme is not reproducible. 

But the code of getLastRealOrder is clear: if the orderId is not set it returns an instance of order from the factory. 
And the getPayment() of the smile class returns null raising the error mentioned in my previous comment.

Regards,
Mirko

mirkocesaro:
Hi @rbayet , 
what do you think? Is it required to have a way to reproduce it? 

Thanks,
Mirko

PR #2108, opened 2021-04-13
Fix redirect if one result config scope

mbijnsdorp:
Config was always taken from the default scope, while the option can be configured on store scope.


mbijnsdorp:
@romainruaud There appears to be an issue with the integration checks (php 7.3 not found)

Issue #2103, opened 2021-04-06
Field name mapping updateable from constructor arguments for product fulltext collection


sky-hub:
**Is your feature request related to a problem? Please describe.**
Sometimes you need to add custom filters with custom behavior which may require adding a new field mapping in the `\Smile\ElasticsuiteCatalog\Model\ResourceModel\Product\Fulltext\Collection::$fieldNameMapping` array.
Currently there is no way in adding additional fields unless:
- rewriting collection 
- reimplement method \Smile\ElasticsuiteCatalog\Model\Layer\Filter\DecimalFilterTrait::apply at filter level to pass `$this->getFilterField()` instead of `$this->getAttributeModel()->getAttributeCode()`
- intercepting \Smile\ElasticsuiteCatalog\Model\ResourceModel\Product\Fulltext\Collection::addFieldToFilter with a before plugin and change $field value

**Describe the solution you'd like**
Add a new argument to `\Smile\ElasticsuiteCatalog\Model\ResourceModel\Product\Fulltext\Collection::__construct` named `$fieldNameMapping` of type array, which will be merged with class property `$fieldNameMapping`. This way new mappings can be configured from di.xml

**Describe alternatives you've considered**
A new class injected into `\Smile\ElasticsuiteCatalog\Model\ResourceModel\Product\Fulltext\Collection::__construct` with the sole purpose of handling field mappings?

**Additional context**
N/A

romainruaud:
I agree on this, this has been on our todo from so long. But we ended up duplicating this `$fieldNameMapping` at many other locations (API, GraphQL, attributes rules, etc...).

So I'd be happy to review a PR that could fix this, with an unique object that could be also overriden/enriched through DI.

regards

sky-hub:
@romainruaud , I will think of a solution in the following days and post it here so we can discuss it, if that is ok with you.
I guess it would be nice to offer some backwards compatibility also.

romainruaud:
Yes @sky-hub feel free to propose a PR associated to this.

sky-hub:
@romainruaud, this is next on my TODO list. I hope to have a PR soon.

sky-hub:
@romainruaud, here's a draft PR: https://github.com/Smile-SA/elasticsuite/pull/2121

The PR only covers field mapping for search request builder, layer navigation filters, product fulltext collection and catalog rule attributes.

In regards of layer navigation filters I did not implement field mappings for 
[...]

PR #2121, opened 2021-04-21
Allow search request field mapping to be configured from di.xml

sky-hub:


romainruaud:
Looks promising, even if I did not test it deeply actually.

What remains to be done ?

sky-hub:
@romainruaud , please see my last comment in here https://github.com/Smile-SA/elasticsuite/issues/2103. We will see after your reply what else needs to be done here.

romainruaud:
This looks clean for me, can you rebase it against a recent 2.10.x and also fix the Code quality issues ?

I'll try to test it by myself then.

Regards

sky-hub:
@romainruaud , Code Quality Issues have been resolved and also rebased to master branch.

src/module-elasticsuite-virtual-category/Model/Layer/Filter/Category.php
```diff
@@ -56,6 +64,7 @@ public function __construct(
         \Magento\Framework\App\Config\ScopeConfigInterface $scopeConfig,
         \Smile\ElasticsuiteCore\Api\Search\ContextInterface $context,
         \Smile\ElasticsuiteVirtualCategory\Model\Category\Filter\Provider $filterProvider,
+        RequestFieldMapper $requestFieldMapper,
```
romainruaud, line 67: This should also be injected into the parent::__construct() call just below, shouldn't it ?

src/module-elasticsuite-catalog/etc/di.xml
```diff
@@ -330,4 +330,41 @@
...
+
+    <type name="Smile\ElasticsuiteCatalogRule\Model\Rule\Condition\Product\AttributeList">
+        <arguments>
+            <argument name="requestFieldMapper" xsi:type="object">Smile\ElasticsuiteCatalog\Model\Layer\Filter\Field\Mapper</argument>
+        </arguments>
+    </type>
+    <virtualType name="Smile\ElasticsuiteCatalog\Model\Layer\Filter\Field\Mapper" type="Smile\ElasticsuiteCore\Model\Search\Request\Field\Mapper">
+        <arguments>
+            <argument name="fieldNameMapping" xsi:type="array">
+                <item name="price" xsi:type="string">price.price</item>
```
romainruaud, line 352: I'm not sure the Mapper should vary as a virtual type. A mapping is something "consistant", it can be a bit blurry that in some contexts the mapping for "price" could  be different.

And about "categories" mapping, I'm not sure it should "appear" only in particular contexts.

src/module-elasticsuite-catalog/etc/di.xml
```diff
@@ -330,4 +330,41 @@
...
+
+    <type name="Smile\ElasticsuiteCatalog\Plugin\Search\RequestMapperPlugin">
+        <arguments>
+            <argument name="requestFieldMapper" xsi:type="object">Smile\ElasticsuiteCatalog\Search\Request\Field\Mapper</argument>
+        </arguments>
+    </type>
+    <virtualType name="Smile\ElasticsuiteCatalog\Search\Request\Field\Mapper" type="Smile\ElasticsuiteCore\Model\Search\Request\Field\Mapper">
+        <arguments>
+            <argument name="fieldNameMapping" xsi:type="array">
+                <item name="category_id" xsi:type="string">category.category_id</item>
```
romainruaud, line 366: Same as above, if "category_id" can exist in the mapping, this should probably exist globally in the object.

src/module-elasticsuite-catalog/etc/di.xml
```diff
@@ -330,4 +330,41 @@
...
+    <type name="Smile\ElasticsuiteCatalogRule\Model\Rule\Condition\Product\AttributeList">
+        <arguments>
+            <argument name="requestFieldMapper" xsi:type="object">Smile\ElasticsuiteCatalog\Model\Layer\Filter\Field\Mapper</argument>
+        </arguments>
+    </type>
+    <virtualType name="Smile\ElasticsuiteCatalog\Model\Layer\Filter\Field\Mapper" type="Smile\ElasticsuiteCore\Model\Search\Request\Field\Mapper">
+        <arguments>
+            <argument name="fieldNameMapping" xsi:type="array">
+                <item name="price" xsi:type="string">price.price</item>
+                <item name="categories" xsi:type="string">category.category_id</item>
```
romainruaud, line 353: Maybe the special case about "categories" should be injected by the virtual-category module, since this field naming is introduced by `Smile\ElasticsuiteVirtualCategory\Model\Layer\Filter\Category`

src/module-elasticsuite-catalog/Model/Layer/Filter/Category.php
```diff
@@ -231,7 +241,7 @@ protected function getChildrenCategories()
      */
     protected function getFilterField()
     {
-        return 'category.category_id';
+        return $this->requestFieldMapper->getMappedFieldName('categories');
```
romainruaud, line 244: Why not keeping "category_id" here ? This could ensure we have a difference between legacy category filter and the virtual one.

romainruaud (changes requested):

src/module-elasticsuite-virtual-category/Model/Layer/Filter/Category.php
sky-hub: Agreed

src/module-elasticsuite-catalog/Model/Layer/Filter/Category.php
sky-hub: No need to keep "category_id". 
To keep consistency the actual value that should be is the attribute code which stores product categories, which is "category_ids'.

Changed to: `$this->requestFieldMapper->getMappedFieldName('category_ids');`

src/module-elasticsuite-catalog/etc/di.xml
sky-hub: Actually the "categories" mapping is not needed at all. Further more the method `\Smile\ElasticsuiteCatalog\Model\Layer\Filter\Category::getFilterField` is not used and can not be used in neither category legacy filter or virtual category filter. 

The reason it can not be used is because
`\Smile\ElasticsuiteCatalog\Search\Request\Product\Attribute\Aggregation\Category::getAggregationData`
and
`\Smile\ElasticsuiteVirtualCategory\Search\Request\Product\Attribute\Aggregation\Category::getAggregationData`

are using custom bucket naming for categories bucket which is not consistent with the other filters, for example:
`\Smile\ElasticsuiteCatalog\Search\Request\Product\Attribute\Aggregation\Price::getAggregationData`

src/module-elasticsuite-catalog/etc/di.xml
sky-hub: I will need a bit more time to do some more testing on this one. It may be the case that this one can be removed but I need to do some testing with virtual rules to see what happens.

sky-hub: And ... it seems it was added here: https://github.com/Smile-SA/elasticsuite/commit/e5885bb068447e89a6c6acd971931d5d94d9d686#diff-8d2a359b86ebe589e64eb2d7579c6eea4c5679253d3b1ef130eb7e41bec8fbf2

Can you comment on this one?

src/module-elasticsuite-catalog/etc/di.xml
sky-hub: Virtual type has been removed.

src/module-elasticsuite-catalog/etc/di.xml
sky-hub: From what I understand the "category_id" mapping was added since GraphQl, and considering the latest changes of the module I think we can add this one `module-elasticsuite-catalog-graph-ql/etc/graphql/di.xml`

`<type name="Smile\ElasticsuiteCatalog\Model\Search\Request\Field\Mapper">
        <arguments>
            <argument name="fieldNameMapping" xsi:type="array">
                <item name="category_id" xsi:type="string">category.category_id</item>
            </argument>
        </arguments>
    </type>`

sky-hub: I have done some tests and adding
```xml
<type name="Smile\ElasticsuiteCatalog\Model\Search\Request\Field\Mapper">
    <arguments>
        <argument name="fieldNameMapping" xsi:type="array">
            <item name="category_id" xsi:type="string">category.category_id</item>
        </argument>
    </arguments>
</type>
``` 
to 
`module-elasticsuite-catalog-graph-ql/etc/graphql/di.xml`
does the job, however even if we add it here it will still affect the `$attributeNameMapping` used in
`\Smile\ElasticsuiteCatalogRule\Model\Rule\Condition\Product\AttributeList::getAttributeCollection`

It does no harm as long as the `category_id` field is located in the array after the 'category_ids' field but if the order is changed it will cause issue because `category_id` does not exist as eav_attribute.

The `$attributeNameMapping` will look like this 
```php
[
    'category_ids' => 'category.category_id',
    'position' => 'category.position',
    'price' => 'price.price',
    **'category_id' => 'category.category_id',**
]
```
and `$fieldNames` like this
```php
[
    ...
    'category.category_id' => 'category_ids'
    'category.position' => 'position'
    'category.is_parent => 'category.is_parent'
    'category.name' => 'category.name'
    ...
]
```

If you have any suggestions how to handle this one let's discuss it, however in my opinion there should be no duplicate mappings for the same attribute.

romainruaud:
We merged it as being a part of #2274 and #2272 (your commit was kept and you still appear as the author, we just added some more features on top of your Request Mapper).

So I close this PR, thank you for contributing !

PR #2129, opened 2021-04-28
Fixes #2078, add support numbers including zero to validate facet max size value for ElasticSuite 2.10.x

### LINKED ISSUE #2078: Facet max. size cannot be zero (although it can)
Setting the value "Facet max. size" in an attribute's settings the form cannot be submitted if the value in that field is 0.
I can empty the field and submit ok which results in the field acutally being set to 0. A subsequent submission will result in an error message "Please enter a number greater than 0 in this field.".

### Preconditions
PHP 7.3

Magento Version : 2.4.2 (CE)

ElasticSuite Version : latest

Environment : dev

### Steps to reproduce

1. Go to stores -> attributes -> product and pick any attribute
2. Enter attribute's edit form
3. set "Facet max. size" to "0"
4. Hit "Save Attribute"

### Expected result

1. The form should submit ok and save the values to database

### Actual result

1. Submission halts with validation error "Please enter a number greater than 0 in this field."

Basically the validator should handle 0 and an empty field the same. Or it should not automatically set an empty field to 0.

vahonc:


romainruaud:
Sounds good, you can duplicate it for 2.9.x

Regards

PR #2135, opened 2021-04-30
Catalog rules special attr. interface update

rbayet:
Allows to have operators and values for special attributes.
Allows building a custom search query using those operator and values.


romainruaud:
That's working perfectly on my 2.10.x environment.

We can merge

Issue #2077, opened 2021-02-16
Products List Preview and Sorting products are missing.


nagarajuk12:
### Preconditions

Magento Version : Magento/2.3 (B2B)

ElasticSuite Version : 6.8.0

Environment :  Development.

Third party modules : No

### Steps to reproduce

1. There are 16 products assigned to one category. But in the Products List Preview and Sorting showing only 9 products.

4. All products are simple & configurable with catalog search & assigned to the same category and website.

### Expected result

1. In the products list preview and sorting it should display 16 products.

### Actual result

1. https://prnt.sc/zmfq7u
2. https://prnt.sc/zmfuk7
3. https://drive.google.com/file/d/1Qns8-rfmPAhNuuTgA5WDUAagVyR5-vcu/view

Could you please advise how to fix this issue.

KZNcode:
We're having an issue that may relate to this as well... 

When we do a 'select all' the preview section completely disappears, and there's a serialization error after hitting save...

Selecting each product 1 by 1 works as expected... But as soon as you hit "select all" at any point of the selection process, the same above issue takes place.

romainruaud:
@KZNcode please fill a new issue, this seems to be different than OP's thread.

@KnK123 it looks like you have product without prices, if that's the case, they won't be indexed in the Magento legacy stock index, which may cause them to not appear in frontend (and also not on the BO preview, which reflects exactly the frontend).

What is your configuration about displaying out of stock products in frontend ? If your store is configured to not display them, that can also cause them to not appear in the preview.

Regards

KnK123:
@romainruaud  Thanks for your reply. Actually those are configurable products. Hence price is zero.

KnK123:
@KZNcode @romainruaud Could you please advise how to resolve this issue?

rameshre-kensium:
@romainruaud  Any update on this?

kerlama:
I have the same issue ... latest smile module and magento 2.4.1 and 2.4.2

KZNcode:
> I have the same issue ... latest smile module and magento 2.4.1 and 2.4.2
> 
> EDIT: I changed alias name and did full reindex and all products appeared.

@kerlama tried doing the same; didn't work in our case.

KnK123:
> I have the same issue ... latest smile module and magento 2.4.1 and 2.4.2
> 
> EDIT: I changed alias name and did full reindex and all products appeared.

@kerlama I have tried the same from my end but did not work.

[...]

PR #2147, opened 2021-05-27
Fix for #2077

thampe:
Hi, 

this should fix the described error in #2077. Feel free to check it out.


romainruaud:
Hi @thampe 

since you've found a fix, can you describe more the problem that is happening in this case and how this is fixing it ?

Regards

romainruaud:
@thampe I will not merge this without more explanations, can you please give me more details ?

PR #2205, opened 2021-07-14
Correct version comparsion to handle 7.10.0

alexeyamitech:
`strcmp` does a lexicographical comparison, it considers that '7.10.0' is less than '7.3.0'.
Fixed to use `version_compare` to make this plugin work for Elasticsearch 7.10.0


romainruaud:
cherry-picked into 2.10.x branch, thanks for contributing !

PR #2212, opened 2021-07-22
Adds <normsDisabled> tag for disabling field length normalization

gusdemayo:
Allows users to disable field length normalization on fields within their field mappings


romainruaud:
Looks promising.

I'm thinking about adding automatically "norms => false" to all non-searchable fields also, since norms are only used for scoring, it might save some disk space. I'll have to try this also.

romainruaud:
cherry-picked into 2.10.x, thank you for contributing !!

PR #2225, opened 2021-08-05
enhance server config comment

torhoehn:
Because the configured server list is only exploded by `,` I enhanced the input comment a little bit.


romainruaud:
The host is (or should be) trimmed here : https://github.com/Smile-SA/elasticsuite/blob/2.10.x/src/module-elasticsuite-core/Client/ClientBuilder.php#L138

If there is a bug when setting the configuration with a space, I'd rather also fix it and not just enhance the comment :) 

Let me know

PR #2228, opened 2021-08-06
[Optimizer] Add Request Type & Boost Weight in Optimizers List #ESP-208 #ESP-209

PierreGauthier:


src/module-elasticsuite-catalog-optimizer/Ui/Component/Optimizer/Listing/DataProvider.php
```diff
@@ -66,6 +66,15 @@ public function __construct(
...
         $this->collection = $collectionFactory->create();
+        // Add config column in order to get boost weight value.
+        $this->collection->addFieldToSelect('config');
+
+        // Join search container table in order to show request type in grid.
+        $this->collection->join(
+            ['search_container' => 'smile_elasticsuite_optimizer_search_container'],
+            'search_container.optimizer_id = main_table.optimizer_id',
+            ['search_container']
+        );
```
rbayet, line 77: The best approach is to use `$addFieldStrategies` and `$addFilterStrategies` if you're joining an extra table, it's less instrusive.
It might probably apply to `config` too.

Anyway, locally (with the third party AB testing module) 

- the request type column is empty
- I get an error when applying the filter

PS: I haven't tested yet on a legacy/pure CE 2.3.4

src/module-elasticsuite-catalog-optimizer/Ui/Component/Listing/Column/BoostWeight.php
```diff
@@ -0,0 +1,72 @@
...
+     *
+     * @param array $dataSource
+     * @return array
+     */
+    public function prepareDataSource(array $dataSource)
+    {
+        if (isset($dataSource['data']['items'])) {
+            foreach ($dataSource['data']['items'] as & $item) {
+                $config = $this->serializer->unserialize($item['config']);
+                $item[$this->getData('name')] = $config['constant_score_value'] ?? '';
```
rbayet, line 66: Either there or somewhere else, I would add a "%" symbol, since we are displaying a score boost (50 = 50% boost of the score), and align the value to the right (I think it can be done through the listing XML).
See the explain module for reference.

rbayet (changes requested):

src/module-elasticsuite-catalog-optimizer/Ui/Component/Optimizer/Listing/DataProvider.php
PierreGauthier: I updated with the `$addFieldStrategies` and the column seems ok and the filtering is working
There was indeed a problem with the explain module which was overriding the optimizer data provider with inherit from it

PierreGauthier: (https://github.com/Elastic-Suite/magento2-elasticsuite-premium/pull/54)

src/module-elasticsuite-catalog-optimizer/Ui/DataProvider/AddSearchContainerFieldToCollection.php
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function addField(Collection $collection, $field, $alias = null)
+    {
+        $collection->join(
+            ['search_container' => $this->resource->getTableName('smile_elasticsuite_optimizer_search_container')],
+            'search_container.optimizer_id= main_table.optimizer_id',
+            [$field]
+        );
```
rbayet, line 50: Actually, a straight join will not work for optimisers configured with multiple request types, the relation between smile_elasticsuite_optimizer and smile_elasticsuite_optimizer_search_container is 1..N

You will get duplicate smile_elasticsuite_optimizer rows the the classical 
```
Item (Smile\ElasticsuiteCatalogOptimizer\Model\Optimizer) with the same ID "1" already exists.
```

rbayet (changes requested):

src/module-elasticsuite-catalog-optimizer/Ui/Component/Listing/Column/BoostWeight.php
```diff
@@ -0,0 +1,72 @@
...
+     * Prepare Data Source
+     *
+     * @param array $dataSource
+     * @return array
+     */
+    public function prepareDataSource(array $dataSource)
+    {
+        if (isset($dataSource['data']['items'])) {
+            foreach ($dataSource['data']['items'] as & $item) {
+                $config = $this->serializer->unserialize($item['config']);
```
rbayet, line 65: By using a field strategy for the request types and reverting the changes to the data provider, you also removed the "config" field so this throws a notice (and probably does not work anymore at all ?)

rbayet (approved):

PR #2235, opened 2021-08-25
Is discount configurable grouped

Bashev:
This pull request mark configurable and grouped products as discouned (is_discount = true) when some (or all) of their childs are discounted.

#866


Bashev:
Wrong source branch of the issue. See #2237

PR #2236, opened 2021-08-25
Fix issues where added type was replacing and not being added to the original TypeLocators

igorwulff:
When adding a type for Magento where other items are already inserted by other modules, the same scope needs to be used. Otherwise the item will replace instead of be added to the argument value list.

Also see: 
https://stackoverflow.com/questions/34635615/magento-2-how-to-add-new-value-to-the-array-on-di-xml

I compared two projects one with out smile and one with smile, and noticed this peculair behaviour.

This can be tested by doing a POST on rest/all/V1/products with a body like this:
{
    "product": {
        "sku": "sku",
        "custom_attributes": [
            {
                "attribute_code": "test",
                "value": "one"
            }
        ]
    }
}

The attribute needs to be a dropdown with options. If something else is given besides an int value, the following error should occur:
{
    "message": "Error occurred during \"custom_attributes\" processing. Attribute \"test\" has invalid value. The \"\" value's type is invalid. The \"int\" type was expected. Verify and try again.",
    "trace": null
}

This doesn't happen when the Smile module is installed, since the Typelocators are being replaced and not checked during a rest call.


romainruaud:
Good catch, I was not aware of this particular behavior.

PR #2237, opened 2021-08-25
Discount grouped and configurable

### LINKED ISSUE #866: Configurable products shown in virtual category as discounted products, when they are not

Bashev:
Product will be marked as "discounted" if some (or all) of their children have some discount. 
- works with configurable products which not matter what are their prices - equal or not. 
- works with grouped products.

On my tests (20k catalog, 5k grouped) i not see performance problems on reindex process.

#### Related Issue

- Resolve #866


romainruaud:
I'm afraid about performances, even if you say there is no issue, I can see a product load through the repository :/

Bashev:
@romainruaud, thank you for your comment. I have the same concerns, but I haven't made any additional tests.

I'll try to benchmark with large catalog in three different scenarios:
1. Without this pull request.
2. With this code.
3. Fill arrays with configurable and grouped by ProductCollection

Do you have some different suggestion based on your exprience?

Bashev:
@romainruaud can you review this pull request. All of ProductRepository relations are removed.

https://github.com/Smile-SA/elasticsuite/pull/2677

PR #2278, opened 2021-10-08
Feat esp 252 fix boolean and filter field mapping 2.10.x

rbayet:


MaximGns:
Hi, we are running on a magento 2.4.3. And we encounter the same issue that boolean attributes only display the 'yes' option but not the 'No'. Is this going to be released soon?

PR #2286, opened 2021-10-12
Fix use magento parameters

Teods:
## Problem still present
A problem still persist, on first deployment the search-engine "elasticsuite" is not found.
It seems comming from magento does not have the di:compile runned and search the engine inside the directory `setup/src/Magento/Setup/Model/SearchConfigOptionsList.php`

## Fixed
For the rest now all "magento elasticsearch" parameters are correctly used on first installation (with no env.php and config.php file).
The elasticsearch index are correctly prefixed, the env.php file contains the elasticsuite configuration etc...


romainruaud:
I close this one, PR #2285 fixes the initial issue, thank you for pointing us the issue.

PR #2326, opened 2021-11-01
Update category_uid attribute mapping

JeroenVanLeusden:
Update of https://github.com/Smile-SA/elasticsuite/pull/2315.

`category_uid` is the base64 encoded value of `category_id` and not a copy.

- https://devdocs.magento.com/guides/v2.4/graphql/queries/category-list.html
- https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/CatalogGraphQl/Model/Category/Hydrator.php#L66

### DIFF
--- a/src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/CategoryData.php
+++ b/src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/CategoryData.php
@@ -14,6 +14,7 @@
 
 namespace Smile\ElasticsuiteCatalog\Model\Product\Indexer\Fulltext\Datasource;
 
+use Magento\Framework\GraphQl\Query\Uid;
 use Smile\ElasticsuiteCore\Api\Index\DatasourceInterface;
 use Smile\ElasticsuiteCatalog\Model\ResourceModel\Product\Indexer\Fulltext\Datasource\CategoryData as ResourceModel;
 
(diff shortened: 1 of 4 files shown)


src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/CategoryData.php
```diff
@@ -57,9 +57,10 @@ public function addData($storeId, array $indexData)
             $categoryDataRow = array_merge(
                 $categoryDataRow,
                 [
-                    'category_id' => (int) $categoryDataRow['category_id'],
-                    'is_parent'   => (bool) $categoryDataRow['is_parent'],
-                    'name'        => (string) $categoryDataRow['name'],
+                    'category_id'   => (int) $categoryDataRow['category_id'],
+                    'category_uid'  => base64_encode((string) $categoryDataRow['category_id']),
```
romainruaud, line 61: can you add some local caching ? 

and also do it with ``$this->uidEncoder->encode((string) $category->getId());`` ? This will require injecting the uidEncoder but will ensure that if Magento implementation of the encoder changes, we also benefit the change.

romainruaud (changes requested):

src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/CategoryData.php
JeroenVanLeusden: Why is caching the `category_uid` necessary if every `$categoryDataRow` is unique?

romainruaud: let's imagine we have product A, B and C belonging to category Y.

we will compute base64_encode(Y) three times (for each row), but we could compute it only one time and get it from cache on next calls.

JeroenVanLeusden: @romainruaud Added local caching.

romainruaud:
Okay, sounds good to me like this (technically).

Can you just fix the remaining code quality failures (you can see them by opening the check details) and this will be ready for merge.

Regards

JeroenVanLeusden:
@romainruaud done!

romainruaud:
Sorry, I finally reverted your PR : ```Magento\Framework\GraphQl\Query\Uid``` does not exist in Magento 2.4.1 which is a supported version for us.

Shame on me, I forgot to launch the integration tests on your PR, I only saw the failure after it was merged :/ 

Can you open a new PR containing something that could be backward compatible ? I think just using base64_encode for now would be fine until we drop support of Magento 2.4.1 and are able to use their Uid Encoder. 

Or maybe something inspired of this, which can be seend as similar (test if a class provided by MSI is existing, otherwise do something else) : https://github.com/Smile-SA/elasticsuite/blob/2.10.x/src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/InventoryData.php#L88

Sorry

PR #2327, opened 2021-11-02
Wrong casing (typo) in LayeredNavAttribute namespace

peterjaap:
This causes our installation to break since we dump the composer autoloader with the `--classmap-authoritative` option


romainruaud:
Good catch ! We probably need to add this to our automated test scenario.

I'll merge it soon and probably release a 2.10.8.1 containing the fix.

Thank you !

romainruaud:
Feel free to move to 2.10.8.1 which contains two urgent bugfixes including this one.

Issue #2343, opened 2021-11-11
Invalid argument supplied for foreach()


hbarza:
In case there is a virtual category with no virtual rule defined the following error will occur when opening the category in storefront.

```
Error filtering template: Warning: Invalid argument supplied for foreach() in .../vendor/smile/elasticsuite/src/module-elasticsuite-virtual-category/Plugin/Widget/ProductsListPlugin.php on line 101
```

### Preconditions

PHP Version: 7.4

Magento Version : 2.4.3-p1

ElasticSuite Version : 2.10.8.1

Environment : developer

Third party modules :

### Steps to reproduce

1. Create a new category
2. Enable the "Virtual Category" option under "Products in Category" section
3. Select a category as root for the "Virtual Category Root" option
4. Set the "Virtual rule" option into "If **ALL**  of these conditions are **TRUE**"; BUT define no condition
5. Save the category and open it in the storefront

### Expected result

View products in the category/widget

### Actual result

```
Error filtering template: Warning: Invalid argument supplied for foreach() in .../vendor/smile/elasticsuite/src/module-elasticsuite-virtual-category/Plugin/Widget/ProductsListPlugin.php on line 101
```

romainruaud:
I'll review PR #2344 

Thank you for contributin.

romainruaud:
Fix has been merged and will be part of next minor release scheduled for january.

Regards

PR #2344, opened 2021-11-11
fix/virtual-category (#2343): invalid argument supplied for foreach

hbarza:
In case there is a virtual category with no virtual rule defined the following
error will occur when opening the category in storefront.

    ```
    Error filtering template: Warning: Invalid argument supplied for foreach() in .../vendor/smile/elasticsuite/src/module-elasticsuite-virtual-category/Plugin/Widget/ProductsListPlugin.php on line 101
    ```

Reproduce:

1. Create a new category
2. Enable the "Virtual Category" option under "Products in Category" section
3. Select a category as root for the "Virtual Category Root" option
4. Set the "Virtual rule" option into "If **ALL**  of these conditions are **TRUE**"; BUT define no condition
5. Save the category and open it in the storefront


src/module-elasticsuite-virtual-category/Plugin/Widget/ProductsListPlugin.php
```diff
@@ -94,10 +94,11 @@ public function afterCreateCollection(ProductsList $subject, $collection)
...
         if ($subject->getData('condition_option') == 'condition' || !$subject->getData('condition_option')) {
-            $conditions = $subject->getData('conditions_encoded') ?: $subject->getData('conditions');
-            if ($conditions) {
-                $conditions = $this->conditionsHelper->decode($conditions);
+            if (!$conditions = $subject->getData('conditions_encoded') ?: $subject->getData('conditions')) {
+                return $collection;
             }
+
+            $conditions = $this->conditionsHelper->decode($conditions);
             foreach ($conditions as $condition) {
```
romainruaud, line 102: I think your solution is over-complicated.

I'd prefer something like this (with just one guard condition) : 

```php
$conditions = $subject->getData('conditions_encoded') ?: $subject->getData('conditions');

if ($conditions) {
    $conditions = $this->conditionsHelper->decode($conditions);
    foreach ($conditions as $condition) {
    ... legacy ...
    }    
}

return $collection
```

romainruaud (changes requested):

PR #2386, opened 2021-12-10
Fixes #2368, avoid use 'match' as an identifier

### LINKED ISSUE #2368: Cannot use 'match' as an identifier. It is a reserved keyword since PHP 8.0
Hello folks!

Elasticsuite 2.10.8.1 is not compatible with PHP 8.0 (Magento 2.4.4-beta2)

### Preconditions

- PHP 8.0
- Magento Version : (ee) 2.4.4-beta2
- ElasticSuite Version : 2.10.8.1
- Magento Environment : Developer (same with Production mode)

### Steps to reproduce

```
composer create-project --repository-url=https://repo.magento.com/ magento/project-enterprise-edition=2.4.4-beta2 ee_244-beta2
composer require smile/elasticsuite ~2.10.0
php bin/magento setup:di:compile
```

### Expected result

Elasticsuite should be compiled in `generated` without errors

### Actual result

```
syntax error, unexpected token "match", expecting identifier#0 /var/www/html/vendor/composer/ClassLoader.php(429): Composer\Autoload\includeFile('/var/www/html/v...')
#1 [internal function]: Composer\Autoload\ClassLoader->loadClass('Smile\\Elasticsu...')
```
My error is truncated, but it's:
- `Smile\ElasticsuiteCore\Search\Adapter\Elasticsuite\Request\Query\Builder\Match` class
- `Smile\ElasticsuiteCore\Search\Request\Query\Match` class
- `Smile\ElasticsuiteCore\Search\Request\Query\Common extends Match`
IDE Error: Cannot use 'match' as an identifier. It is a reserved keyword since PHP 8.0

vahonc:


src/module-elasticsuite-core/Search/Adapter/Elasticsuite/Request/Query/Builder/QueryMatch.php
```diff
@@ -24,7 +24,7 @@
  * @package  Smile\ElasticsuiteCore
  * @author   Aurelien FOUCRET <aurelien.foucret@smile.fr>
  */
-class Match implements BuilderInterface
+class QueryMatch implements BuilderInterface
```
romainruaud, line 27: Can you name it "MatchQuery" instead of "QueryMatch" ?

src/module-elasticsuite-core/Test/Unit/Search/Adapter/Elasticsuite/Request/Query/Builder/MatchTest.php
```diff
@@ -13,8 +13,8 @@
  */
 namespace Smile\ElasticsuiteCore\Test\Unit\Search\Adapter\Elasticsuite\Request\Query\Builder;
 
-use Smile\ElasticsuiteCore\Search\Request\Query\Match as MatchQuery;
-use Smile\ElasticsuiteCore\Search\Adapter\Elasticsuite\Request\Query\Builder\Match as MatchQueryBuilder;
+use Smile\ElasticsuiteCore\Search\Request\Query\QueryMatch as MatchQuery;
```
romainruaud, line 16: If you name it "MatchQuery" instead of "QueryMatch" this will allow to get rid of this aliasing "QueryMatch as MatchQuery".

Also, since the builder was named "MatchQueryBuilder", this would be more consistent.

romainruaud (changes requested):

src/module-elasticsuite-core/Search/Adapter/Elasticsuite/Request/Query/Builder/QueryMatch.php
vahonc: Done.

src/module-elasticsuite-core/Test/Unit/Search/Adapter/Elasticsuite/Request/Query/Builder/MatchTest.php
vahonc: Done.

PR #2419, opened 2022-01-10
Fix for missing server name definition

adamkarnowka:
In some environments (for example CI/CD) server name value is not set, resulting in:

```
Notice: Undefined index: SERVER_NAME in /var/www/html/vendor/smile/elasticsuite/src/module-elasticsuite-tracker/Block/Variables/Page/Base.php:137.
```

This change uses `requestInterface` to fetch server name if exists, returning empty string otherwise.


LukasKiederle:
Thx for the fix, we have the same problem.

PR #2422, opened 2022-01-13
Fix #2413 Remove AdminAnalytics Dependency

### LINKED ISSUE #2413: Telemetry causing admin errors
Magento Version : 2.4.3-p1 CE

ElasticSuite Version : 2.10.9

Environment : Developer

### Steps to reproduce

1. As part of our optimization/performance effort, we disable magento core modules that we don't use. Therefore, Magento_AdminAnalytics module is disabled
2. Upon admin login, we get an error that table `admin_analytics_usage_version_log` does not exist. We created the table manually so the admin will load.
3. I go to ES config settings and disable Telemetry
4. When I return to the dashboard, there is a console error for `Uncaught Error: Script error for: Magento_AdminAnalytics/js/modal/component`

### Expected result

1. Support should be added for when Magento_AdminAnalytics is not available

### Actual result

1. Currently, errors are given, even when ES Telemetry is disabled

PierreGauthier:


src/module-elasticsuite-tracker/ViewModel/Notification.php
```diff
@@ -29,35 +27,26 @@
...
-    private $canViewNotificationElasticsuite;
-
-    /**
-     * @var AdminAnalyticsNotification
+     * Prefix for admin analytics cache
+     *
+     * @var string
      */
-    private $canViewNotificationAnalytics;
+    private static $adminAnalyticsCachePrefix = 'admin-usage-notification-popup';
```
romainruaud, line 34: @PierreGauthier is this variable really used ? PHPMD is complaining on the 2.9.x based PR : https://github.com/Smile-SA/elasticsuite/runs/4805899394?check_suite_focus=true

PR #2459, opened 2022-02-24
Fixes #2410, replace attributes collection by aggregating data of "indexed_attributes"

### LINKED ISSUE #2410: Attributes/Filters missing in category admin
On the category administration page under "Display Settings" I can configure the attributes being used in a particular category.
In a project of mine there seems to be an issue with the attributes since there are quite some attributes missing on that page and I have no way to configure them.

### Preconditions

Magento Version : 2.4.3-p1

ElasticSuite Version : 7

Environment : issue happens in both modes

Here you can see a list of configured attributes. Please don't get confused by multiple attributes with the same title. They are actually separate attributes with different attribute codes only with the same title.

smile-cat-1

The goal here was to only have the manufacturer ("Hersteller") attribute shown in this category.
But in the frontend there are three more attributes which don't show up in the backend configuration:

smile-cat-2

I have no idea what is going on here or why they are missing on the admin page. 
All that data gets written to the table `smile_elasticsuitecatalog_category_filterable_attribute`. I added the missing attributes into this table manually after which they disappeared in the frontend as expected. But of course another save in the backend will always kill these lines again.

Are there any known limitations I might have crossed?
If this is a bug can you point me to the relevant parts of the code to debug?

Thank you

vahonc:


romainruaud:
Hi Vadym, I think GraphQL is now broken. It attempts to fetch data from "indexed_attributes" aggregated results, which fails.

Can you add it there ? : https://github.com/Smile-SA/elasticsuite/blob/2.10.x/src/module-elasticsuite-catalog-graph-ql/DataProvider/Product/LayeredNavigation/Builder/Attribute.php#L59

This should be ignored as "attribute_set_id". 

Regards

PR #2467, opened 2022-03-04
Make jQuery ajax request for autocomplete cacheable

lumnn:
I've noticed on my magento installation 2.4.3 search requests are appended with `&_=[timestamp]` which makes requests slower than they could be.

[jQuery ajax documentation](https://api.jquery.com/jquery.ajax/#jQuery-ajax-settings) mentions a `cache` option for ajax calls, which controls this behaviour.

The ajax() funtction is called without `cache` parameter making it work kind of automatically as documentation says.
I'm not sure how that exactly is resolving to false - I guess it may be considering dataType beeing 'jsonp' or 'script'. Therefore, in this PR I'm adding a `cache: true` option to make sure the timestamp is not appended as well as I'm setting `dataType: 'json'` to make sure jQuery expects 'json' which is what is beeing returned.

This PR also adds extra space for style purposes.

Lastly here's a comparasion on my local machine with large database of products before and after changes:

With cache timestamp:

Screenshot from 2022-03-04 12-54-42

Without cache timestamp:

Screenshot from 2022-03-04 12-55-16

### DIFF
--- a/src/module-elasticsuite-core/view/frontend/web/js/form-mini.js
+++ b/src/module-elasticsuite-core/view/frontend/web/js/form-mini.js
@@ -239,7 +239,9 @@ define([
                 this.currentRequest = $.ajax({
                     method: "GET",
                     url: this.options.url,
-                    data:{q: value},
+                    cache: true,
+                    dataType: 'json',
+                    data: {q: value},
                     // This function will ensure proper killing of the last Ajax call.
                     // In order to prevent requests of an old request to pop up later and replace results.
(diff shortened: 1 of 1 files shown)


romainruaud:
Hi @lumnn 

I'm not sure we should merge this "as is".

Adding caching capabilities on autocomplete results is fine for me, but it would imply also to add a proper clean-cache mechanism for them, as it's done by Magento for the catalogsearch result pages.

regards

PR #2500, opened 2022-03-30
Fixes #2490, replacing retrieving queryId instead queryText

### LINKED ISSUE #2490: Optimizers: Are Search Terms Case Sensitive?
When creating a new optimizer for the term "Straw", if I use the "Preview Search" of an optimizer and place the term "straw" into the preview engine it does not result in the same result.  Should I expect this field to be case-sensitive? 

I created a messy ticket about this: https://github.com/Smile-SA/elasticsuite/issues/2456 but I think a Question would be a first start on half of the issue.  

Ex :
- I cannot figure how [...]
- Is ElasticSuite able to [...]

**Additional context**
Add any other context or screenshots about your question here.

vahonc:


vahonc:
Done. But now we have warnings for the PHPUnit section..

romainruaud:
Yes, in the PHPUnit test we instantiate the Preview object, so if you added a parameter to the constructor, you have to also add it on the test when we create the object.

https://github.com/Smile-SA/elasticsuite/blob/2.10.x/src/module-elasticsuite-catalog-optimizer/Test/Unit/Model/Optimizer/PreviewTest.php#L90

Here you have to create a mock for Search Context (like for other parameters), and inject it into the Preview Object that we are testing.

Then we'll be able to see if the legacy unit tests are still passing or if we need to adapt something.

Regards

Issue #2494, opened 2022-03-18
Magento 2.4.4 compatibility


romainruaud:
Magento 2.4.4 will be live at the end of April.

Actually, we only "build" the 2.10.x branch against beta builds of Magento, and everything goes well.

However, we now need to enter the last stage of testing it, by going through all the Elasticsuite related screens in the back-office on a 2.4.4 version of Magento to check everything is operating normally.

@vahonc as soon as other priority tasks are done, you can engage some time on this one.

Regards

vahonc:
It seems that on Magento 2.4.4 we are not able to create New Optimizers because this page is broken.

![Screenshot from 2022-04-07 10-18-40](https://user-images.githubusercontent.com/67791361/162152962-854025db-12f1-404c-b236-5e10ab1d7601.png)

We can't add any Rule conditions for Optimizer and Boost value (%) slider from the Boost & Bury section for Constant Model is also broken.

In Console I see errors: [ERROR] Failed to load the "..." component.

I'll try to dig into this.

romainruaud:
![image](https://user-images.githubusercontent.com/15340849/162155950-c9510565-907c-4ce4-9f47-caf3b014b022.png)

This file was here on Magento 2.4.3 (and since the inception).

https://github.com/magento/magento2/blob/2.4.3/lib/web/MutationObserver.js

It's missing in 2.4.4 (probably they removed it ...) : 

https://github.com/magento/magento2/tree/2.4-develop/lib/web

vahonc:
Yes, according to Magento 2.4.4 Release Notes the jQuery UI has been upgraded to the latest version (v1.13.0). And some jQuery components for the v1.10.0 version have been removed. Therefore many pages where we use these old components could be broken.

romainruaud:
https://community.magento.com/t5/Magento-DevBlog/Technical-platform-changes-to-expect-in-Adobe-Commerce-2-4-4/ba-p/485506

Built-in JS libraries 

Upgrade of jQuery, TinyMCE, knockout and underscore library. 
**Removal of Modernizr, es6-collections, FormData.js & MutationObserver.js library** 

Thank you Adobe... once again, nice backward compatibility for a minor release.

romainruaud:
I think we can re-add the MutationObserver.js library, probably in the ElasticsuiteCatalogRule module since it's used both by virtual categories and optimizers.

[...]

PR #2516, opened 2022-04-07
2.4.4 Support (#2494) - Adding base OpenSearch compatibility checks

chris-brabender:
Adds checks to version comparison functions to support 1.x versioning for OpenSearch to help with 2.4.4 Compatibility - https://github.com/Smile-SA/elasticsuite/issues/2494

Follows some guidelines outlined here - https://github.com/opensearch-project/OpenSearch/issues/693

Checks for a distribution variable in version info returned by search service.

### DIFF
--- a/src/module-elasticsuite-core/Cluster/ClusterInfo.php
+++ b/src/module-elasticsuite-core/Cluster/ClusterInfo.php
@@ -36,6 +36,11 @@ class ClusterInfo implements ClusterInfoInterface
     private $serverVersion;
 
     /**
+     * @var string
+     */
+    private $serverDistribution;
+
+    /**
      * Constructor.
      *
      * @param \Smile\ElasticsuiteCore\Api\Client\ClientInterface $client ElasticSearch client.
--- a/src/module-elasticsuite-core/Cluster/ClusterInfo.php
+++ b/src/module-elasticsuite-core/Cluster/ClusterInfo.php
@@ -56,4 +56,16 @@ class ClusterInfo implements ClusterInfoInterface
 
         return $this->serverVersion;
     }
+
+    /**
+     * {@inheritDoc}
+     */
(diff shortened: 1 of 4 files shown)


src/module-elasticsuite-core/Cluster/ClusterInfo.php
```diff
@@ -56,4 +56,16 @@ public function getServerVersion()
 
         return $this->serverVersion;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getServerDistribution()
+    {
+        if ($this->serverDistribution === null) {
```
romainruaud, line 65: Can you declare this property ? Build is failing with PHP 7.4 and I think that's because it's less permissive with undeclared variable. For now we still hope our next minor would be compatible will also Magento 2.4.3.

romainruaud (changes requested):

src/module-elasticsuite-core/Cluster/ClusterInfo.php
romainruaud: Notice: Undefined property: Smile\ElasticsuiteCore\Cluster\ClusterInfo::$se  
  rverDistribution in /var/www/magento/vendor/smile/elasticsuite/src/module-e  
  lasticsuite-core/Cluster/ClusterInfo.php on line 65

romainruaud:
Hi @chris-brabender, many many thanks for taking some time to contribute this, this saves us a lot of time. There is just one minor thing to change on this one and I think we'll be good.

regards

src/module-elasticsuite-core/Cluster/ClusterInfo.php
chris-brabender: Eesh sorry about that, I had it in my main test env but forgot for this PR, added now.

PR #2519, opened 2022-04-08
[FEATURE][GWSM2-1057] - Applied autocomplete on search results page

rensieeee:
When entering a partial search term, the search results page shows different results than the autocomplete. This is because the search terms are not fully added upon on search, but are on autocomplete. This PR fixes that. It does need a bit more refactoring since there's now an unused argument, but this is the idea, and it works on 2.4.3.


romainruaud:
For me this does not make sense to introduce such changes for the search result page. It would lead to a lot of inconsistency for many existing customers.

If that fixes issues for your particular project, feel free to use it, but we'll not merge this into the core.

Regards

PR #2523, opened 2022-04-09
fallback to magento category product positions

Cloudy777:
uses magento default product category position if non are set in elasticsuite table to respect position of migrations or imports


rbayet:
Hello @Cloudy777,

What about people who migrated from a non-Elasticsuite install with legacy product positions to an Elasticsuite install seeing the benefit of Optimizers for automated / semi-manual category merchandizing ?

At the first mandatory reindex, old positions would creep pop in the search engine index and mess with the natural ordering of products in a ninja way since those positions will not be displayed as "manual positions" in the category preview zone.

Any optimizer applicable to catalog navigation would seem to randomly work or not.

I'd be more in favor of a (documented) cli command to migrate the legacy positions into the Elasticsuite specific table.

Regards

PR #2544, opened 2022-04-26
[BUGFIX][GWSM2-1070] Change categoryfilter back to filter behaviour i…

florisschreuder:
…n searchcontext

Currently, when 'use URL rewrites' is on in the configuration, the category-filter behaves not as a filter, but as a hard link to the categories. This is expected - and wanted - behaviour within category context, but this behaviour does not make sense when searching.

After all, when searching, the user doesn't expect to get sent from the catalogsearch page to a categorypage when using the categoryfilter.

Another option would be to add this to the configuration ("Use URL Rewrites within Catalog Search"), but I don't think there's a good use-case for this.

### DIFF
--- a/src/module-elasticsuite-catalog/Model/Layer/Filter/Category.php
+++ b/src/module-elasticsuite-catalog/Model/Layer/Filter/Category.php
@@ -216,6 +216,11 @@ class Category extends \Magento\CatalogSearch\Model\Layer\Filter\Category
      */
     protected function useUrlRewrites()
     {
+        /** If within search context, always make categories behave as filter */
+        if ($this->searchContext->getCurrentSearchQuery()) {
+            return false;
+        }
+        
         return $this->useUrlRewrites;
     }
 
--- a/src/module-elasticsuite-catalog/Model/Layer/Filter/Category.php
+++ b/src/module-elasticsuite-catalog/Model/Layer/Filter/Category.php
@@ -220,7 +220,7 @@ class Category extends \Magento\CatalogSearch\Model\Layer\Filter\Category
         if ($this->searchContext->getCurrentSearchQuery()) {
             return false;
         }
-        
+
         return $this->useUrlRewrites;
     }
 


rbayet:
Hello @florisschreuder,

Are you sure the behaviour you're encountering is not due to some customization on your part/some DI problem ?

Because the layered navigation category filter model is not **quite** the same between catalog navigation and fulltext search, see : 
* https://github.com/Smile-SA/elasticsuite/blob/2.10.x/src/module-elasticsuite-catalog/etc/frontend/di.xml#L61-L91 
* and https://github.com/Smile-SA/elasticsuite/blob/2.10.x/src/module-elasticsuite-catalog/Model/Layer/Filter/Category.php#L82-L114

Per DI argument injection, ` $this->useUrlRewrites` should always be false for the fulltext search layered navigation category filter.

And this is a behavior that the Virtual Category module (if enabled) keeps : https://github.com/Smile-SA/elasticsuite/blob/635635a21ce73ccafccc129799403c1976794346/src/module-elasticsuite-virtual-category/etc/frontend/di.xml#L21-L25

Regards,

florisschreuder:
Hi @rbayet ,

Thanks for the quick response!

If I'm reading this right, this means:
- The Di.xml 'forces' the categoryfilter to use the 'hard' URLs on category-pages
- The setting in the backend (`smile_elasticsuite_catalogsearch_settings/catalogsearch/category_filter_use_url_rewrites`) is only used on search pages

Apparently, it was indeed a change in a module of mine that changes the CategoryFilter which also changes this behaviour (removes the virtualType). Apologies!

However, I then would suggest changing this in the system.xml:
image

I read this as "category navigation" = "category context". And was therefore confused.

florisschreuder:
Hi @rbayet ,

Thanks for your patient explanation!

It took me a while, but I found my mistake. My preference on that class to change the behaviour of the filter also - wrongfully - put the _useUrlRewrites_ to true always.

Hopefully next PR I submit **is** useful ;)

PR #2593, opened 2022-06-03
Fixes #2590 Possible XSS issues from contributed data

### LINKED ISSUE #2590: Elasticsuite affects Magneto 2.4.3-p1 XSS prevention strategies
After installation of the Elasticsuite module, both magento2 frontend and backend allows script execution over input tags.
### Preconditions

Magento Version : Magneto 2.4.3-p1  CE
PHP Version: 7.4.3
ElasticSuite Version :2.10.10
Environment : Developer and Production

### Steps to reproduce
1.Install using composer require smile/elasticsuite ~2.10.0
2.create a product with name  - <script>alert("hello")</script>
3. Add it to default category.
4. Load the default category page in admin dashboard.  It will show an alert message.
5. The same will happen in frontend product detail page also.

### Expected result
1. It should not execute the script.

### Actual result

1. It is executing the script and show an alert message.
image
image

rbayet:
By using correct escaping/binding methods of underscore (frontend) and knockoutjs (backend)

### DIFF
--- a/src/module-elasticsuite-catalog/view/frontend/web/template/attribute-filter.html
+++ b/src/module-elasticsuite-catalog/view/frontend/web/template/attribute-filter.html
@@ -14,13 +14,13 @@
                    data-bind="checked: is_selected, attr: {id: id}"
                    onclick="this.parentNode.click();" />
             <label data-bind="attr: {for: id}">
-                <span data-bind="html: label"></span>
+                <span data-bind="text: label"></span>
                 <span class="count" data-bind="text: count, visible: $parent.displayProductCount"></span>
             </label>
         </a>
 
         <div data-bind="visible: count < 1">
-            <span data-bind="html: label"></span>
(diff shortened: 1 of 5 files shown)


src/module-elasticsuite-catalog/view/frontend/web/template/attribute-filter.html
```diff
@@ -14,13 +14,13 @@
...
             <label data-bind="attr: {for: id}">
-                <span data-bind="html: label"></span>
+                <span data-bind="text: label"></span>
                 <span class="count" data-bind="text: count, visible: $parent.displayProductCount"></span>
             </label>
         </a>
 
         <div data-bind="visible: count < 1">
-            <span data-bind="html: label"></span>
+            <span data-bind="text: label"></span>
```
rbayet, line 23: A bit of an overkill considering the label is already HTML escaped in PHP.

PR #2637, opened 2022-07-13
Add "IsSaleable" special attribute for optimizers

### LINKED ISSUE #2618: Optimizer "Only in stock products" with backorders
When creating an optimizer, the "Only in stock products" is only effective when backorders are disabled. If backorders are enabled then all products are set to "In Stock" despite the quantity being <= 0.

With backorders enabled, you may still wish to prioritise products which as in stock and available to ship immediately, but this is not currently possible.

### Preconditions
* Enable backorders
* Set product(s) to "In Stock" and QTY to 0 or less
* Set up Optimizer to prioritise "Only in stock products"

Magento Version : 2.4.3-p2

ElasticSuite Version : 2.10.10

Environment : Developer/Production

### Steps to reproduce

1. Enable backorders
2. Set product(s) to "In Stock" and QTY to 0 or less
3. Set up Optimizer to prioritize "Only in stock products"
4. Visit category page

### Expected result

1. Products with less than 1 quantity should be at bottom of the product grid.

### Actual result
1. Products with less than 1 quantity are not at bottom of the product grid.

vseager:
Resolves #2618 .

Currently if you use the "Only in stock products" rule in an optimizer when you have backorders enabled, you don't get the expected results.

To be able to backorder a product it needs to be set to "In Stock" even if the quantity is 0.

This means that products which are available to order but not physically in stock are currently prioritised in the optimizer and there is no way to prioritise ONLY product which are physically in stock and ready to ship.

This PR achieves two things:

* Modify `IsInStock` special attribute to return products which are both set to "In Stock" and have a `qty` greater than `0`.
* Add new `IsSaleable` special attribute which returns all "In Stock" products regardless of qty.

### DIFF
--- a/src/module-elasticsuite-catalog-rule/Model/Rule/Condition/Product/SpecialAttribute/IsSaleable.php
+++ b/src/module-elasticsuite-catalog-rule/Model/Rule/Condition/Product/SpecialAttribute/IsSaleable.php
@@ -13,39 +13,37 @@
  */
 namespace Smile\ElasticsuiteCatalogRule\Model\Rule\Condition\Product\SpecialAttribute;
 
-use Magento\Config\Model\Config\Source\Yesno;
+use Magento\Config\Model\Config\Source\Yesno\Magento\Config\Model\Config\Source\Yesno;
 use Smile\ElasticsuiteCatalogRule\Api\Rule\Condition\Product\SpecialAttributeInterface;
 use Smile\ElasticsuiteCatalogRule\Model\Rule\Condition\Product as ProductCondition;
 use Smile\ElasticsuiteCore\Search\Request\QueryInterface;
 use Smile\ElasticsuiteCore\Search\Request\Query\QueryFactory;
 
 /**
- * Special "is_in_stock" attribute class.
+ * Special "is_saleable" attribute class.
  *
  * @category Smile
  * @package  Smile\ElasticsuiteCatalogRule
- * @author   Romain Ruaud <romain.ruaud@smile.fr>
  */
-class IsInStock implements SpecialAttributeInterface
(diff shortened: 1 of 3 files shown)


romainruaud:
Hi @vseager 

Please see my comments. I don't really understand the point of doing this like that.

Your "is saleable" is equivalent to our "only in stock products" => we should keep our legacy as is.

I dont think "stock.qty > 0 and stock.is_in_stock=true" should be a dedicated special attribute.

I'd rather go with introducing a new special attribute based only on stock.qty.

However, it will not work for configurable products as their stock.qty is always 0 (thanks to Magento for this).

vseager:
@romainruaud Thanks for your comments.

> Your "is saleable" is equivalent to our "only in stock products" => we should keep our legacy as is.

This is correct, it was really down to the terminology being understandable by the user. as "In Stock Products" to most administrators (not programmers) would imply the item is physically in stock, on the shelf (and doesn't require backorder). 
Therefore, "Only in stock products" = products that are in stock, on the shelf, ready to ship, and "Only saleable products" = available to purchase but not necessarily ready to ship yet (for example if backorders are enabled).

> I dont think "stock.qty > 0 and stock.is_in_stock=true" should be a dedicated special attribute.
> I'd rather go with introducing a new special attribute based only on stock.qty.

A stock.qty option is of course a perfectly good alternative solution.

Nuranto:
Be careful, a product with `stock.qty > 0 and stock.is_in_stock=true` can be backorderable, if `min_sale_qty` or `qty_increments is greater than 1` and `stock.qty = 1` for example. 
With this in mind, that special attribute idea could be legitimate.

PR #2566, opened 2022-05-13
Made the searchbar instantiatable multiple times

cartok:
Also rughly cleaned up the template 😉

PR #2669, opened 2022-08-01
Fix #2668 Mobile search bar is not opening on click

### LINKED ISSUE #2668: Mobile search bar is not opening on click
The bug was introduced with https://github.com/Smile-SA/elasticsuite/pull/2566

cartok:
* Generate an id also for the input field and use it to link label and input again


rbayet:
Mentionned as ok in https://github.com/Smile-SA/elasticsuite/pull/2566

Issue #2663, opened 2022-07-28
Magento 2.4.5 compatibility.


romainruaud:
Magento 2.4.5 will land on August 9th.

It's already added in our integration test suite for branch 2.10.x.

All tests are working actually which is a good start.

We need to check that all features are functioning as expected.

vahonc:
When clicking on the `Magento BO -> ElasticSuite -> Search Relevance` section we get an error:

<pre>
[2022-08-03T13:47:34.670004+00:00] report.CRITICAL: Error: Call to a member function getEuCountryList() on null in /home/vahonc/Projects/m245ee/vendor/magento/module-config/view/adminhtml/templates/system/config/js.phtml:81
Stack trace:
#0 /home/vahonc/Projects/m245ee/vendor/magento/framework/View/TemplateEngine/Php.php(71): include()
#1 /home/vahonc/Projects/m245ee/vendor/magento/framework/View/Element/Template.php(263): Magento\Framework\View\TemplateEngine\Php->render()
#2 /home/vahonc/Projects/m245ee/vendor/magento/framework/View/Element/Template.php(293): Magento\Framework\View\Element\Template->fetchView()
#3 /home/vahonc/Projects/m245ee/vendor/magento/module-backend/Block/Template.php(141): Magento\Framework\View\Element\Template->_toHtml()
#4 /home/vahonc/Projects/m245ee/vendor/magento/framework/View/Element/AbstractBlock.php(1095): Magento\Backend\Block\Template->_toHtml()
#5 /home/vahonc/Projects/m245ee/vendor/magento/framework/View/Element/AbstractBlock.php(1099): Magento\Framework\View\Element\AbstractBlock->Magento\Framework\View\Element\{closure}()
#6 /home/vahonc/Projects/m245ee/vendor/magento/framework/View/Element/AbstractBlock.php(660): Magento\Framework\View\Element\AbstractBlock->_loadCache()
#7 /home/vahonc/Projects/m245ee/vendor/magento/framework/View/Layout.php(578): Magento\Framework\View\Element\AbstractBlock->toHtml()
#8 /home/vahonc/Projects/m245ee/vendor/magento/framework/View/Layout.php(555): Magento\Framework\View\Layout->_renderBlock()
#9 /home/vahonc/Projects/m245ee/vendor/magento/framework/View/Layout.php(510): Magento\Framework\View\Layout->renderNonCachedElement()
[...]

PR #2674, opened 2022-08-03
Magento 2.4.5 (#2663): Fixing call to a member function getEuCountryList() on null

vahonc:


romainruaud:
Hi Vadym, is this still backward-compatible with lower Magento versions ? 2.4.3/2.4.4 at least, I don't know if "Magento\Config\ViewModel\CountryFilter" exist in core Magento when using those versions.

vahonc:
Yes, it looks like "Magento\Config\ViewModel\CountryFilter" doesn't exist in the previous version of Magento and it was released in M2.4.5. But I tried adding my fixes to M2.4.3 and didn't notice any issues, at least visible on the page, Search Relevance displaying as expected.

walterbak:
In vendor/smile/elasticsuite/src/module-elasticsuite-core/view/adminhtml/layout/smile_elasticsuite_search_request_relevanceconfig_edit.xml, the block (without name and alias!) that instantiate Magento_Config::system/config/js.phtml 

`<block class="Magento\Backend\Block\Template" template="Magento_Config::system/config/js.phtml"/>`

must declare as argument 

```
<arguments>
        <argument name="countryFilter" xsi:type="object">Magento\Config\ViewModel\CountryFilter</argument>
</arguments>
```

If block had a name (or just an alias) it may be extended from an external module... but an anonymous block as it, cannot be extended out of the module.

walterbak:
If you just add a name to the block
```
<block class="Magento\Backend\Block\Template" template="Magento_Config::system/config/js.phtml"/>
```
, we can hook it and add arguments from a custom module.

romainruaud:
If I report this fix on a Magento 2.4.4 instance, I have a fatal error on Search Relevance page : 

```
Exception #0 (ReflectionException): Class "Magento\Config\ViewModel\CountryFilter" does not exist
```

I'll try to check how to handle this more cleanly.

romainruaud:
Replaced by #2702

PR #2677, opened 2022-08-04
IsDiscount for configurable and grouped products with optimized performance

Bashev:
This pull request is part of #2237 closed issue, because of bad performance.

Resolve https://github.com/Smile-SA/elasticsuite/issues/866


Bashev:
Performance is optimized, use data directly from db instead of using productRepository

Bashev:
@romainruaud can you reviewing this PR?

composer.json
```diff
@@ -45,7 +45,8 @@
         "magento/module-inventory-sales": ">=1.1.0",
         "magento/module-inventory-indexer": ">=2.0.0",
         "magento/magento-composer-installer": "*",
-        "elasticsearch/elasticsearch": "~6.7|~7.2"
+        "elasticsearch/elasticsearch": "~6.7|~7.2",
+        "smile/magento2-smilelab-phpcs": "^2.0"
```
romainruaud, line 49: This should not be in "require" part of the composer.json but remains in require-dev.

src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/PriceData.php
```diff
@@ -65,10 +76,26 @@ public function addData($storeId, array $indexData)
             $originalPrice = $priceModifier->getOriginalPrice($priceDataRow);
             $price         = $priceModifier->getPrice($priceDataRow);
 
+            $isDiscount    = $price < $originalPrice;
+            if (in_array($productTypeId, $this->attributeResourceModel->getCompositeTypes())) {
+                $isDiscount = false;
+                $priceModifier = $this->getPriceDataReader('default');
+                $allChildrenIds = $this->attributeResourceModel->loadChildrens([$productId], $storeId);
```
romainruaud, line 83: This causes a load query per each configurable product in the batch.

It's possible to load all of them at once, like in https://github.com/Smile-SA/elasticsuite/blob/2.10.x/src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/AttributeData.php#L70

Maybe even better : at this step, we might already have the "children_ids" key inside the product data (that's true if the AttributeData datasource is run before the Price, which I'm not sure about) : 

https://github.com/Smile-SA/elasticsuite/blob/2.10.x/src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/AttributeData.php#L87

src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/PriceData.php
```diff
@@ -65,10 +76,26 @@ public function addData($storeId, array $indexData)
...
+                $childPriceData = $this->resourceModel->loadPriceData($storeId, array_keys($allChildrenIds));
```
romainruaud, line 84: same as above, you are doing a query for each "batch" of chil products, that could probably be done with an unique query pre-emptively.

romainruaud:
Hi @Bashev 

I merged this into 2.10.x branch.

Please take into account that I plugged this with a configuration variable, so that only people needing it will be able to enable it.

Thank you for contributing this !

Issue #2709, opened 2022-09-13
Cluster mis-configuration warning for shards


romainruaud:
**User story :** 

I'd like to have a warning displayed in the Elasticsuite Back-Office when the following event is matched : 

- people configured a number of shards > 1
- they do not have any indices that are > 10Go

If that's the case, a warning message should be displayed on top of the Back-Office to indicate them that their configuration is incorrect.

**Technical implementation :**

To display a message in the back-office, you have to inject it like this : https://github.com/Smile-SA/elasticsuite/blob/2.10.x/src/module-elasticsuite-core/etc/adminhtml/di.xml#L21

And you have to implement a message class like this : https://github.com/Smile-SA/elasticsuite/blob/2.10.x/src/module-elasticsuite-core/Model/System/Message/NotificationAboutVersions.php

For the ```isDisplayed()``` it should return ```true``` if : 

- the number_of_shards parameter is > 1 (if not, don't do the second test). https://github.com/Smile-SA/elasticsuite/blob/2.10.x/src/module-elasticsuite-core/etc/config.xml#L31
- AND there is no index having a size > 10 Gb

To get the size of all indices, you can have a look in the "module-elasticsuite-indices" that is displaying the indices grid in Back-Office. You should find in this module some utility methods that are able to fetch indices sizes.

For the ```getIdentity()``` method we should generate an ID that is unique.

For the ```getSeverity()``` method this should be a warning.

For the ```getText()``` method we should return the following message : 

"The number of shards configured for Elasticsuite is incorrect. You do not need to use $X shards since your biggest Elasticsuite index is only $Y. Please **modify your configuration** (and here we put a link to Elasticsuite Config) according to our **wiki page** (and here we will put a link to a page in our wiki where we detail how many shards to put)."

vahonc:
The last thing left here is to add a translation of the text message into French and maybe German.

I changed the text a little, I think so it will be more understandable for everyone.

![Screenshot from 2022-09-27 11-31-41](https://user-images.githubusercontent.com/67791361/192476951-6c4f94d9-5776-440f-939c-d8d3871ff1fc.png)

So, here's the text message:
<pre>
The number of shards configured for Elasticsuite is incorrect. You do not need to use $X shards since your biggest Elasticsuite index is only $Y.
Click here to go to the Elasticsuite Config page and change your Number of Shards per Index parameter according to our wiki page.
</pre>

romainruaud:
I added the translations, I close this as completed

PR #2720, opened 2022-09-27
#2709, Adding cluster misconfiguration warning for shards

vahonc:


src/module-elasticsuite-core/Model/System/Message/WarningAboutClusterShardsMisconfig.php
```diff
@@ -0,0 +1,169 @@
...
+        $this->urlBuilder         = $urlBuilder;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isDisplayed()
+    {
+        $numberOfShards = $this->helper->getNumberOfShards();
+        $indexMaxSize = $this->getElasticsuiteIndexMaxSize()['size_in_bytes'];
```
romainruaud, line 83: You could move this inside the ```if ($numberOfShards > 1)```, by doing so, we would only retrieve the indices collection when it's needed, so this will be more performant.

src/module-elasticsuite-core/Model/System/Message/WarningAboutClusterShardsMisconfig.php
```diff
@@ -0,0 +1,169 @@
...
+
+        foreach ($elasticsuiteIndices as $indexName => $indexAlias) {
+            $indexData = $this->indexStatsProvider->indexStats($indexName, $indexAlias);
+            $indexSizes[] = ['human_size' => $indexData['size'], 'size_in_bytes' => $indexData['size_in_bytes']];
+        }
+
+        $indexSizesInBytes = array_column($indexSizes, "size_in_bytes");
+        array_multisort($indexSizesInBytes, SORT_DESC, $indexSizes);
+
+        return current($indexSizes);
```
romainruaud, line 154: How does it work if $indexSizes is empty ?

src/module-elasticsuite-core/Model/System/Message/WarningAboutClusterShardsMisconfig.php
romainruaud: ```['size_in_bytes']``` could be undefined if there are no indices actually. You should add some guard functions to check that ```getElasticsuiteIndexMaxSize()``` is returning something, and if the result has this key.

romainruaud (changes requested):

src/module-elasticsuite-core/Model/System/Message/WarningAboutClusterShardsMisconfig.php
vahonc: Fixed.

src/module-elasticsuite-core/Model/System/Message/WarningAboutClusterShardsMisconfig.php
vahonc: Fixed.

romainruaud:
tested ok, even after manually deleting all ES indices.

I'll merge this as soon as the build finishes.

Good job @vahonc !

PR #2742, opened 2022-10-13
Allow SKU field in rules to use both "contains" and "is one of"

romainruaud:


sheepfy:
@romainruaud check this file module-elasticsuite-catalog-rule/Model/Rule/Condition/Product.php
     public function getDefaultOperatorInputByType()
     {
-        if (null === $this->_defaultOperatorInputByType) {
-            $this->_defaultOperatorInputByType = [
-                'string'      => ['{}', '!{}'],
-                'numeric'     => ['==', '!=', '>=', '>', '<=', '<'],
-                'date'        => ['==', '>=', '>', '<=', '<'],
-                'select'      => ['==', '!='],
-                'boolean'     => ['==', '!='],
-                'multiselect' => ['()', '!()'],
-                'grid'        => ['()', '!()'],
-                'category'    => ['()', '!()'],
-            ];
-            $this->_arrayInputTypes            = ['multiselect', 'grid', 'category'];
-        }

The PR will only allow "is one of" or "is not one of". But for string, we should not use "is" ? (operator ==)

romainruaud:
SKU is chosen through a grid, so for me it's rather a multiselect than a string.

In any case, they are always treated this way (kinda hardcoded) with a "match" query : https://github.com/Smile-SA/elasticsuite/blob/2.10.x/src/module-elasticsuite-catalog-rule/Model/Rule/Condition/Product/QueryBuilder.php#L139

So changing the operator only changes one thing, here : https://github.com/Smile-SA/elasticsuite/blob/2.10.x/src/module-elasticsuite-catalog-rule/Model/Rule/Condition/Product/QueryBuilder.php#L290

When using "contains", we use the default analyzer of the field, which means, for SKU, we use "sku.reference".

When using "something else", we use the "untouched" analyzer (so the non-altered version).

That's why you did spot the difference when the operators were modified.

Maybe we could allow both "contains" and "is one of" for SKU specifically, to allow doing an exact match if wanted, or a more approximative matching.

Regards

sheepfy:
@romainruaud Yes, agree, using both is fine, all good from me

romainruaud:
So I updated this PR @sheepfy 

Can you check if that's doing what you want ?

Regards

Issue #2733, opened 2022-10-11
Incorrect results because of indexed child SKUs in GraphQL Query


lewisvoncken:
### Preconditions

Magento Version : 2.4.5

ElasticSuite Version : 2.10.11

Environment : Production

### Steps to reproduce

1. Create a bundle product with simple products that are visible on Catalog, Search
2. Reindex
3. execute graphql to retrieve the simple product based on the SKU

variables:
```json
filters: {
            "sku": { "eq": "childskuhere" },
        }
```
```graphql
   query GetProducts($filters: ProductAttributeFilterInput!) {
            products(filter: $filters) {
                items {
                    sku
                }
            }
        }
```

### Expected result

1. if you request the child sku it should return the actual child.

### Actual result

1. The bundle product is returned

#### Add child_skus to mapping
https://github.com/Smile-SA/elasticsuite/blob/6f1de645d7b4cd8f8a5551d293f953c9adc8c352/src/module-elasticsuite-catalog/etc/elasticsuite_indices.xml#L32

```xml
                <field name="child_skus" type="text">
                    <isSearchable>1</isSearchable>
                    <isUsedForSortBy>1</isUsedForSortBy>
                    <isUsedInSpellcheck>1</isUsedInSpellcheck>
                    <isFilterable>1</isFilterable>
                    <defaultSearchAnalyzer>reference</defaultSearchAnalyzer>
                </field>
```

#### Change function to use child_skus
https://github.com/Smile-SA/elasticsuite/blob/e6ac45c2112ec0a99b2373c85b1be97ad9b62bbc/src/module-elasticsuite-catalog/Model/Product/Indexer/Fulltext/Datasource/AttributeData.php#L226
```php
    private function addChildSku(&$parentData, $relation)
    {
        if (isset($parentData['sku']) && !isset($parentData['child_sku'])) {
            $parentData['child_sku'] = [$parentData['sku']];
        }

        $parentData['child_sku'][] = $relation['sku'];
        $parentData['child_sku'] = array_unique($parentData['child_sku']);
    }
```

romainruaud:
@vahonc as we discussed, can you implement this solution but with a way to allow enabling/disabling it ?

You can add a system config parameter somewhere in **Stores > Configuration > Elasticsuite > Catalog Search**

This parameter should be named "Index child product SKU in dedicated subfield". It's a boolean that defaults to "0" (in config.xml).

[...]

PR #2753, opened 2022-10-21
#2733, Fix indexing of child products SKUs

vahonc:
Probably need to add text translation for a new setting in Magento BO, at least for French.


romainruaud:
Hi @vahonc there is still a PHPMD issue.

It's due to the usage of "else" which is not recommended, but in this case, I think we can ignore the warning by added a PHPMD statement. You can find how to do it by searching for "ElseExpression" in the source code.

PR #2761, opened 2022-10-24
Added some isset checks for array keys being used in GraphQl

### LINKED ISSUE #2760: GraphQL - Undefined array key "eq" on category page
After upgrade from 2.10.11 to 2.10.12 (also present in 2.10.12.1) whenever I open a category listing page the GraphQL query returns an Internal Server Error

### Preconditions

* No preconditions other than having a category with products and navigating to it via the Venia app. 
* All indexes up to date.

Magento Version: 2.4.5-p1 EE
PWA Studio Verison: 12.6
ElasticSuite Version: 2.10.12
Environment: Production 
Third party modules: None

### Steps to reproduce

1. Navigate to a category page, which triggers the following GraphQL query:

```graphql
query getCategoryAvailableSortMethods(
    $categoryIdFilter: FilterEqualTypeInput!
) {
    products(filter: { category_uid: $categoryIdFilter }) {
        sort_fields {
            options {
                label
                value
            }
        }
    }
}
```

With the default Venia variables:

```json
{
    "categoryIdFilter": {
        "in": "<your category UID>"
    }
}
```

Source: https://github.com/magento/pwa-studio/blob/v12.6.0/packages/peregrine/lib/talons/RootComponents/Category/useCategoryContent.js#L76-L86

### Expected result

```json
{
    "data": {
        "products": {
            "sort_fields": {
                "options": [
                    {
                        "label": "Position",
                        "value": "position"
                    },
                    {
                        "label": "Product Name",
                        "value": "name"
                    },
                    {
                        "label": "Price",
                        "value": "price"
                    }
                ]
            }
        }
    }
}
```

### Actual result

```json
{
    "errors": [
        {
            "message": "Internal server error",
            "extensions": {
                "category": "internal"
            },
            "locations": [
                {
                    "line": 4,
                    "column": 9
                }
            ],
            "path": [
                "products"
            ]
        }
    ],
    "data": {
        "products": null
    }
}
```

Also the following is written to `system.log`

```
report.ERROR: Warning: Undefined array key "eq" in /app/vendor/smile/elasticsuite/src/module-elasticsuite-catalog-graph-ql/Model/Resolver/Products/ContextUpdater.php on line 89
```

### Additional Notes

The problem seems to be this commit https://github.com/Smile-SA/elasticsuite/commit/146f3be8965e1f4722d7537214b0c2a0ad711cd6#diff-488118360edec55ff92bb44921725156684c926165adc4d7ce9922ab2f263ba0L76-R88 which removed the check for `$args['filter']['category_id']['eq']` that previously existed in line 76.

barbazul:
Fixes #2760 by adding some defensive isset() checks on the array keys being accessed.

### DIFF
--- a/src/module-elasticsuite-catalog-graph-ql/Model/Resolver/Products/ContextUpdater.php
+++ b/src/module-elasticsuite-catalog-graph-ql/Model/Resolver/Products/ContextUpdater.php
@@ -85,7 +85,7 @@ class ContextUpdater
 
             $this->context->setCurrentSearchQuery($query);
         } elseif (!empty($args['filter']) && (!empty($args['filter']['category_id']) || !empty($args['filter']['category_uid']))) {
-            if (isset($args['filter']['category_uid'])) {
+            if (isset($args['filter']['category_uid']) && isset($args['filter']['category_uid']['eq'])) {
                 $categoryUid = $this->uidEncoder->decode($args['filter']['category_uid']['eq']);
             }
             $categoryId = $args['filter']['category_id']['eq'] ?? $categoryUid ?? false;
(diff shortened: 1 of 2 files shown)


barbazul:
I wonder why the other checks are not running

romainruaud:
Just because I need to flag the PR as "safe to check", otherwise, you could introduce malicious code in your PR, and the tests might execute them.

That's a security flaw in public repos :/

barbazul:
Nice! Thanks for the reply :)

PR #2774, opened 2022-11-09
Fix #2772 by removing magento inventory plugin.

### LINKED ISSUE #2772: 2.4.5-p1 Search Slow - show_out_of_stock enabled
When Show Out Of Stock products in enabled, search with ElasticSuite is very slow

### Preconditions
Magento Version : CE 2.4.5-p1
ElasticSuite Version : 2.10.12.1
Environment : Production

### Steps to reproduce

1. Configuration>Catalog>Inventory>Stock Options>Display Out Of Stock Products: YES
2. Slow search on front end

Same issue occurs using default ElasticSearch 7, the speed issue can be solved with this fix:
https://github.com/magento/magento2/issues/36375

With ElasticSuite enabled, the above fix does not affect the load speed and it is still extremely slow.

romainruaud:


Jarzebowsky:
When it will be released?

romainruaud:
Very soon, I drafted the release this morning. So probably at the beginning of next week.

If you cannot wait, this fix is just an XML to copy-paste on your project :) 

Regards

Jarzebowsky:
Yeap, we are doing a composer patch for it instead. Looking forward to upgrade the module itself asap :) Thank you.

PR #2791, opened 2022-11-28
Update Decimal.php

MTheProgrammer:
Hello,

we have a custom range type filter that is based on decimal attribute. An exception is thrown when passed value has no minimal limit, e.g:

error: `https://magento.domain/exapmple-category.html?depth=-281`
ok: `https://magento.domain/exapmple-category.html?depth=0-281`

Everything works perfectly for the price filter, but for decimal attribute filter there is an error:

Exception #0 (BadMethodCallException): Missing required argument $values of Smile\ElasticsuiteCore\Search\Request\Query\Terms.

Exception is raised by factory on the line 153 in `Smile\ElasticsuiteCore\Search\Request\Query\Filter\QueryBuilder` as $condition has no `values` key:
`$query = $this->queryFactory->create($queryType, $condition);`

The PHP `current` function returns false when array is empty. Condition `is_infinite` in the loop evaluates to false and validation passes with incorrect filter values. Decimal attribute properties are slightly different than those found in prices. We can't use 0 to limit the minimal value because there are SI units that might have negative values: sleeping bag's comfortable temperature, electronic component's voltage. We can safely assume that the range filter is invalid without both values.


romainruaud:
Cherry picked into 2.10.x branch, thank you for contributing !

Regards

Issue #2763, opened 2022-10-25
Analyzers screens


romainruaud:
This screen is a new Back-Office screen, so you'll probably need custom controllers, layout, blocks, etc... to be created from scratch.

It must be under the **Elasticsuite > System** menu, just below "Indices".

The code must be in the "Elasticsuite_Indices" module.

The goal is to replicate a part of the "Analysis" tool of cerebro.

![image](https://user-images.githubusercontent.com/15340849/197717432-fcb51fbd-d67c-4bdb-885f-99c4fe596408.png)

**Index name** : 

- a dropdown which is filled with all the index of the platform. "ghost", "external" and "undefined" indices should not be added to this select.

**Analyzer name** : 

- a dropdown which is filled with all the analyzers that are defined for the currently selected index.
- therefore, this dropdown is empty until we select an index.
- the analyzers can be fetched from the index settings that we already have in the Elasticsuite_Indices module.

Probably, you could fetch all the indices and their associated analyzers in a single query. Maybe encode them in JSON to use them in the form ?

**Text to analyze** : 

- this is just a text field where the user can submit a term to check how it's analyzed by the chosen analyzer.

**Analyze button** : 

- this is a button that will trigger the analyze query to the Elasticsearch engine.

The analyze query to use is this API : https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-analyze.html

We have an entry point for this here : https://github.com/Smile-SA/elasticsuite/blob/2.10.x/src/module-elasticsuite-core/Client/Client.php#L205

In the call we must send : 
- the index name
- the chosen analyzer
- the text to analyze

The call has to be made in Ajax (so a custom controller is needed).

**Results** : 

This is a table that will render the results of the call to "analyze" : 

- token
- position	
- start offset
- end offset

We just have to format them appropriately.

romainruaud:
This was merged, I close.

PR #2794, opened 2022-12-08
#2763 Add analyzer screen to Elasticsuite_Indices module

vahonc:


src/module-elasticsuite-indices/Controller/Adminhtml/Analysis/Request.php
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function execute()
+    {
+        $errors = [];
+
+        if ($_POST['index'] == '0') {
```
romainruaud, line 66: $_POST['index'] should be replaced by $this->getRequest()->getPost($index, null);

Same for other posted variables.

src/module-elasticsuite-indices/Block/Adminhtml/Analysis/Analyzer.php
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Adminhtml Analysis by Analyzer Block.
+ *
+ * @category Smile
+ * @package  Smile\ElasticsuiteIndices
+ * @author   Vadym HONCHARUK <vahonc@smile.fr>
+ */
+class Analyzer extends Template
+{
+    public const GHOST_STATUS = 'ghost';
```
romainruaud, line 30: I think those constants exists elsewhere in the same module, no ?

vahonc: Yes, you're right. I fixed this too.

Issue #2810, opened 2023-01-03
Condition "category is one of .." accounts only for first category on the list


pomppa:
There is an issue with conditon "category is one of", as this condition accounts only for the first category on the list.

### Preconditions
Magento Version : 2.4.5-p1
ElasticSuite Version : 2.10.12.1
Environment : Developer / Production
Third party modules : None

### Steps to reproduce
1. Navigate to Magento admin, choose some page (e.g. 404 no-route)
2. Open content with PageBuilder, add content "Products"
3. Add "Select Products By" -> "Condition"
4. Create condition "Category is one of" and add two categories
5. Sort e.g. with price from highest to lowest for easier validation
6. Save page and view page on FE

### Expected result

1. Products should be displayed from both categories selected on the condition

### Actual result
1. Products are displayed only from the first category selected on the condition

romainruaud:
Inject this class : https://github.com/Smile-SA/elasticsuite/blob/2.10.x/src/module-elasticsuite-catalog/Model/Category/Filter/Provider.php

Inject also this : https://github.com/Smile-SA/elasticsuite/blob/2.10.x/src/module-elasticsuite-core/Search/Request/Query/QueryFactory.php

Check if `$value` contains an array or comma-separated value.

Loop on the array.

For each category inside, load it, and retrieve a query filter from the `Filter\Provider`.

Assemble all the query under a "boolean should" query : 

```
$queries = $this->queryFactory->create(QueryInterface::TYPE_BOOL, [ 'should' => $queries ]);
```

And apply the resulting query to the collection with `$collection->addQueryFilter($query)`

PR #2820, opened 2023-01-16
#2810, Fix using 'category_ids' as condtion for virtual category rule on widget collection

vahonc:


src/module-elasticsuite-virtual-category/Plugin/Widget/ProductsListPlugin.php
```diff
@@ -136,15 +155,31 @@ public function afterCreateCollection(ProductsList $subject, $collection)
                 $conditions = $this->conditionsHelper->decode($conditions);
                 foreach ($conditions as $condition) {
                     if (!empty($condition['attribute'])) {
-                        if ($condition['attribute'] == 'category_ids') {
-                            if (array_key_exists('value', $condition)) {
-                                $categoryId = $condition['value'];
-                                try {
+                        if ($condition['attribute'] == 'category_ids' && array_key_exists('value', $condition)) {
```
romainruaud, line 158: I think we could do something more simple.

We should be handle to handle $condition['value'] in the same manner, no matter if it contains "1" or "1,2", the explode will always return an array, no ?

I have something like this in my mind : 

```php
  if ($condition['attribute'] == 'category_ids' && array_key_exists('value', $condition)) {
          $filterQueries = [];
          
          // This should work even when $condition['value'] contains only one value, no ?

          $categoryIds = array_map("trim", explode(',', (string) $condition['value']));
          foreach ($categoryIds as $categoryId) {
              try {
                  $category = $this->categoryRepository->get($categoryId, $storeId);
                  $filterQueries[] = $this->filterProvider->getQueryFilter($category);
              } catch (NoSuchEntityException $exception) {
                  continue;
              }
          }

          if (!empty($filterQueries) {
              $query = $this->queryFactory->create(QueryInterface::TYPE_BOOL, ['should' => $filterQueries]);
              $collection->addQueryFilter($query);
          }
          continue;
      }
  }
```

vahonc: Yes, you're right. I refactored the code.

PR #2838, opened 2023-02-06
Fix #2835, avoiding deprecated functionality in PHP 8.2

### LINKED ISSUE #2835: Deprecated Functionality: Creation of dynamic property
Hello here,

### Preconditions

I am testing this module with PHP8.2 and Magento 2.4.6-beta2 before it will be released. 

Magento Version :
2.4.6-beta2

ElasticSuite Version :
2.10.13.1

Environment :
Developer

### Steps to reproduce

1. Install Magento 2.4.6-beta2 with PHP 8.2
2. Run bin/magento setup:upgrade

### Expected result
command setup:upgrade runs without exceptions

### Actual result
The directory '/var/www/html/generated/metadata/' doesn't exist - skipping cleanup
Updating modules:
Deprecated Functionality: Creation of dynamic property Smile\ElasticsuiteCore\Setup\Validator::$client is deprecated in /var/www/html/vendor/smile/elasticsuite/src/module-elasticsuite-core/Setup/Validator.php on line 37

vahonc:


src/module-elasticsuite-indices/Model/ResourceModel/Index/Collection.php
```diff
@@ -59,7 +59,7 @@ public function getItemsByColumnValue($column, $value)
 
         $res = [];
         foreach ($this as $item) {
-            if (strpos($item->getData($column), $value) !== false) {
+            if (str_contains($item->getData($column), $value)) {
```
romainruaud, line 62: What was the problem with `strpos` ? 

`str_contains` exists only with PHP > 8.0.0 which would break BC with PHP 7.4

vahonc: This was changed due to the error in the exception.log: `Deprecated Functionality: strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated in ../smile/elasticsuite/src/module-elasticsuite-indices/Model/ResourceModel/Index/Collection.php on line 62` and also according to the advice of PHPStorm `strpos()' call can be converted to 'str_contains()'`. But I forgot that `str_contains` exists only with PHP > 8.x

vahonc: Fixed.

PR #2910, opened 2023-04-14
Allow to deal with elasticsearch/elasticsearch-php v8.

romainruaud:
- Different namespaces to deal with
- Different exceptions to deal with
- How to reproduce the old "parallel/async" mode ?
- ~The logger with client v8 is not as verbose as previously, how to improve ?~ it's just that now the content of queries is in debug.log instead of system.log. Impossible to change :/

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,9 @@
         "magento/module-inventory-sales": ">=1.1.0",
         "magento/module-inventory-indexer": ">=2.0.0",
         "magento/magento-composer-installer": "*",
-        "elasticsearch/elasticsearch": "~7.17.0 || ~8.5.0"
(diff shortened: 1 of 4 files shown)


romainruaud:
I went into something way too complicated for this one I guess and I'm not a huge fan of having a ClientBuilder that can generate 3 different objects.

We could probably rely "only" on the OpenSearch client, which is integrated by default on Magento now, to communicate with ES7, ES8, OS1, OS2.

To avoid misunderstanding by the endusers, we should maybe prevent installing the ES8 PHP client, or the magento/module-elaticsearch-8 package.

This would probably be the cleanest option, at least for now, until we cannot manage all flavors and versions with an unique PHP client.....

romainruaud:
Replaced by #3131

Issue #2097, opened 2021-03-18
Slow Search Queries - Possibly related to virtual categories


harrigo:
I have been using this module for a few years now and have found it relatively quick however over time have noticed a bit of a slow down. I have been playing with this on lesser hardware recently and found queries to be quite slow... here is a section of my profiler:

```

magento | 1.741605 | 1.741605 | 1 | 28,880,656 | 48,234,496
-- | -- | -- | -- | -- | --
...
·  ·  ·  ·  ·  ·  ·  ·  TEMPLATE:/var/www/buzzdevb/public_html/var/view_preprocessed/pub/static/vendor/magento/module-catalog-search/view/frontend/templates/result.phtml | 1.354794 | 1.354794 | 1 | 5,564,800 | 27,262,976
·  ·  ·  ·  ·  ·  ·  ·  ·  EAV: Magento\Eav\Model\Config::getAttribute | 0.001753 | 0.000008 | 212 | 9,776 | 0
·  ·  ·  ·  ·  ·  ·  ·  ·  ·  EAV: Magento\Eav\Model\Config::getEntityType | 0.000604 | 0.000003 | 212 | 11,000 | 0
·  ·  ·  ·  ·  ·  ·  ·  ·  EVENT:catalog_block_product_list_collection | 0.000271 | 0.000271 | 1 | 85,608 | 0
·  ·  ·  ·  ·  ·  ·  ·  ·  ·  OBSERVER:review | 0.000255 | 0.000255 | 1 | 83,800 | 0
·  ·  ·  ·  ·  ·  ·  ·  ·  EAV:load_collection | 1.241051 | 1.241051 | 1 | 2,616,152 | 27,262,976
·  ·  ·  ·  ·  ·  ·  ·  ·  ·  ES:Execute Search Query | 1.107998 | 0.276999 | 4 | 1,628,528 | 10,485,760
·  ·  ·  ·  ·  ·  ·  ·  ·  ·  ES:Virtual Rule getSearchQueriesByChildren | 0.076594 | 0.038297 | 2 | 20,891,272 | 14,680,064
```

So the whole page is taking around 1.74 seconds which is pretty slow for my site with uncached product and category pages taking like 0,5-0,7s TTFB. The main issue I am having is certain search queries. My store has about 20,000 products and feel that should be relatively small but searching plates can take a while as have a few 1000 of them and I feel this shouldn't be the case.

[...]

PR #2926, opened 2023-05-01
#2097 Virtual Categories - Slow Search Queries fix

Nuranto:
see #2097


rbayet:
See also https://github.com/Smile-SA/elasticsuite/pull/3193 (which added cache AROUND)

Nuranto:
Seems to be a duplicate of https://github.com/Smile-SA/elasticsuite/pull/3326 (which is far more elegant than my solution)

PR #2935, opened 2023-05-14
Fix Unit Test Command issue in core module

maheshmakwana-code:
When I ran the unit test command for a core module I got the below error : 
Screenshot from 2023-05-14 17-20-13 (1)

I  fixed the above issue by updating class name :
Screenshot from 2023-05-14 17-20-13 (2) (1)


romainruaud:
Replaced by #2939 : proper fix is to rename the file.

Regards

Issue #3026, opened 2023-08-14
Cannot use default value of facet_sort_order in category filterable attribute config


Nuranto:
### Preconditions

Magento Version : 2.4.6-p2

ElasticSuite Version : 2.11.3.2

Environment :

Third party modules :

### Steps to reproduce

1. Go to a backend category page, check "Use default value" for Facet sort order of any attribute
2. Save category

### Expected result

1. After page gets reloaded, the checkbox is still checked.

### Actual result

1.  After page gets reloaded, the checkbox is no longer checked, so next time it will be saved, the facet sort order value will change for first value.

This is probably because facet_sort_order column cannot be null.

vahonc:
Hello @Nuranto,

Thanks for your fix. 

I only have one suggestion, please also remove `default="_count"` in your PR because it shouldn't be there either. It's my fault.

BR,
Vadym

Nuranto:
You're right, I missed it.

PR #3027, opened 2023-08-14
#3026 - Fix "Cannot use default value of facet_sort_order in category filterable attribute config

Nuranto:
see #3026


rbayet:
Refactored as single commit on top of 2.11.x in #3043

rbayet:
Refactored PR #3043 was merged

PR #3028, opened 2023-08-14
Smile_ElasticsuiteTracker: Only Add Frontend Tracker Blocks If Config Enabled

pykettk:
# Description
This is a commonly disabled feature so it would be nice to use `ifconfig` layout XML property to optionally include frontend `Smile_ElasticsuiteTracker` blocks.

# Reviewer Notes
I was torn between using `smile_elasticsuite_tracker/general/enabled` and `smile_elasticsuite_telemetry/telemetry/enabled`. If there's a preference for the latter then I'm happy to edit these changes to reflect that.

### DIFF
--- a/src/module-elasticsuite-tracker/view/frontend/layout/catalogsearch_result_index.xml
+++ b/src/module-elasticsuite-tracker/view/frontend/layout/catalogsearch_result_index.xml
@@ -16,10 +16,11 @@
  */
 -->
 <page xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:View/Layout/etc/page_configuration.xsd">
-
     <body>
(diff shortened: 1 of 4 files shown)


rbayet:
Hello @pykettk,

`smile_elasticsuite_tracker/general/enabled` makes more sense since it's the legacy feature and the telemetry is just piggy-backing on it.
Besides, the condition inside the `Smile\ElasticsuiteTracker\Block\Config` template already checks `smile_elasticsuite_tracker/general/enabled`.

Regards,

rbayet:
Hello @pykettk,

Could you relocate your commit on a branch based on 2.10.x ?
That will make the feature available on both 2.10.x and 2.11.x after merge.

Regards,

rbayet:
Commit moved to PR #3051 where it's rebased on 2.10.x

Issue #3038, opened 2023-09-02
Make use of product attribute's display_pattern on product view page


ecrocombe:
Product attributes configured as a slider with a <code>display_pattern</code> display's correctly in layered navigation, but not on frontend product view page. 

Without displaying the unit of measurement for the attribute, the information displayed is pretty useless, this could be grams, ounces, the customer viewing the product wouldn't know.

Also, could we scope `display_pattern` to store_view? Since it could include 'pieces' or 'куски'.

### Sample image of layered navigation in category view on frontend displaying correctly.
![image](https://github.com/Smile-SA/elasticsuite/assets/42088209/2cb0e94d-f868-4234-83f8-b47fd3c85632)

### Sample image of the same attribute on the product view on frontend:
![image](https://github.com/Smile-SA/elasticsuite/assets/42088209/1722424e-4f4f-4bf5-9b78-681e6dcde931)

### Result of <code>select * from eav_attribute as e join catalog_eav_attribute as c on c.attribute_id = e.attribute_id where attribute_code = 'ica_features_maximumweightcapacity';</code>
+--------------+----------------+------------------------------------+-----------------+--------------------------------------------------------+--------------+---------------+----------------+----------------+-------------------------+-----------------+--------------+-------------+-----------------+---------------+-----------+------+--------------+-------------------------+-----------+------------+---------------+---------------+---------------+---------------------+--------------------------+-------------------------+-------------------------+-------------------------+------------------+----------+-------------------------------+----------+--------------------+-------------------------+----------------------------+-----------------+--------------------+-----------------------+---------------+------------------------+-----------------+------------------------------+-----------------------+-------------------------+----------------+------------------+---------------------+-----------------+-------------------+------------------------+-------------------------+-------------------------+---------------------------+--------------+----------------+------------------+
[...]

PR #3047, opened 2023-09-12
Feature #3038 Use display_pattern on product view, product compare pages

rbayet:
(disabled by default)


rbayet:
Hello @ecrocombe,

Actually, testing the feature locally, I think it needs a bit more polish : the precision is actually not taken into account at all.

It seems that the slider cheats by systematically using the JS price formatter because the locale oriented PHP classes of Magento2 Framework does not seem to really care about it (hence probably because you went for "round").
I've tried using the approach of `\Smile\ElasticsuiteCatalog\Model\Layer\Filter\Decimal::formatValue` but that does not work either.

So I'll try to figure out a way to correctly have the value formatted **exactly** like in the slider.

rbayet:
Replaced by #3062 so @vahonc can contribute to it

PR #3069, opened 2023-10-03
Fix #3061, fixing PHPStan level 0 errors

### LINKED ISSUE #3061: Fixing phpstan level 0 errors
On branch 2.10.x, eliminate all PHPStan level 0 errors which are basically inconsistancies between method signatures or PHPDoc and actual code.

# Running PHPStan
On a Magento+Elasticsuite 2.10.x installation, you might not have phpstan available in vendor/bin/phpstan, add it with 
<pre>
composer require --dev smile/magento2-smilelab-phpstan 1.0.3
</pre>
(this will install phpstan/phpstan as a dependency).

Copy the vendor/smile/elasticsuite/phpstan.neon.dist at the root of magento if there is none or if its content is different, 
then
<pre>
vendor/bin/phpstan analyze --level=0 vendor/smile/elasticsuite
</pre>

# Missing returns in the code
For instance 
<pre>
  Line   src/module-elasticsuite-catalog-optimizer/Controller/Adminhtml/Optimizer/Preview.php                                                                                                                
  117    Method Smile\ElasticsuiteCatalogOptimizer\Controller\Adminhtml\Optimizer\Preview::execute() should return Magento\Framework\App\ResponseInterface|Magento\Framework\Controller\ResultInterface but  
         return statement is missing.                                                                                                                                                                        
</pre>
is because the \Smile\ElasticsuiteCatalogOptimizer\Controller\Adminhtml\Optimizer\Preview::execute inherits its PHPdoc which states 
<pre>
    /**
     * Execute action based on request and return result
     *
     * @return \Magento\Framework\Controller\ResultInterface|ResponseInterface
     * @throws \Magento\Framework\Exception\NotFoundException
     */
</pre>

In this case, we should indeed add a `return` in front of `$this->getResponse()->representJson($json);` since `representJson` returns its parent `\Magento\Framework\App\Response\Http` object which is a `\Magento\Framework\Controller\ResultInterface|ResponseInterface`.

# Invalid PHPdoc
Sometimes, especially in UI components' dataproviders, the PHPdoc should be fixed to reflect the actual usage.
For instance 
<pre>
  Line   src/module-elasticsuite-catalog-optimizer/Ui/Component/Optimizer/Listing/DataProvider.php                                                                
  110    Method Smile\ElasticsuiteCatalogOptimizer\Ui\Component\Optimizer\Listing\DataProvider::addFilter() should return mixed but return statement is missing.  
</pre>
The \Magento\Framework\View\Element\UiComponent\DataProvider\DataProviderInterface PHPDoc says the method `addFilter` should return mixed 
<pre>
    /**
     * Add field filter to collection
     *
     * @param \Magento\Framework\Api\Filter $filter
     * @return mixed
     */
</pre>
when actually, the parent class of `Smile\ElasticsuiteCatalogOptimizer\Ui\Component\Optimizer\Listing\DataProvider`, does not either :
<pre>
    /**
     * @inheritdoc
     */
    public function addFilter(\Magento\Framework\Api\Filter $filter)
    {
        $this->getCollection()->addFieldToFilter(
            $filter->getField(),
            [$filter->getConditionType() => $filter->getValue()]
        );
    }
</pre>

So the fix in this case is altering the PHPDoc to silence the error : 
<pre>
    /**
     * {@inheritdoc}
     *
     * @return void
     */
    public function addFilter(\Magento\Framework\Api\Filter $filter)
    {
        if (isset($this->addFilterStrategies[$filter->getField()])) {
</pre>

vahonc:


rbayet:
Replaced by #3072

PR #3094, opened 2023-10-22
Spellchecker constructor settings unit tests

rbayet:


rbayet:
Replaced by #3093 which will be merged into 2.10.x and then 2.10.x -> 2.11.x

PR #3111, opened 2023-11-13
[Tracker] Prevent partial/invalid events to be indexed

rbayet:


src/module-elasticsuite-tracker/Model/ResourceModel/EventQueue.php
```diff
@@ -76,11 +77,14 @@ public function getEvents($limit = null)
...
+        $select->where('is_invalid = 0');
+
+        $eventData = $this->getConnection()->fetchAssoc($select);
 
         foreach ($eventData as &$currentEvent) {
             try {
                 $currentEventData = $this->jsonSerializer->unserialize($currentEvent['data']);
+                $currentEventData['is_invalid'] = $this->isEventInvalid($currentEventData);
             } catch (\InvalidArgumentException $exception) {
                 $currentEventData = [];
```
rbayet, line 89: Also add "is_invalid" here instead of indexing an empty event because the data could not be deserialized ?

PR #3112, opened 2023-11-13
Fix #2913, undefined array key date for elasticsuite tracker event index

### LINKED ISSUE #2913: Search Usage Analytics - tracking_log_session index does not exist yet - 2.11.1
I have seen a previously raised issue regarding this, however the issue that I am having can be traced to an error in the indexing.

The elasticsuite_tracker_log_event table is being populated with front end visit information as expected, the cronjob is being attempted, but fails with the following error in the cron log:

main.ERROR: Cron Job elasticsuite_index_log_event has an error: Warning: Undefined array key "date" in /html/vendor/smile/elasticsuite/src/module-elasticsuite-tracker/Model/EventIndex.php on line 79. 

Which is the line:

`$date = substr($event['date'], 0, 7);`

From this function: 

```
public function indexEvents($events)
    {
        $bulk    = $this->indexOperation->createBulk();
        $indices = [];
        foreach ($events as $event) {
            if (isset($event['page']['store_id'])) {
                $date = substr($event['date'], 0, 7);
                $index = $this->indexResolver->getIndex(self::INDEX_IDENTIFIER, $event['page']['store_id'], $date);
                if ($index !== null) {
                    $event = $this->mappingEnforcer->enforce($index, $event);
                    $indices[$index->getName()] = $index;
                    $bulk->addDocument($index, $event['event_id'], $event);
                }
            }
        }
        if ($bulk->isEmpty() === false) {
            $this->indexOperation->executeBulk($bulk);
        }
        foreach ($indices as $index) {
            $this->indexOperation->refreshIndex($index);
        }
    }
```
        

This in turn makes the cron job fail and the index never appear.

Magento Version : 2.4.6
ElasticSuite Version : 2.11.1
Environment : Developer 
Elasticsearch 7.17

Third party modules :
Mirasvit Cache Warmer
BSS Commerce  Out of stock Notification
BSS Commerce B2B Registration
BSS Commerce Hide Price
Xtento Product Feed Export Module

rbayet:


src/module-elasticsuite-tracker/Model/EventIndex.php
```diff
@@ -76,6 +76,11 @@ public function indexEvents($events)
 
         foreach ($events as $event) {
             if (isset($event['page']['store_id'])) {
+                // Previous "date" column has been renamed to "created_at" in db_schema.xml
```
romainruaud, line 79: missing a dot at the end, PHPcs will complain :)

rbayet: Yes, saw that. The only remaining diff between 2.10.x and 2.11.x tracker module is now a comment in the db_schema.xml

PR #3124, opened 2023-11-28
Fixes #3123 [Search Merchandizing] Show back hidden products in preview

### LINKED ISSUE #3123: [Search Merchandizing] Hidden/blacklisted products disappear from the preview
In the merchandizer of the search term, after hiding a product and saving the search term, the product is no longer visible, which prevents un-hiding the product.

Introduced by 1c8e1d7137e263b412efe808aef936f18df0f05a 
* previously the \Smile\ElasticsuiteCatalog\Model\Product\Search\Request\Container\Filter\SearchBlacklist filter did not apply because the current search query was not set in the context
   * thus the products configured to be hidden were still part of the response and just flagged as such
* now the filter always apply unconditionnally
   * thus the products configured to be hidden are no longer part of the response and cannot be drawn in the preview area

ElasticSuite Version : => 2.10.3, => 2.11.0

rbayet:


romainruaud (approved):
All good you can merge this

PR #3127, opened 2023-11-30
Added configuration for ssl verification

gabrielLumao:
Added configuration, default value for it, and elasticsearch call option for SSL Verification. This allows to turn it off, for ES clients like opensearch, on local environments.


rbayet:
Hello @gabrielLumao,

Just to be sure, does the SSL verification needs to be disabled systematically if using HTTPS with OpenSearch (since that's what the configuration field says) or this is only for a local/dev OpenSearch instance ?

Regards,

gabrielLumao:
Hello @rbayet,

It was added for local/dev instances with self-signed SSL certificates.

Regards,

Issue #3117, opened 2023-11-20
Attribute option translation


Nuranto:
### Preconditions

Magento Version : 2.4.6-p3

ElasticSuite Version : 2.11.4.1

### Steps to reproduce

1. Create filterable attribute with custom source model : `[   ['value'=>1, 'label' => __('Apple')->render()], ['value'=>2, 'label' => __('Car')->render()].  ]` 
2. Add in fr_FR translation file : "Car,Voiture" and "Apple,Pomme"
3. Reindex catalogsearch_fulltext
4. Check filters on your fr_FR store

### Expected result

1. Filter options are in french (Pomme, Voiture)

### Actual result

1. Filter options are in english (Apple, car)

### Additional informations

Quick fix I found is to manually load translations in my source model : 

```
        $areaList = ObjectManager::getInstance()->get(\Magento\Framework\App\AreaList::class);
        $area = $areaList->getArea($this->appState->getAreaCode());
        $area->load(\Magento\Framework\App\Area::PART_TRANSLATE);
```

But that's obviously not a good solution, that should be done somewhere in smile module. I did not found the correct place to do it though, we should probably emulate store somewhere in index rebuilding process.

vahonc:
Hello @Nuranto,

I was able to reproduce your issue. Indeed the attribute options translation does not work through any language (e.g. `fr_FR`) translation file no matter if it's using the custom source model or default. This is due to the fact that Elasticsearch indexes an `Apple` value, so if you try to search `Pomme` you don't find any products.

We're not a huge fan of adding an emulate store somewhere in the index rebuilding process. As an easy solution, Magento proposes a way to add translation via `Admin Panel -> Attribute Properties -> Manage Options (Values of Your Attribute)`. Therefore I have a question, why you are using a custom source model and why is translation via Admin not suitable for you?

BR,
Vadym

Nuranto:
Hi @vahonc,

It reminds me a similar question from Romain here : https://github.com/Smile-SA/elasticsuite/issues/2932

Here are some examples : 
- An attribute that you want custom methods for. For example, a month attribute explained in #2932
- An attribute that has a dynamic source. For example, you may want to create a Model/DB schema/CRUD for brands, and link those to products, so you can then easily, for instance, add a link to the custom brand page on product view, or display extra brand informations, etc..
- An attribute that you don't want administrators to mess with options. For instance, native `status`, `price_type`, `shipment_type`, ... attributes.

[...]

PR #3135, opened 2023-12-15
#3117 - Fix attribute option translation while reindexing

Nuranto:
see https://github.com/Smile-SA/elasticsuite/issues/3117


Nuranto:
This works perfectly when launching a reindex manually. But when it's reindexing via mview update cron, it does not anymore. So I have most of my options translated, but some are not, leaving filters very messy :/

I don't understand how that's possible, if you have any idea, it'll be helpfull.

Anyway, don't merge this PR for now.

Nuranto:
Latest commits fixed issues.

PR #3144, opened 2023-12-30
ci: Use GITHUB_OUTPUT envvar instead of set-output command

arunsathiya:
`save-state` and `set-output` commands used in GitHub Actions are deprecated and [GitHub recommends using environment files](https://github.blog/changelog/2023-07-24-github-actions-update-on-save-state-and-set-output-commands/).

This PR updates the usage of `set-output` to `$GITHUB_OUTPUT`

Instructions for envvar usage from GitHub docs:

https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

### DIFF
--- a/.github/workflows/02-phpstan.yml
+++ b/.github/workflows/02-phpstan.yml
@@ -74,7 +74,7 @@ jobs:
 
       - name: "[Init] Determine composer cache directory"
         id: composer-cache-directory
-        run: "echo \"directory=$(composer config cache-dir)\"" >> $GITHUB_OUTPUT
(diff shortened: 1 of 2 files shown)


romainruaud:
Thanks for reporting this, but please change your PR so that it contain changes only related to the GITHUB_OUTPUT.

Actually you changed all the indent of the file, I cannot merge as is.

arunsathiya:
Thanks for highlighting that @romainruaud, that was my IDE's doing. It should be fixed now: efc00aadf84df9b02d9098a0567142dcb1accb98.

PR #3152, opened 2024-01-15
update of BO comment LU-140

gabrielLumao:
Updated comment of enable_certificate_validation field, to make it clearer to the users.


rbayet:
Thanks @gabrielLumao !

Indeed that's clearer for the end user (as you might have guessed from my question in the PR ;))

PR #3168, opened 2024-02-06
allow exact match for strings

mvenghaus:
Sometimes you are in need to make an exact match instead of like in strings. For example I have a string with values "5", "50". With like I have no possibility to get only the value "5".

### DIFF
--- a/src/module-elasticsuite-catalog-rule/Model/Rule/Condition/Product.php
+++ b/src/module-elasticsuite-catalog-rule/Model/Rule/Condition/Product.php
@@ -265,7 +265,7 @@ class Product extends \Magento\Rule\Model\Condition\Product\AbstractProduct
     {
         if (null === $this->_defaultOperatorInputByType) {
             $this->_defaultOperatorInputByType = [
-                'string'      => ['{}', '!{}'],
+                'string'      => ['{}', '!{}', '=='],
                 'numeric'     => ['==', '!=', '>=', '>', '<=', '<'],
                 'date'        => ['==', '>=', '>', '<=', '<'],
                 'select'      => ['==', '!='],


romainruaud:
Hi @mvenghaus 

what's the output of this in the rule form ?

mvenghaus:
german .. but i think you get it ;)

image

rbayet:
Hello @mvenghaus,

Adding the operator is not enough though, one also needs to implement the specific query that must be generated to perform the "IS" operation which is to be intended as "contains this ... AND ONLY THIS".
Otherwise, you will have the exact same "match" query generated [by those lines](https://github.com/Smile-SA/elasticsuite/blob/a08fd850dafaeb685a970e3f3608371875525fc7/src/module-elasticsuite-catalog-rule/Model/Rule/Condition/Product/QueryBuilder.php#L91-L93).

The initial idea is probably to perform a "[term](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-term-query.html)" query.
The issue is that "string" searchable-only attributes do not have an "untouched" subfield version (like option_text_color.untouched) 
of type "[keyword](https://www.elastic.co/guide/en/elasticsearch/reference/current/keyword.html#keyword-field-type)" on which you can safely perform it.
So you'll probably need to target the base field of a searchable-only text attribute (like "name") which is of type "text" but has the "keyword" analyzer.

It looks like a query like this works as expected : 
```
{
  "size": 20,
  "sort": [
    {
      "_score": {
        "order": "desc"
      }
    },
    {
      "entity_id": {
        "order": "desc",
        "missing": "_first",
        "unmapped_type": "keyword"
      }
    }
  ],
  "from": 0,
  "query": {
      "term": {
          "name": "Joust Duffle Bag"
      }
  }
}
```

You would need to check that this approach works for all string/text attributes susceptible of appearing in the rules engine either because 
* they are searchable only
* or they are filterable (we allow it, so meh...) only
* or they are sortable only
* or they have "use in promo rules" flag set to yes only
* any combination of the above

Regards,

PS: As a side note, if you add "is", you should also probably add "is not" as a way to exclude a particular product in the context of a more complex rule.
The negation should be taken care of automatically in the QueryBuilder [on those lines](https://github.com/Smile-SA/elasticsuite/blob/a08fd850dafaeb685a970e3f3608371875525fc7/src/module-elasticsuite-catalog-rule/Model/Rule/Condition/Product/QueryBuilder.php#L101-L103).

mvenghaus:
hey @rbayet ,

sorry for the late response! 

I've created another pull request, because it combines another issue.

https://github.com/Smile-SA/elasticsuite/pull/3230

Regards
Marcus

PR #3174, opened 2024-02-08
Allow configurable msm for fuzzy search

romainruaud:
It was previously hardcoded to "100%", leading to some strange cases if the default minimum_should_match was not 100% but configured to something else.

Eg : 

- default MSM of "2<80%"
- search for "open front dress" on a Venia Catalog => 12 results.
- search again for "open front dr**i**ss" => 0 results.

because the search with a typo triggered a fuzzy search that has an hardcoded value of 100% for minimum_should_match.


rbayet:
Replaced by #3196 on 2.10.x

PR #3183, opened 2024-02-15
Change !empty check to isset

peterjaap:
See https://github.com/magento/magento2/commit/7ab48b08dc6cda2853c95033e44152fa98f89258

### DIFF
--- a/src/module-elasticsuite-catalog-graph-ql/DataProvider/Product/SearchCriteriaBuilder.php
+++ b/src/module-elasticsuite-catalog-graph-ql/DataProvider/Product/SearchCriteriaBuilder.php
@@ -63,7 +63,7 @@ class SearchCriteriaBuilder extends \Magento\CatalogGraphQl\DataProvider\Product
     public function build(array $args, bool $includeAggregation): SearchCriteriaInterface
     {
         $searchCriteria = $this->builder->build('products', $args);
-        $isSearch       = !empty($args['search']);
+        $isSearch = isset($args['search']);
 
         $requestName = 'catalog_view_container';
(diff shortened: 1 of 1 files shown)


romainruaud:
Hi @peterjaap 

what's the point of changing this, exactly ? Does it change the behavior ?

The notable change in their PR is the fact that they now check whether they have a category_id filter or not. Previously they were checking only if there is a filter.

peterjaap:
@romainruaud I don't know exactly why, but our upgrade testing tooling flagged this is a change in Magento which wasn't reflected in this preference in your extension

romainruaud:
Ok @peterjaap 

this upgrade testing tool looks nice, is it public ? I'd like to have a peek on that, this might help us a lot as well, especially with Magento 2.4.7 coming.

regards

peterjaap:
@romainruaud yes it is;
- https://github.com/ampersandHQ/ampersand-magento2-upgrade-patch-helper/ which output files you can feed to:
- https://github.com/elgentos/magento2-upgrade-gui

rbayet:
Meh, @romainruaud, @peterjaap, I'm not so sure about this change.

In the pointed commit, they fix their sort of invalid test on the whether there is an applied category filter but IMHO they worsen the one on the search parameter.

I bet there are a ton of hastily coded custom frontend themes relying on the GraphQL API were people provide systematically the "search" parameter with an empty string value when there is no actual user search query instead of checking if the parameter should be provided in the first place.
That legacy `!empty` test prevented to enable the search mode if the query string was null or an empty string.
The new `isset` only protects against null values.

Regards,
Richard

PR #3189, opened 2024-02-20
Fix tracking image 404

chrisastley:
The tracking image URL was h.png and not hit.png as per the image in the module.


rbayet:
Sorry about the delay @chrisastley, this one slipped in the cracks.

Actually, the issue is not the filename/filepath to the image, the "hit.png" image present in the source code is there to confuse people.
=> The requested image (when the tracker is in "invisible pixel" mode) is actually sent by directly [providing the image content in the response body](https://github.com/Smile-SA/elasticsuite/blob/8ed9cd2ef3f95f3c7f2ef6139f341856c4a264b0/src/module-elasticsuite-tracker/Controller/Tracker/Hit.php#L57).

I'm replacing your PR by this one : #3642

PR #3194, opened 2024-02-23
Update CategoryData.php

notimplementedyet:
wrong base category postitions when virtual category is enabled


romainruaud:
This is not a good solution. I'd prefer to have a setup file copying positions from the legacy table in our table, this will be more future-proof and allow back-office editing of the positions.

Regards

PR #3197, opened 2024-02-26
Enable Elasticsuite cache tag automatically.

romainruaud:


rbayet:
Hello @romainruaud,

Isn't there another way to enable it but only once ?
As it stands, it being in a recurring profile means that if you disable it manually (once it has been enabled by running setup:upgrade) and re-run setup:upgrade (for another reason) then it's enabled again.

Basically, there is no way to forcefully disable it.

Regards,

romainruaud:
As discussed, I updated so that we'll enforce enabling the cache only if it does not exists in the current config.

rbayet (approved):

PR #3199, opened 2024-02-26
Spellchecker : Fix multi index by alias error

PierreGauthier:


src/module-elasticsuite-core/Search/Adapter/Elasticsuite/Spellchecker.php
```diff
@@ -177,10 +177,14 @@ private function getTermVectors(RequestInterface $request)
...
         $docs = [];
 
-        // Compute the mtermvector query on all shards to ensure exhaustive results.
-        foreach (range(0, $shards - 1) as $shard) {
-            $doc['routing'] = sprintf("[%s][%s]", $request->getIndex(), $shard);
-            $docs[] = $doc;
+        // Compute the mtermvector query on all indices.
+        foreach (array_keys($stats['indices']) as $indexName) {
+            // Compute the mtermvector query on all shards to ensure exhaustive results.
+            foreach (range(0, $shards - 1) as $shard) {
```
romainruaud, line 183: If you have 2 indices having the same alias, and each of these indices having 3 shards.

What do you have in $shards at this step ? 6 ?

PierreGauthier: Yes, but you don't have the information of which shard is assigned to which index

romainruaud: hmmm so if I understand properly, if we have : 

- index "aaaa" with 3 shards
- index "bbbb" with 3 shards
- they both share the "product" alias

In the end, in routing, we'll have : 

- [aaaa][0], [aaaa][1], [aaaa][2], [aaaa][3], [aaaa][4], [aaaa][5]
- [bbbb][0], [bbbb][1], [bbbb][2], [bbbb][3], [bbbb][4], [bbbb][5]

And it will work even if shards 3 4 and 5 does not exist for these index ?

PierreGauthier: Yes your example is correct and it seems to work properly even if the shard doesn't exist

PR #3210, opened 2024-03-14
Revert "Change !empty check to isset"

rbayet:
Reverts Smile-SA/elasticsuite#3183


rbayet:
See [my comment](https://github.com/Smile-SA/elasticsuite/pull/3183#issuecomment-1995693696) in #3183 
Agreed with @romainruaud

PR #3215, opened 2024-03-16
Change opensearch deps to make it compatible with magento 2.4.7-beta3

Bashev:
I'm not sure why these limitation was set for version below 2.0.1, but this reflects to critical errors in magento 2.4.7-beta3 (and probably the final release).

I put the same requirements as they are in the Opensearch module in Magento https://github.com/magento/magento2/blob/b2286ecfb014dbe05a45004526541aa8250259c2/app/code/Magento/OpenSearch/composer.json#L12

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,7 @@
         "magento/module-inventory-sales": ">=1.1.0",
         "magento/module-inventory-indexer": ">=2.0.0",
         "magento/magento-composer-installer": "*",
-        "opensearch-project/opensearch-php": "^1.0 || ^2.0, <2.0.1"
+        "opensearch-project/opensearch-php": "^1.0 || ^2.0"
     },
     "replace": {
         "smile/module-elasticsuite-core": "self.version",


rbayet:
Hello @Bashev,

I'll let @romainruaud have a look, may'be this was cause by a specific issue in version 2.0.1 of the OpenSearch PHP client (which was obviously not the latest when we made the switch from the Elasticsearch PHP client to the OpenSearch one).

Regards,

romainruaud:
I just copy pasted it from Magento 2.4.6 requirements guys : 

https://github.com/magento/magento2/blob/2.4.6-p4/composer.json#L79

So I have no idea why they made this :)

hostep:
Hi @romainruaud: when can we expect a new release with this fix included? Would be great if this can happen somewhere in the next few days, so when Magento 2.4.7 is released on the 9th of April, this module will not cause upgrade issues.

Thanks! :)

PR #3221, opened 2024-03-19
Remove invalid characters

thomas-kl1:
Invalid comment tags prevent load of js scripts in adminhtml.

### DIFF
--- a/src/module-elasticsuite-tracker/view/adminhtml/requirejs-config.js
+++ b/src/module-elasticsuite-tracker/view/adminhtml/requirejs-config.js
@@ -1,4 +1,3 @@
-<!--
 /**
  * DISCLAIMER
  *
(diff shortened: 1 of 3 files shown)


thomas-kl1:
@romainruaud Hello could process it quickly? It's breaking the adminhtml area.

romainruaud:
do you have a particular configuration for JS files ? We're not having the issue locally.

thomas-kl1:
When I deploy statics with minification I have those characters in the generated requires config file and it throws a runtime error

rbayet:
Replaced by #3222 on 2.10.x

rbayet:
Nice catch @thomas-kl1 
I've rebased your commits on 2.10.x and squashed them in a new PR #3222

PR #3229, opened 2024-03-22
Add Ingest Pipeline support.

romainruaud:


src/module-elasticsuite-core/Index/IndexOperation.php
```diff
@@ -138,6 +147,12 @@ public function createIndex($indexIdentifier, $store)
         // @codingStandardsIgnoreEnd
         $indexSettings['settings']['analysis'] = $this->indexSettings->getAnalysisSettings($store);
 
+        // Add (and create, if needed) default pipeline.
+        $pipeline = $this->pipelineManager->createByIndexIdentifier($indexIdentifier);
+        if ($pipeline !== null) {
+            $indexSettings['settings']['default_pipeline'] = $pipeline->getName();
+        }
```
rbayet, line 154: So this will create a pipeline to every index created by Elasticsuite, whatever its type ?

src/module-elasticsuite-core/etc/di.xml
```diff
@@ -82,6 +82,17 @@
     <preference for="Magento\Framework\Api\Search\SearchCriteriaInterface"
                 type="Smile\ElasticsuiteCore\Search\SearchCriteria" />
 
+    <!-- Pipeline Management related interfaces -->
+
```
rbayet, line 86: Unnecessary line feed ?

src/module-elasticsuite-core/Index/IndexOperation.php
romainruaud: The pipeline created will depend on the index identifier. 

The pipeline manager will only create a pipeline if it has processors (if processors are injected via DI).

So for an indexIdentifier without any processor, no pipeline will be created, therefore, $pipeline will be null and not applied as the default one.

PR #3242, opened 2024-04-09
[Analysis] Allow ignoring leading, trailing and consecutive zeroes in…

rbayet:
… reference fields


rbayet:
Replaced by #3243 on 2.10.x

PR #3243, opened 2024-04-09
[Analysis] Allow ignoring leading, trailing and consecutive zeroes in…

rbayet:
… reference fields


src/module-elasticsuite-core/Helper/IndexSettings.php
```diff
@@ -40,6 +41,28 @@ class IndexSettings extends AbstractConfiguration
...
+
+    /**
+     * @var string
+     */
+    const REFERENCE_ANALYZER_CONFIG_XML_PREFIX = 'reference_analyzer';
+
+    /**
+     * @var string
+     */
+    const OLD_DEFAULT_INDICES_PATTERN = '{{YYYYMMdd}}_{{HHmmss}}';
```
romainruaud, line 59: do we need this in this commit ?

src/module-elasticsuite-core/Helper/IndexSettings.php
```diff
@@ -40,6 +41,28 @@ class IndexSettings extends AbstractConfiguration
...
+
+    /**
+     * @var string
+     */
+    const DEFAULT_INDICES_PATTERN = '{{Ymd}}_{{His}}';
```
romainruaud, line 64: same as above

src/module-elasticsuite-core/Helper/IndexSettings.php
rbayet: Strange, this is something related to the cherry-pick from the based-on-2.11.x feature branch that I thought was removed.

src/module-elasticsuite-core/Helper/IndexSettings.php
rbayet: Same as above

PR #3277, opened 2024-05-14
[Tracker] Fix inline script execution on Checkout page according to CSP restrictions

vahonc:
According to the CSP updates:
>The default CSP configuration for payment pages for Commerce Admin and storefront areas is now restrict mode. For all other pages, the default configuration is report-only mode. In releases prior to 2.4.7, CSP was configured in report-only mode for all pages.

now an inline scripts should be wrapped using `$secureRenderer` variable in the `.phtml` templates, but for some reason this blocks the work of our Tracker script on Checkout page.

See more:
- https://developer.adobe.com/commerce/php/development/security/content-security-policies/#troubleshooting 
- https://developer.adobe.com/commerce/php/development/security/content-security-policies/#whitelist-an-inline-script-or-style

### DIFF
--- a/src/module-elasticsuite-tracker/view/frontend/templates/config.phtml
+++ b/src/module-elasticsuite-tracker/view/frontend/templates/config.phtml
@@ -14,27 +14,47 @@
  * @license   Open Software License ("OSL") v. 3.0
  */
 ?>
-<?php /** @var $block Smile\ElasticsuiteTracker\Block\Config **/ ?>
-<?php $jsonHelper = $block->getJsonHelper(); ?>
-<?php if ($block->isEnabled()) : ?>
-<script>
+<?php
+/**
+ * @var $block Smile\ElasticsuiteTracker\Block\Config
+ * @var $secureRenderer Magento\Framework\View\Helper\SecureHtmlRenderer
+ */
+?>
+<?php
+$jsonHelper = $block->getJsonHelper();
+if ($block->isEnabled()) {
+$beaconUrl = $this->escapeJsQuote($block->getBeaconUrl());
+$telemetryUrl = $this->escapeJsQuote($block->getTelemetryUrl());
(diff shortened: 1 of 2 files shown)


src/module-elasticsuite-tracker/view/frontend/templates/config.phtml
```diff
@@ -14,34 +14,56 @@
...
-    };
-<?php if ($block->isUsingAPI()): ?>
-    trackerConfig.endpointUrl = '<?= /* @noEscape */ $this->escapeJsQuote($block->getEndpointUrl()); ?>';
-<?php endif; ?>
-    smileTracker.setConfig(trackerConfig);
+    smileTracker.setConfig({
+        beaconUrl: '{$beaconUrl}',
+        telemetryUrl: '{$telemetryUrl}',
+        telemetryEnabled: '{$telemetryEnabled}',
+        sessionConfig: '{$sessionConfig}'
```
hostep, line 38: @vahonc: I believe there is a bug introduced on this line, the single quotes should not be needed around `{$sessionConfig}`. If the json data contains a single quote, this will break. In the old code, there also weren't single quotes surrounding this value.

hostep: My bad, it looks like @romainruaud already fixed this shortly after merging this, in https://github.com/Smile-SA/elasticsuite/commit/daff073c9e16f2e7a44a5b90cc9fc01d5f3fb4ce, thanks guys!

romainruaud: Yes correct @hostep :)

vahonc: Sorry guys, it's just a typo, my bad

PR #3280, opened 2024-05-24
[VirtualCategories] Making virtual categories queries named for debugging purposes

rbayet:


romainruaud:
Can we merge this @rbayet even if it's WIP ? this could help a lot even in this current state, no ?

PR #3283, opened 2024-05-29
[Core] Ability to set name to a QueryInterface post creation

rbayet:


rbayet:
Integration failures on 2.4.4 are experimental so are to be ignored

PR #3286, opened 2024-05-30
[VirtualCategories] Option to force empty category if virtual category root disabled

rbayet:
Work in Progress

### DIFF
--- a/src/module-elasticsuite-virtual-category/Model/Rule.php
+++ b/src/module-elasticsuite-virtual-category/Model/Rule.php
@@ -23,6 +23,7 @@ use Smile\ElasticsuiteCore\Search\Request\QueryInterface;
 use Magento\Catalog\Api\Data\CategoryInterface;
 use Smile\ElasticsuiteVirtualCategory\Api\Data\VirtualRuleInterface;
 use Smile\ElasticsuiteCatalogRule\Model\Rule\Condition\Product\QueryBuilder;
+use Smile\ElasticsuiteVirtualCategory\Helper\Config;
 use Smile\ElasticsuiteVirtualCategory\Model\ResourceModel\VirtualCategory\CollectionFactory;
 use Smile\ElasticsuiteCore\Search\Request\Query\QueryFactory;
 use Magento\Catalog\Model\CategoryFactory;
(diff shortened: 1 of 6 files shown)


src/module-elasticsuite-virtual-category/Model/Rule.php
```diff
@@ -189,6 +198,7 @@ public function getCategorySearchQuery($category, $excludedCategories = []): ?Qu
             ]
         );
 
+        $query = false;
```
rbayet, line 201: Why is this necessary ?

src/module-elasticsuite-virtual-category/Model/Rule.php
```diff
@@ -306,9 +316,14 @@ private function buildCategorySearchQuery($category, $excludedCategories = []):
         if (!in_array($category->getId(), $excludedCategories)) {
             $excludedCategories[] = $category->getId();
 
-            if ((bool) $category->getIsVirtualCategory() && $category->getIsActive()) {
+            if (!$category->getIsActive()
+                && $this->config->isForceZeroResultsForDisabledCategoriesEnabled($this->getStoreId())) {
+                return $this->getNoResultsQuery();
+            }
+
+            if ($category->getIsVirtualCategory()) {
```
rbayet, line 324: Be careful, you are creating a specific route that is not correctly covered.
You cannot remove the test on `$category->getIsActive()` on lines 324 and 326 since the test on 319 also relies on the new setting.
So if "isForceZero..." returns false, the category could be disabled on lines 324 and 326...

src/module-elasticsuite-virtual-category/Model/Rule.php
```diff
@@ -189,6 +198,7 @@ public function getCategorySearchQuery($category, $excludedCategories = []): ?Qu
             ]
         );
```
rbayet, line 199: Shouldn't the value of the setting be taken into account for the cacheKey ?

src/module-elasticsuite-virtual-category/Model/Rule.php
PierreGauthier: I thought we might don't want to return "null" query anymore that can have a quite unexpected behavior ?
The opposite of forcing an empty result for empty disabled category should be to treat them as normal categories no ?

src/module-elasticsuite-virtual-category/Model/Rule.php
PierreGauthier: But if we add it to the cache key, it will invalidate all the category cache if the user update this config as none of the cache key will match anymore ?

rbayet:
We need to discuss this some more.

PR #3303, opened 2024-06-21
[Tracker] Fix #3301, improving tracker compatibility with template minifying

### LINKED ISSUE #3301: Broken tracking code when template minifying is enabled
The Elasticsuite Tracker JavaScript is not compatible with template minifying. The output is invalid syntax and does not work.

### Preconditions

Config `dev/template/minify_html` set to `1`.
Config `smile_elasticsuite_tracker/general/enabled` set to `1`.

Magento Version : 2.4.6-p6

ElasticSuite Version : 2.11.7

Environment : Production

Third party modules :

### Steps to reproduce

1. Tracking and template minifying need to be enabled
2. Open up a page containing the tracking code, e.g. the home page

### Expected result

1. Tracking is compatible with template minifying

### Actual result

1. Output will look something like this:
```html
<script> //<![CDATA[ try { smileTracker.addPageVar('type.identifier', 'cms_index_index');
smileTracker.addPageVar('type.label', 'CMS-Startseite');

smileTracker.addPageVar('locale', 'de_DE');
} catch
(err)
{
    ;
} //]]> </script>
```
Note how the first line of code is commented out.

### More info
I believe this bug was introduced in this commit: https://github.com/Smile-SA/elasticsuite/commit/da235f79945aaf2137737b1d5bc562c0e6ff0bf2

The original file contains a line break before the `try` statement:
```php
$scriptString = "
//<![CDATA[
try {
";
```

The minified file has these lines merged into one, breaking the code:
```php
$scriptString = " //<![CDATA[ try { ";
```

EDIT: I've only analyzed [this file](https://github.com/Smile-SA/elasticsuite/blob/2.11.7/src/module-elasticsuite-tracker/view/frontend/templates/variables/page.phtml), but I believe more files might be affected

vahonc:


romainruaud:
tested OK but was replaced by #3318 to be integrated in 2.10.x directly.

PR #3368, opened 2024-09-04
[Catalog] Fix #2948, fallback to the last known valid product positions

### LINKED ISSUE #2948: Manual sort gets set back to automatic sort
Our customer has a problem where when they change the sort order and/or change some products back to automatic search many products automatically get set back to automatic sort only some remain in manual sort.

We found out that a reload of the site may change it back.
Also sometimes after rearranging the sort order, the category does not show all products.

### Environment
Magento Version: 2.4.5-p1
ElasticSuite Version: 2.10.15.1

### Steps to reproduce
1. Change the sort order of products
2. Repeat as often as necessary

### Expected result
1. Manual sort order remaining

### Actual result
1. Many products get set back to automatic sort

vahonc:


src/module-elasticsuite-virtual-category/Plugin/Catalog/Category/SaveProductsPositions.php
```diff
@@ -145,17 +153,23 @@ private function getAffectedProductIds($category)
...
      * @param \Magento\Catalog\Model\Category $category Category
      *
-     * @return array
+     * @return $this
      */
     private function unserializeProductPositions(\Magento\Catalog\Model\Category $category)
     {
-        $productPositions = $category->getSortedProducts() ? $category->getSortedProducts() : [];
+        // Get product positions from the category.
+        $productPositions = $category->getSortedProducts();
```
rbayet, line 161: I don't understand why this line was changed but it does not seem to affect the behaviour of the feature or fix so far, even with multi-store and per-store view positions.

vahonc: The code now directly fetches product positions from the category using `$category->getSortedProducts()`. It will return either the current list of sorted products or null/an empty array if none are set. Therefore, there's no need to use a ternary operator to check and assign an empty array. Then if the data returned by `$category->getSortedProducts()` is not a valid JSON string (e.g., it's malformed or truncated), the jsonDecode() function will fail to parse it, the fallback is to use `getProductPositionsByCategory($category)`. The warning message is displayed to notify the admin about the fallback, providing a clear indication that something went wrong.

rbayet: My point is about he ternary operator removal.
> It will return either the current list of sorted products or null/an empty array if none are set. 

Before, it would either return the currently submitted list of sorted product (in a not 100% determined format) **or an empty array**.
The only thing preventing to do a `$category->setSortedProducts($productPositions);` is the initial check `if ($category->getId() && $category->getSortedProducts()) {`.

rbayet:
Replaced by #3386

PR #3399, opened 2024-10-01
Improve sortOrder direction fallback in case of empty string

frank-bokdam:
when direction is an empty string, no results will be display in preview due to in error in the call to elastic:

```
{"error":{"root_cause":[{"type":"x_content_parse_exception","reason":"[1:50] [field_sort] failed to parse field [order]"}],"type":"x_content_parse_exception","reason":"[1:50] [field_sort] failed to parse field [order]","caused_by":{"type":"illegal_argument_exception","reason":"No enum constant org.elasticsearch.search.sort.SortOrder."}},"status":400}
```


romainruaud:
Hi @frank-bokdam this is a good guard function, but in which case could this parameter be empty ? and by the way, I thought the getParam() second parameter was intended to be a fallback for such cases ?

Best regards

PR #3404, opened 2024-10-03
Fix #3403 DateTime constructor in IndexStatusProvider

### LINKED ISSUE #3403: DateMalformedStringException - Failed to parse time string (Ymd) at position 0 (Y) in module-elasticsuite-indices/Model/IndexStatusProvider.php:120
### Preconditions

Rocky Linux release 9.4 (Blue Onyx), PHP 8.3.10, Elasticsearch 7.17.9

Magento Version : CE 2.4.7

ElasticSuite Version : 2.11.9

Environment :  Production Mode

Third party modules : Yes. Too many and that's confidential.

### Steps to reproduce

1. In CLI type `php bin/magentoindexer:reset catalogsearch_fulltext && php bin/magento indexer:reindex catalogsearch_fulltext`
2. Check the exception log `tail -n 200 var/log/exception.log`

### Expected result

1. No errors should be visible in exception log.

### Actual result

1. `[2024-10-03T14:12:30.202727+00:00] main.ERROR: Error when loading/parsing statistics for index "magento2_default_catalog_category_20240729_111506" {"exception":"[object] (DateMalformedStringException(code: 0): Failed to parse time string (Ymd) at position 0 (Y): The timezone could not be found in the database at vendor/smile/elasticsuite/src/module-elasticsuite-indices/Model/IndexStatusProvider.php:120)"} []` is visible.

`   $today = new DateTime('Ymd');` line was introduced in https://github.com/Smile-SA/elasticsuite/blame/2.11.x/src/module-elasticsuite-indices/Model/IndexStatusProvider.php#L120C18-L120C50 within Zend libraries removal in #2818 but that line doesn't actually make any sense.

According to https://www.php.net/manual/en/class.datetime.php DateTime class should be called with $datetime string not the format (e.g. `new DateTime(''now')` or `new DateTime('20231003')`

Calling DateTime class with param Ymd is an obvious bug non compatible with any previous PHP version.

Tomasz-Silpion:


romainruaud:
Replaced by #3418

Regards,

PR #3407, opened 2024-10-08
Typo fix

Lemongrabb:


romainruaud:
Hi,

anonymization delay is in days. It's the retention delay which is in month. I changed this as part of #3414 

Regards

PR #3414, opened 2024-10-17
Prevent tracker to open sessions when registering events.

romainruaud:


src/module-elasticsuite-tracker/Model/Customer/TrackingService.php
```diff
@@ -142,6 +142,7 @@ private function addCustomerLink($eventData)
 
                 $this->customerLinkResource->saveLink($data);
             }
+            unset($eventData['customer']['id']); // Do not persist the customer_id in ES index to preserve anonymization.
```
rbayet, line 145: Hmmm, unsetting something in $eventData in addCustomerLink will hardly have any effect to the event written to database beforehand on line 79, won't it ?
If we don't want to commit the customer id to the ES index, should we re-implement the whole "addCustomerLink" as eventProcessor ?
Or at least extract in a local variable the customerId, remove it from eventData and pass both to addCustomerLink ?
Or again call addCustomerLink *before* eventQueue->addEvent ?

rbayet (changes requested):

src/module-elasticsuite-tracker/Model/Customer/TrackingService.php
romainruaud: yes, lines 79 and 80 should be substituted

romainruaud:
I switched the lines @rbayet 

I agree at some point this should just be moved as an event processor

Issue #3340, opened 2024-08-02
[B2B] Add the company_id and customer_group_id in the tracking data


romainruaud:
**Step 1.**

When using the Magento B2B modules and browsing the front-office as a logged-in customer belonging to a particular company_id, add this data to the data being sent by the tracker to the backend.

Add the field here : https://github.com/Smile-SA/elasticsuite/blob/2.11.x/src/module-elasticsuite-tracker/etc/elasticsuite_indices.xml#L30
And here : https://github.com/Smile-SA/elasticsuite/blob/2.11.x/src/module-elasticsuite-tracker/etc/elasticsuite_indices.xml#L122

Do the same for the customer_group_id.

**Step 2.**

Then in the Search Analytics screen : 

- append a Customer group selector in the top of the page. Selecting a customer should refresh the tracking data and display only data for this customer group.

- if the B2B module is enabled, append a Company selector in the top of the page. Same, selecting a company should refresh the tracking data and display only data for this company.

Then, you'll need to add these data into the Report context : https://github.com/Smile-SA/elasticsuite/blob/2.11.x/src/module-elasticsuite-analytics/Model/Report/Context.php

And for the reports to work, you'll have to implement two classes for each company_id/customer_group_id : 

Smile\ElasticsuiteAnalytics\Model\Report\Event\CustomerGroupIdFilterQueryProvider
Smile\ElasticsuiteAnalytics\Model\Report\Event\CompanyIdQueryFilterProvider

Smile\ElasticsuiteAnalytics\Model\Report\Session\CustomerGroupIdFilterQueryProvider
Smile\ElasticsuiteAnalytics\Model\Report\Session\CompanyIdQueryFilterProvider

And inject them via the DI to the proper reports : https://github.com/Smile-SA/elasticsuite/blob/2.11.x/src/module-elasticsuite-analytics/etc/di.xml

The implementation of these classes should be inspired from the DateFilterQueryProvider. Most probably, to filter on customer_group_id and company_id, a TERM query should be sufficient.

Something like : 

```
$companyId = $this->context->getCompanyId();
if ($companyId) {
    $query = $this->queryFactory->create(QueryInterface::TYPE_TERM, ['value' => $companyId]);
}
```

romainruaud:
this has been merged

PR #3457, opened 2024-11-28
[Analytics][Tracker] Feature #3340, add company_id and customer_group_id in tracking data

vahonc:


rbayet (changes requested):

rbayet:
Replaced by https://github.com/Smile-SA/elasticsuite/pull/3459

PR #3467, opened 2024-12-10
Adding fine-tuning options for the autocompletion

romainruaud:


src/module-elasticsuite-catalog/etc/adminhtml/system.xml
```diff
@@ -41,6 +41,7 @@
                     <comment><![CDATA[Maximum number of categories to display in autocomplete results.]]></comment>
                 </field>
             </group>
+
```
rbayet, line 44: So the idea is removing that "empty" change on that file.

rbayet (dismissed):

src/module-elasticsuite-catalog/etc/adminhtml/system.xml
rbayet, line 44: So the idea is to remove that "empty" change on that file.

rbayet (changes requested):

src/module-elasticsuite-catalog/etc/adminhtml/system.xml
romainruaud: I removed this line break. Regards

rbayet:
Replaced by #3487 to avoid the merge bubble (the merge conflict editor is :hankey: )

Issue #3452, opened 2024-11-26
[Indices] Add number of primary shards and number of replicas in the indices grid


rbayet:
**Is your feature request related to a problem? Please describe.**
It is now possible to configure Elasticsuite to have a different number of primary shards and/or replicas per index type.
Additionally, behavioral data indices are never "re-constructed" so their number of primary shards and/or replicas might not reflect the **current** configuration settings but instead will reflect the number of primary shards and/or replicas **at the moment they were created**.

It could be nice to see at the first glance those indices' configuration in the indices grid.

**Describe the solution you'd like**
Add two columns "Shards" and "Replicas" to the indices grid to display respectively for each index 
* the number of configured primary shards
* the number of replicas

rbayet:
Implemented through #3477

PR #3477, opened 2025-01-06
[Indices] Feature #3452, add number of primary shards and replicas in the Indices grid

vahonc:


src/module-elasticsuite-indices/Model/IndexStatsProvider.php
```diff
@@ -140,6 +178,8 @@ public function indexStats($indexName, $alias): array
                 $data['size'] = $this->sizeFormatted((int) $indexStats['total']['store']['size_in_bytes']);
                 $data['size_in_bytes'] = $indexStats['total']['store']['size_in_bytes'];
             }
+            $data['number_of_shards'] = $this->getShardsConfiguration($indexName);
+            $data['number_of_replicas'] = $this->getReplicasConfiguration($indexName);
```
rbayet, line 182: Please move this at the top of the try/catch block because when an index is closed, it is still possible to read the index settings (which the shards and replicas number come from) while it is not possible to read the stats.
Please also add default values 'undefined' for number_of_shards and number_of_replicas at the top of the method, where you also have default values for size and number of documents : 
```
    public function indexStats($indexName, $alias): array
    {
        $data = [
            'index_name'          => $indexName,
            'index_alias'         => $alias,
            'number_of_documents' => 'undefined',
            'size'                => 'undefined',
        ];
```

Please also, as a whole, take the opportunity to replace 'undefined' as a whole in the class by 'N/A' which is more accurate.
But be careful to then also change the const "\Smile\ElasticsuiteIndices\Model\System\Message\WarningAboutClusterShardsMisconfig::UNDEFINED_SIZE" to "N/A".

src/module-elasticsuite-indices/Model/IndexStatsProvider.php
```diff
@@ -151,6 +191,84 @@ public function indexStats($indexName, $alias): array
...
+     *
+     * @SuppressWarnings(PHPMD.ElseExpression)
+     */
+    public function getIndexSettings(string $indexName): array
+    {
+        $cacheKey = self::CACHE_KEY_PREFIX . $indexName;
+
+        // Check if the settings are already in memory.
+        if (!isset($this->cachedIndexSettings[$cacheKey])) {
+            $cachedData = $this->cache->load($cacheKey);
```
rbayet, line 208: See top comment, use \Smile\ElasticsuiteCore\Helper\Cache::loadCache instead

src/module-elasticsuite-indices/Model/IndexStatsProvider.php
```diff
@@ -151,6 +191,84 @@ public function indexStats($indexName, $alias): array
...
+            } else {
+                $settingsData = $this->client->getSettings($indexName);
+
+                // Save to cache with a tag.
+                $this->cache->save(
+                    $this->serializer->serialize($settingsData),
+                    $cacheKey,
+                    $this->getCacheTags(),
+                    self::CACHE_LIFETIME
+                );
```
rbayet, line 221: See top comment, use \Smile\ElasticsuiteCore\Helper\Cache::saveCache instead

rbayet (changes requested):

rbayet (changes requested):

rbayet (approved):

PR #3486, opened 2025-01-14
[Virtual Category] Prevent using cached query for sub virtual category

PierreGauthier:
If we use this cached query we will build a category query with many useless clauses as the optimization of the method \Smile\ElasticsuiteVirtualCategory\Model\Rule::isVirtualCategoryRootInExcludeCategories will be avoided


src/module-elasticsuite-virtual-category/Model/Rule.php
```diff
@@ -630,4 +623,30 @@ private function getNoResultsQuery(): QueryInterface
...
+     * @throws NoSuchEntityException
+     */
+    private function isVirtualCategoryRootInExcludeCategories(CategoryInterface $category, array $excludedCategories): bool
+    {
+        $rootCategory = $this->getVirtualRootCategory($category);
+
+        return $category->getIsVirtualCategory()
+            && $rootCategory
+            && $rootCategory->getPath()
+            && array_intersect(explode('/', (string) $rootCategory->getPath()), $excludedCategories);
```
rbayet, line 650: Shouldn't the strict, exact test, be this ?
```
!empty(array_intersect(explode('/', (string) $rootCategory->getPath()), $excludedCategories)
```

PR #3487, opened 2025-01-14
[Core] Adding fine-tuning options for the autocompletion

rbayet:
Replaces #3467 (rebased with conflicts resolved)


src/module-elasticsuite-core/etc/config.xml
```diff
@@ -53,6 +53,12 @@
             <term_autocomplete>
                 <max_size>3</max_size>
             </term_autocomplete>
+            <advanced>
+                <extension_enabled>1</extension_enabled>
+                <extension_limit>0</extension_limit>
```
rbayet, line 58: Default value should be greater than 0 or the helper method should be reworked.

src/module-elasticsuite-core/Helper/Autocomplete.php
```diff
@@ -51,6 +51,68 @@ public function isEnabled($type)
...
+     * Check if Autocomplete "extension" system is limited.
+     *
+     * @return bool
+     */
+    public function isExtensionLimited()
+    {
+        return (bool) $this->scopeConfig->isSetFlag(
+            self::AUTOCOMPLETE_SETTINGS_CONFIG_XML_PREFIX . "/advanced/extension_limit",
+            \Magento\Store\Model\ScopeInterface::SCOPE_STORE
+        );
```
rbayet, line 77: If you want to allow a value of "0" for the extension_limit, ie, "No limit", then it should be 
```
isSetFlag && getExtensionSize() > 0
```

src/module-elasticsuite-core/etc/adminhtml/system.xml
```diff
@@ -180,6 +180,48 @@
...
+                    <label>Limit the amount of suggested search terms used for autocomplete matching</label>
+                    <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
+                    <depends>
+                        <field id="extension_enabled">1</field>
+                    </depends>
+                    <comment><![CDATA[Default: No. When set to "No", this setting will use the maximum number of suggested popular search terms configured in the dedicated section.]]></comment>
+                </field>
+                <field id="extension_size" translate="label comment" type="text" sortOrder="30" showInDefault="1" showInWebsite="1" showInStore="1" canRestore="1">
+                    <label>The number of suggested popular search terms to use for fetching results.</label>
+                    <validate>integer</validate>
```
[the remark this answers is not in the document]
romainruaud: I agree we could have a min value of 1, because there is already an option to completely disable the extension mechanism, which would be the same as having a combination of extension enabled and a value of 0.

src/module-elasticsuite-core/etc/config.xml
romainruaud: 0 should be "no limit" so let's rework the helper method instead.

src/module-elasticsuite-core/etc/adminhtml/system.xml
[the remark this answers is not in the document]
romainruaud: I agree this should be strictly greater than 0. No sense to put zero here because one can just disable the extension mechanism with the other parameter just above.

src/module-elasticsuite-core/Helper/Autocomplete.php
romainruaud: Nope, the "no limit" is the case where **"extension_limit = false".**

romainruaud:
So, I added 2 commits : 

- the first one to ensure the "size of search terms to use", aka **extension_size** must be **>0**
- the second one to rename the**extension_limit** to **extension_limited** because it's a **boolean** which defines if the extension mechanism must be limited or not.

src/module-elasticsuite-core/etc/adminhtml/system.xml
```diff
@@ -180,6 +180,48 @@
                     <comment><![CDATA[Maximum number of popular search terms to display in autocomplete results.]]></comment>
                 </field>
             </group>
+
+            <group id="advanced" translate="label" type="text" sortOrder="99" showInDefault="1" showInWebsite="1" showInStore="1">
+                <label>Advanced Settings</label>
+                <field id="extension_enabled" translate="label comment" type="select" sortOrder="10" showInDefault="1" showInWebsite="1" showInStore="1" canRestore="1">
+                    <label>Use suggested search terms to fetch results</label>
+                    <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
+                    <comment><![CDATA[Default: Yes. When enabled, autocompletion box results will be computed from the suggested search terms. This is intended to suggest results when the user is in the process of typing. Eg : When the user is typing "comp", the popular search term "computer" will be suggested, and therefore products or categories matching "computer" will be shown.]]></comment>
```
rbayet, line 189: First "sentence" should be "Default: Yes (legacy behavior)".
I would change the rest for clarity ("autocompletion box results" is a bit vague) to something like : 
"When set to Yes, products and category autocomplete will use the search terms suggested in the popular term autocomplete. This is intended to help find results when the user is in the process of typing a word." + 
(untouched) "Eg : When the user is typing "comp", the popular search term "computer" will be suggested, and therefore products or categories matching "computer" will be shown."

src/module-elasticsuite-core/etc/adminhtml/system.xml
```diff
@@ -180,6 +180,48 @@
...
+                </field>
+                <field id="preserve_base_query" translate="label comment" type="select" sortOrder="15" showInDefault="1" showInWebsite="1" showInStore="1" canRestore="1">
+                    <label>Always use the user raw input for suggestions.</label>
+                    <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
+                    <depends>
+                        <field id="extension_enabled">1</field>
+                    </depends>
+                    <comment><![CDATA[Default: No. When set to "Yes", this setting will always proceed to fetch suggestions on the query typed by the user in addition to the extended terms. Eg : When the user is typing "comp", the popular search term "computer" will be suggested, but products or categories matching either "comp" or "computer" will be shown.]]></comment>
```
rbayet, line 197: This also needs a bit of clarification IMHO.
I would change 
"Default: No." to "Default: No (legacy behavior). When set to "No", the user raw input, which might be a partial word, will only be used to fetch suggestions if no matching popular terms were found. When set to "Yes", it will always be used, whether matching popular terms were found or not. Eg if set to "Yes": when the user is typing "comp", the popular search term "computer" will be suggested, but products or categories matching either "comp" or "computer" will be shown."

src/module-elasticsuite-core/etc/adminhtml/system.xml
```diff
@@ -180,6 +180,48 @@
...
+                </field>
+                <field id="extension_limited" translate="label comment" type="select" sortOrder="20" showInDefault="1" showInWebsite="1" showInStore="1" canRestore="1">
+                    <label>Limit the amount of suggested search terms used for autocomplete matching</label>
+                    <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
+                    <depends>
+                        <field id="extension_enabled">1</field>
+                    </depends>
+                    <comment><![CDATA[Default: No. When set to "No", this setting will use the maximum number of suggested popular search terms configured in the dedicated section.]]></comment>
```
rbayet, line 205: I would precise "Default: No (legacy behavior)."
and
"When set to "No", all the suggested popular search terms will be used for the products and categories autocomplete."

src/module-elasticsuite-core/etc/adminhtml/system.xml
```diff
@@ -180,6 +180,48 @@
...
+                <field id="preserve_base_query" translate="label comment" type="select" sortOrder="15" showInDefault="1" showInWebsite="1" showInStore="1" canRestore="1">
+                    <label>Always use the user raw input for suggestions.</label>
+                    <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
+                    <depends>
+                        <field id="extension_enabled">1</field>
+                    </depends>
+                    <comment><![CDATA[Default: No. When set to "Yes", this setting will always proceed to fetch suggestions on the query typed by the user in addition to the extended terms. Eg : When the user is typing "comp", the popular search term "computer" will be suggested, but products or categories matching either "comp" or "computer" will be shown.]]></comment>
+                </field>
+                <field id="extension_limited" translate="label comment" type="select" sortOrder="20" showInDefault="1" showInWebsite="1" showInStore="1" canRestore="1">
+                    <label>Limit the amount of suggested search terms used for autocomplete matching</label>
```
rbayet, line 200: Not sure about "autocomplete matching". "Matching" makes more sense precisely for the popular search terms autocomplete while the products and categories are actually X dimensions (terms) search.

src/module-elasticsuite-core/etc/adminhtml/system.xml
```diff
@@ -180,6 +180,48 @@
...
+                    <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
+                    <depends>
+                        <field id="extension_enabled">1</field>
+                    </depends>
+                    <comment><![CDATA[Default: No. When set to "No", this setting will use the maximum number of suggested popular search terms configured in the dedicated section.]]></comment>
+                </field>
+                <field id="extension_size" translate="label comment" type="text" sortOrder="30" showInDefault="1" showInWebsite="1" showInStore="1" canRestore="1">
+                    <label>The number of suggested popular search terms to use for fetching results.</label>
```
rbayet, line 208: Shorter label ? Since the comment is basically a copy/paste.
Like "Maximum number of popular search terms to use".

src/module-elasticsuite-core/etc/adminhtml/system.xml
```diff
@@ -180,6 +180,48 @@
...
+                    <validate>integer validate-greater-than-zero</validate>
+                    <depends>
+                        <field id="extension_enabled">1</field>
+                        <field id="extension_limited">1</field>
+                    </depends>
+                    <comment><![CDATA[The number of suggested search terms that will be used for fetching results in the autocompletion box.]]></comment>
```
rbayet, line 214: + "maximum" ?
ie "The maximum number of suggested [...]".

src/module-elasticsuite-core/etc/adminhtml/system.xml
```diff
@@ -180,6 +180,48 @@
...
+                </field>
+                <field id="stop_extension_on_match" translate="label comment" type="select" sortOrder="40" showInDefault="1" showInWebsite="1" showInStore="1" canRestore="1">
+                    <label>Stop the extension mechanism when the current search term of the user is suggested.</label>
```
rbayet, line 217: Something shorter would be nice, like "No extension for actual popular search terms."

src/module-elasticsuite-core/etc/adminhtml/system.xml
```diff
@@ -180,6 +180,48 @@
...
+                    <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
+                    <depends>
+                        <field id="extension_enabled">1</field>
+                    </depends>
+                    <comment><![CDATA[Default: No. When set to "Yes", this setting will discard the extension mechanism as soon as the exact search term entered by the user is part of the suggestions. Eg : When the user has finished to type "computer", if the list of suggested search terms is "computer", "apple computer", "computer for kids", only "computer" will be considered as it's exactly what the user has written.]]></comment>
```
rbayet, line 222: I would streamline the phrase like :
When set to "Yes", the extension mechanism will be discarded when search term entered by the user is amongst the popular terms suggestions. Eg : When the user has finished typing "computer", if the list of suggested search terms is ("computer", "apple computer", "computer for kids"), only "computer" will be taken into account for the products and categories autocomplete.

PR #3541, opened 2025-02-24
Implementing Category permissions

### LINKED ISSUE #2861: Field category_permissions_field does not exists in mapping
Hi,
After Magento Upgrade we found that public graphql query for products is failing 

HTTP Status Code is 200 but response says "message": "Internal server error".

After digging into the logs we found such an error:

`Field category_permissions_field does not exists in mapping`
 

**### Preconditions**

We just deployed Magento EE Upgrade 2.4.5-p1  to production on Magento Cloud and switched from elasticsearch to Opensearch 1.2

**Magento Version** : 2.4.5-p1 EE
**ElasticSuite Version** : 2.10.13.3
**Environmen**t : Production

romainruaud:
Fix #2861

### DIFF
--- a/src/module-elasticsuite-catalog/Model/CategoryPermissions/Filter/Provider.php
+++ b/src/module-elasticsuite-catalog/Model/CategoryPermissions/Filter/Provider.php
@@ -1,0 +1,80 @@
+<?php
+/**
+ * DISCLAIMER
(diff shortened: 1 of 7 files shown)


rbayet:
Replaced by rebased version #3566

PR #3549, opened 2025-03-04
Update default getNumberOfReplicas

ilnytskyi:
Our DevOps team reported that many indexes don't have proper number of replicas despite setting correct default values in config. 

It looks like during index creation the values is 0.

This PR changed default number_of_replicas to be read from system config.

### DIFF
--- a/src/module-elasticsuite-core/Index/IndexSettings.php
+++ b/src/module-elasticsuite-core/Index/IndexSettings.php
@@ -152,7 +152,7 @@ class IndexSettings implements IndexSettingsInterface
     {
         $settings = [
             'requests.cache.enable'            => true,
-            'number_of_replicas'               => 0,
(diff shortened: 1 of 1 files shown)


rbayet:
Hello @ilnytskyi,

This is done on purpose. The configured number of replicas is applied once the index is installed, see 
* https://github.com/Smile-SA/elasticsuite/blob/2.11.x/src/module-elasticsuite-core/Index/IndexOperation.php#L203
* https://github.com/Smile-SA/elasticsuite/blob/a02733b0f739f65f93e3bd747a3cb4e3e6ecb9a3/src/module-elasticsuite-core/Index/IndexSettings.php#L171

So my first hunch is that your indices are not installed properly and end up as ghosts.
But your mileage may vary, so please create an issue with as many details as possible concerning your installation (not forgetting the flavour and version of your Elasticsearch/OpenSearch server).

Regards,

PR #3562, opened 2025-03-18
[CatalogOptimizer] Scale function/factor dependant products exclusion

rbayet:
Conditional low attribute value exclusion


rbayet:
**English version**
image

**French version**
image

Issue #3576, opened 2025-04-09
Index Issue when MSI Inventory is Disabled | Magento 2.4.8


mahesh-makwana-web-vision:
Hello  @vahonc 

**Description**

We are encountering an issue with the ElasticSuite indexing process when the Magento MSI (Multi-Source Inventory) modules are disabled.

Smile Elasticsuite Version: 2.11.12.1
Magento Version: 2.4.8

**Issue Summary:**

The following file attempts to use classes from the Magento_InventoryIndexer module, which causes an error when MSI is disabled and its modules have been removed:

File:
`vendor/smile/elasticsuite/src/module-elasticsuite-catalog/Model/ResourceModel/Product/Indexer/Fulltext/Datasource/InventoryData.php
`
Problematic Dependencies:

```
use Magento\InventoryIndexer\Model\StockIndexTableNameResolverInterface;
use Magento\InventoryIndexer\Indexer\IndexStructure;

```
These classes are part of Magento_InventoryIndexer, which is not available in our installation because MSI is disabled and the related modules were removed.

Error Log:

```
[2025-04-09T09:10:29.919343+00:00] main.CRITICAL: Cannot instantiate interface Magento\InventoryIndexer\Model\StockIndexTableNameResolverInterface [] []
Expected Behavior:

Catalog Rule Product index process error during indexation process:
Cannot instantiate interface Magento\InventoryIndexer\Model\StockIndexTableNameResolverInterface

Catalog Product Rule index process error during indexation process:
Cannot instantiate interface Magento\InventoryIndexer\Model\StockIndexTableNameResolverInterface

```
ElasticSuite should gracefully handle the absence of MSI modules or provide a fallback mechanism when Magento_InventoryIndexer is not available.

**Steps to Reproduce:**

1) Disable and remove MSI-related modules / composer require yireo/magento2-replace-inventory

2) Run full reindex via command.

3) Observe critical error due to missing Magento\InventoryIndexer interfaces.

Thanks

vahonc:
Hello @mahesh-makwana-web-vision,

We integrated this in the core due to the official statement of Adobe back in time which was: "MSI will replace the legacy CatalogInventory in an upcoming release". So we decided to anticipate this and 5 years later, CatalogInventory is still there and the entire Magento community is removing MSI when building a project :)

You can use this patch to remove MSI from Elasticsuite. Not the nicest way but it works.
```
--- src/module-elasticsuite-catalog/etc/di.xml
[...]

Issue #3416, opened 2024-10-21
Move MSI support to separate module / package


jean-bernard-valentaten:
**Is your feature request related to a problem? Please describe.**

A lot of Magento shops do not require the use of MSI (multi stock inventory) and could thus speed up the installation / deployment processes, reduce the overall disk consumption and improve performance by removing the whole MSI feature (roughly 78 packages).

Yireo provides a neat meta-package that helps removing MSI via the `replace` feature of `composer`, as can be seen here: https://github.com/yireo/magento2-replace-inventory

Unfortunately this does not work well in combination with the Smile Elasticsuite, since the latter requires the following MSI packages in its root `composer.json`:

* `magento/module-inventory-sales-api`
* `magento/module-inventory-sales`
* `magento/module-inventory-indexer`

These packages in turn have dependencies to other MSI packages, making it extremely hard to replace MSI when a shop does not actually need it. And since not all MSI packages can be removed the benefit is not as big as it could be.

**Describe the solution you'd like**

I'd suggest to move MSI support to a Smile package of its own. Said package can then be replaced in case MSI is not being used and thus the whole MSI overhead can be removed from shops. Also Yireo could be notified of the existance of said package and add it to the meta-package mentioned above, making it fully compatible to a rather usual setup (Magento + Smile Elasticsuite - MSI).

**Describe alternatives you've considered**

None

**Additional context**

The following packages make use of MSI classes / interfaces:

* `smile/module-elasticsuite-catalog`
* `smile/module-elasticsuite-catalog-rule`

danielraab:
I am looking forward for this feature too.

romainruaud:
Hi guys,

we might consider doing this in a near future or to provide at least a way to have a more loose coupling on MSI.

We integrated this in the core due to the official statement of Adobe back in time which was : "MSI will replace the legacy CatalogInventory in an upcoming release". So we decided to anticipate this and 5 years later, CatalogInventory is still there and the entire Magento community is removing MSI when building a project :)

rbayet:
The content of the PR #3577 is now available in the published releases [2.10.28](https://github.com/Smile-SA/elasticsuite/releases/tag/2.10.28) and [2.11.13](https://github.com/Smile-SA/elasticsuite/releases/tag/2.11.13),

Closing this issue as solved.

PR #3577, opened 2025-04-09
Allow to use ElasticSuite without MSI

romainruaud:
Following the discussions in #3576 and #3416 (and plenty of other issues), here is my proposal to remove our adherence with MSI modules, but still allow them to be used as inventory source when indexing, if they are here.

This should be transparent actually : 

- if you removed MSI (with the yireo/magento2-replace-inventory package, or manually), the indexing phase is using legacy Inventory, and the di:compile does not fail.

- if you are using the MSI modules, the indexing phase will use the appropriate classes.

I let you fellow users give a try to this, before it can land in an upcoming release.


rikwillems:
@romainruaud thank you for looking into this. I replaced my hacky patch with this PR and everything works smooth. No issues in DI compile and indexing.

romainruaud:
Beware, I was not using `ObjectManagerInterface` but the `ObjectManager` directly in the first version of this PR and it was causing a bit of strange dependency issues on our CI/CD process, so I just updated the code to use the interface. If you copy pasted the code somewhere, don't forget to update this :)

Regards

romainruaud:
I let @rbayet review this

rbayet:
Failed Integration tests are purely due to the fact the branch is outdated and does not support OpenSearch 2.x yet.
PR #3590 says it's ok.

rbayet (approved):

Issue #3583, opened 2025-04-16
[Healthcheck] Check default value of category.is_anchor (should be category.is_anchor = 1)


rbayet:
**Is your feature request related to a problem? Please describe.**
For the time being, Elasticsuite relies on the attribute is_anchor of categories to be set at 1.
That's why, when Elasticsuite is installed 
* the default value of the attribute is set to 1
* for all existing categories, the is_anchor is set to 1* 
* the attribute frontend_input is changed to 'hidden' to prevent categories from being switched to "is_anchor" = 0

But sometimes, a few categories could slip through the cracks due to import mechanism, etc and with "strange" behavior happening.

**Describe the solution you'd like**
Add a dual healtcheck that checks that 
* is_anchor attribute 
  * The default value of the is_anchor category attribute is still at 1
  * Its frontend input is still correct at 'hidden'
  * if both conditions are met, check is successful
* categories
  * look for / list for categories (with id <> 1) with is_anchor = 0
     * if none are present, check is successful

**Common text**
* For showing products in categories, Elasticsuite relies on making sure all categories are "anchor" categories, that is displaying both their "own" products and the products associated with their subcategories.

**Test on attribute successful**
[Nothing]

**Test on attribute failed** 
* If the default value of the attribute has been changed
  The default value of the category attribute "Is Anchor (is_anchor)" is no longer set to 1 / Yes. This means all future categories created on the site will no longer automatically be "anchor" categories. 
* If the frontend input is no longer 'hidden'
  The category attribute "Is Anchor (is_anchor)" is no longer hidden on the Admin category edit page. This means the admin users can change categories to no longer being "anchor" categories.
* If any of the above are true, display the message 
  This may cause some issues in the catalog navigation and search results pages.

**Tests on categories successful**
All categories are correctly configured as "anchor" categories.

**Tests on categories failed (at least one category has is_anchor = 0)**
Some of your categories are not configured as "anchor" categories, this may cause some unexpected issues in the catalog navigation and search result pages.

**If the number is lower than or equal to 5 categories**
This is the list of categories : [list of category names with link to their category edit page].

**If the number is greater than 5 categories**
Here are the first 5 categories : [list of category names with link to their category edit page]. There are X more categories in that situation.
(X being Total - 5).

**Update** Severity if any check fails : Critical

romainruaud:
Merged so I close

PR #3594, opened 2025-05-05
[Healthcheck] Feature #3583, introduce category.is_anchor config check

vahonc:


rbayet:
Hello @vahonc,

This sections of the feature is not correctly implemented when the configuration of the attribute is not valid 
> Test on attribute failed
>     If the default value of the attribute has been changed
>          The default value of the category attribute "Is Anchor (is_anchor)" is no longer set to 1 / Yes. This means all future categories created on the site will no longer automatically be "anchor" categories.
>     If the frontend input is no longer 'hidden'
>          The category attribute "Is Anchor (is_anchor)" is no longer hidden on the Admin category edit page. This means the admin users can change categories to no longer being "anchor" categories.

At the moment, you only add the common text, ie "This may cause some issues [...] search results pages".

vahonc:
Fixed.

src/module-elasticsuite-catalog/Model/Healthcheck/CategoryIsAnchorConfigCheck.php
```diff
@@ -0,0 +1,290 @@
...
+                        $defaultText,
+                        __(
+                            'Some of your categories are not configured as "anchor" categories, this may cause some unexpected issues in the catalog navigation and search result pages.'
+                        ),
+                        __(
+                            'This is the list of categories: %1',
+                            [$itemsList]
+                        ),
+                    ]
+                );
```
rbayet, line 125: Why returning here ?
Now you can't have both messages for categories with invalid attribute value and the attribute not having the correct default value ...

src/module-elasticsuite-catalog/Model/Healthcheck/CategoryIsAnchorConfigCheck.php
```diff
@@ -0,0 +1,290 @@
...
+                    __(
+                        'Some of your categories are not configured as "anchor" categories, this may cause some unexpected issues in the catalog navigation and search result pages.'
+                    ),
+                    __(
+                        'Here are the first 5 categories: %1',
+                        [$itemsList]
+                    ),
+                ]
+            ) . __('There are %1 more categories in that situation.', $remainingCount);
+            // @codingStandardsIgnoreEnd
```
rbayet, line 147: Same as above ...

rbayet (changes requested):

src/module-elasticsuite-catalog/Model/Healthcheck/CategoryIsAnchorConfigCheck.php
vahonc: Fixed.

src/module-elasticsuite-catalog/Model/Healthcheck/CategoryIsAnchorConfigCheck.php
vahonc: Fixed.

rbayet (approved):

Issue #3624, opened 2025-06-05
[Indices] Provide a cli command to remove all ghost indices


rbayet:
**Is your feature request related to a problem? Please describe.**
To address the feature request #3600, the PR #3612 adds a button in the admin interface.
The issue is that if the user has a lot of ghost indices (> 100), as the deletion process is not run in asynchronous process, the page could timeout.

**Describe the solution you'd like**
We might improve that in the future but in the meantime 
1)  refactor the RemoveGhostIndices to rely on a model/service what will actually do the removal 
You should move 
https://github.com/Smile-SA/elasticsuite/blob/f574a662ef39a4e754708773ec5e271493a7c9bf/src/module-elasticsuite-indices/Controller/Adminhtml/Index/RemoveGhostIndices.php#L76-L89
and the method (without the $this->_isAllowed() check in the boolean expression)
https://github.com/Smile-SA/elasticsuite/blob/f574a662ef39a4e754708773ec5e271493a7c9bf/src/module-elasticsuite-indices/Controller/Adminhtml/Index/RemoveGhostIndices.php#L108-L112
into that service which will simply return the number of deleted indices

* add a CLI command elasticsuite:ghostindices:purge which will also rely on that model/service

rbayet:
Implemented in #3629

PR #3627, opened 2025-06-05
[Indices] Feature #3624, add a cli command to remove all ghost indices

vahonc:

PR #3628, opened 2025-06-06
[Indices] Feature #3624, add a cli command to remove all ghost indices

rbayet:


rbayet:
Cherry-pick on a new branch of #3627 does not seem enough to unblock the CI...

PR #3629, opened 2025-06-06
[Indices] Feature #3624, add a cli command to remove all ghost indices

rbayet:
+ fixed DI (use of proxy)
+ renamed command as elasticsuite:indices:purgeghosts


rbayet:
Replaces #3627

PR #3631, opened 2025-06-14
Filter only InStock child products in the products index.

### LINKED ISSUE #3630: Incorrect results when filtering by the attribute of an out-of-stock child products.
When Magento Catalog is configured to show only InStock products and we have configurable products (in-stock) and some of the children are out-of-stock issue comes on the listing page with active filters.

Magento Version: >=2.4.7-p4 

ElasticSuite Version: >=2.11.10 (All)

Environment: Both

Third party modules: Not related.

### Steps to reproduce
1. Set Store - Configuration - Catalog - Inventory - Display Out of Stock Products -> No ( php bin/magento config:set cataloginventory/options/show_out_of_stock 0 )
2. Pick-up some configurable product from the catalog and set one of his children as out-of-stock
3. Reindex

### Expected result

1. Image

### Actual result

1. Image

Bashev:
Fix #3630


Bashev:
This should cover all cases, but if I’ve overlooked anything, please let me know. 

Btw, I’m uncertain how to properly address the PHPMD test.

vahonc:
Hot fix: just to add `@SuppressWarnings(PHPMD.CouplingBetweenObjects)` into the class PHPDoc

Long story: The class has a coupling value of 14, exceeding the recommended limit of 13, making it harder to understand, maintain, and test. Therefore, it's need to refactor the class to reduce the number of dependencies on other classes.

src/module-elasticsuite-catalog/Model/ResourceModel/Product/Indexer/Fulltext/Datasource/AttributeData.php
```diff
@@ -229,6 +245,18 @@ private function getRelationQuery($relation, $parentIds, $storeId)
...
+         * then exclude this children from the query results.
+         */
+        if (!$this->scopeConfig->getValue(Configuration::XML_PATH_SHOW_OUT_OF_STOCK)) {
+            $select->joinInner(
+                ['stock' => $inventoryTable],
+                new \Zend_Db_Expr("child.{$entityIdField} = stock.product_id AND stock.is_in_stock = 1"),
+                []
+            );
+        }
+
```
rbayet, line 259: Hello @Bashev,

Wouldn't this piece of code only do what it is intended to do if 
* not using MSI
* using MSI but the current storeId corresponds to a store using the default Stock ?

Regards,

Bashev: Probably,
to be honest, i not test it with MSI which have multiple sources and stocks.

I'll try to check these scenarios and will come back.

Bashev:
Updated with MSI support.

Issue #2766, opened 2022-10-27
The ability to apply optimizer to every category except some.


Hexmage:
**Is your feature request related to a problem? Please describe.**
Currently it's only possible to apply optimizer to all all categories or some selected categories. I currently have a client who has over 1000 categories and wants to apply a booster to all categories except a couple of specific ones. Currently you'd have to select 999+ categories by hand to exclude it from one.

**Describe the solution you'd like**
![Screenshot from 2022-10-27 15-39-28](https://user-images.githubusercontent.com/22170051/198300188-45fb18c7-bd2d-4da8-9ea5-164395ce4c24.png)
A `All but Selected Categories` in this dropdown.

**Additional context**
The currently implementation doesn't easily allow for this logic to be implemented.

Hexmage:
@romainruaud I see that this was given a priority tag. Is there any indication of a timeline?

romainruaud:
@vahonc can you work on this ?

1. Add a "All but Selected Categories" in this dropdown, selecting it should display the Category dropdown. The value associated to this option should be "2" and choosing it should display the Category selector : https://github.com/Smile-SA/elasticsuite/blob/2.11.x/src/module-elasticsuite-catalog-optimizer/view/adminhtml/ui_component/smile_elasticsuite_catalog_optimizer_form.xml#L560

2. Save the data in the appropriate table. The data should contain : 

- "apply_to = 2" for the searchContainer part (this should work automatically here : https://github.com/Smile-SA/elasticsuite/blob/2.11.x/src/module-elasticsuite-catalog-optimizer/Model/ResourceModel/Optimizer.php#L166 )
- and the category_ids should be saved as well without needing to change https://github.com/Smile-SA/elasticsuite/blob/2.11.x/src/module-elasticsuite-catalog-optimizer/Model/ResourceModel/Optimizer/Limitation.php#L85. 

3. Ensure the boosts are properly applied in Front-Office on all categories except the selected. You should probably add something like "OR (osc.apply_to = 2 AND main_table.{$column} != ?)" to this line : https://github.com/Smile-SA/elasticsuite/blob/2.11.x/src/module-elasticsuite-catalog-optimizer/Model/ResourceModel/Optimizer/Limitation.php#L137

[...]

PR #3343, opened 2024-08-07
[Optimizer] Add ability to apply optimizer to all categories except some

vahonc:

PR #3656, opened 2025-07-08
[Optimizer] Feature #2766 Add ability to apply optimizer to all categories except some

rbayet:


rbayet:
Replaces #3343

src/module-elasticsuite-catalog-optimizer/Model/Optimizer/Limitation/Identities.php
```diff
@@ -153,7 +153,11 @@ private function getCategoryIdentities()
 
         $wasAppliedToAllCategories = empty($origData['catalog_view_container']['category_ids']);
 
-        if ($isAppliedToAllCategories || $wasAppliedToAllCategories) {
+        // If the optimizer is or was previously assigned to all but some categories.
+        $appliedToAllCategoriesButSome = ((int) ($data['catalog_view_container']['apply_to'] ?? 0) === 2)
+            || ((int) ($origData['catalog_view_container']['apply_to'] ?? 0) === 2);
+
+        if ($isAppliedToAllCategories || $wasAppliedToAllCategories || $appliedToAllCategoriesButSome) {
```
rbayet, line 160: @romainruaud That's the logical thing to do BUT there is something fishy in 2.4.5 (and maybe above) + Luma where the cat_c tag is actually **not** part of the (FPC) cache tags for a category page.
So in effect, cache does not seem to be really purged.

PR #3667, opened 2025-07-17
[Company] Move company related logic into a helper to make it reusable

PierreGauthier:


rbayet (approved):

rbayet:
There is a slight issue but that was present in the legacy code : in theory, the Company feature can be enabled/disabled at will.
We'll discuss that later with @romainruaud

PR #3692, opened 2025-08-13
[Core] Fix #3623, add supporting getMinPrice() and getMaxPrice() as Magento Core functionality

### LINKED ISSUE #3623: Magento Core getMinPrice() or getMaxPrice() doesn't work with ElasticSuite
The problem occurs because of the Smile ElasticSuite module.

In the vendor/smile/elasticsuite/src/module-elasticsuite-catalog/Model/ResourceModel/Product/Fulltext/Collection.php file, the _renderFiltersBefore method has been overwritten, which resets the filters in the request.

Then, the collection is filtered only by the IDs returned by Elasticsearch.

As a result, when calling the getMinPrice() or getMaxPrice() methods, the returned values ​​do not reflect the actual minimum/maximum prices in the original collection, but only in the subset of products returned by Elasticsearch.

This breaks the core functionality of Magento for determining the min/max price.

On a category page, this problem does not occur with the price range filter (priceRange), since ElasticSuite implements it in an alternative way with a waste of pennies, without using the standard Magento methods for getMinPrice and getMaxPrice.

The goal is to make your module work like this Magento Core functionality.

PierreGauthier:
This PR is the complete version of #3639.
I implemented Richard’s suggestions.
However, I wonder why we chose to create a new search request.
Maybe it would be better to add the aggregation to the main request and run it when we call _prepareStatisticsData ?

### DIFF
--- a/src/module-elasticsuite-catalog/Model/ResourceModel/Product/Fulltext/Collection.php
+++ b/src/module-elasticsuite-catalog/Model/ResourceModel/Product/Fulltext/Collection.php
@@ -13,9 +13,13 @@
  */
 namespace Smile\ElasticsuiteCatalog\Model\ResourceModel\Product\Fulltext;
 
+use Magento\Customer\Model\Group as CustomerGroup;
 use Smile\ElasticsuiteCatalog\Model\Search\Request\Field\Mapper as RequestFieldMapper;
 use Smile\ElasticsuiteCore\Search\Adapter\Elasticsuite\Response\QueryResponse;
+use Smile\ElasticsuiteCore\Search\Request\Aggregation\Bucket\AbstractBucket;
 use Smile\ElasticsuiteCore\Search\Request\BucketInterface;
+use Smile\ElasticsuiteCore\Search\Request\MetricInterface;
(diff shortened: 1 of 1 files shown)


rbayet:
> However, I wonder why we chose to create a new search request. Maybe it would be better to add the aggregation to the main request and run it when we call _prepareStatisticsData ?

You mean by adding the aggregations to the "static" ones defined in the elasticsuite_search_request.xml config as 
* `<aggregation xsi:type="termBucket" name="attribute_set_id" field="attribute_set_id"/>`
* `<aggregation xsi:type="termBucket" name="indexed_attributes" field="indexed_attributes"/>` ?

rbayet:
Something's fishy with multi currencies setups...
But I'm not sure it's 100% Elasticsuite related and definitely NOT specific to this PR...

rbayet:
> ~~Something's fishy with multi currencies setups... But I'm not sure it's 100% Elasticsuite related and definitely NOT specific to this PR...~~

False alarm, it's fixed after cache clearing and a global full reindex.

rbayet (approved):

PR #3698, opened 2025-08-29
[Catalog] Feature #3464, migrate legacy product positions in categories to ElasticSuite

### LINKED ISSUE #3464: Pull previously set sort order of category product when onboarding solution
**Is your feature request related to a problem? Please describe.**
One of our clients set the product sort order for a category with core solution. After Elasticsuite was onboarded, the previously set product order was not applied to the given category.

**Describe the solution you'd like**
Data patch to automatically convert the previous positioning to be reflected on category pages after Elasticsuite is onboarded

**Describe alternatives you've considered**
Magento CLI to pull through previously set sort order of product for a given category (with confirmation question)
eg
```
php bin/magento elasticsuite:category-position:migrate --category [id]
```

**Additional context**
issue resolved for the category by running 
```
INSERT INTO `smile_virtualcategory_catalog_category_product_position` (`category_id`, `product_id`, `store_id`, `position`) SELECT `category_id`, `product_id`, 0 AS `store_id`, `position` + 1 AS `position` FROM `catalog_category_product` WHERE `category_id` = 1095 ORDER BY `position` ASC; 
```

vahonc:


src/module-elasticsuite-catalog/Console/Command/CategoryPositionMigrate.php
```diff
@@ -0,0 +1,402 @@
...
+                    continue;
+                }
+            }
+            $normalized[$newPos] = $pid;
+            $newPos++;
+        }
+
+        // Step 3: persist or preview.
+        if ($dryRun) {
+            $output->writeln("<comment>[Dry-Run]</comment> Category {$categoryId}: would migrate " . count($normalized) . " products.");
```
romainruaud, line 327: It's not "products" but "products positions".

src/module-elasticsuite-catalog/Console/Command/CategoryPositionMigrate.php
```diff
@@ -0,0 +1,402 @@
...
+            if (!empty($toDelete)) {
+                $output->writeln("<comment>[Dry-Run]</comment> Would delete " . count($toDelete) . " products due to conflicts.");
```
romainruaud, line 329: It's not "products" but "products positions".

src/module-elasticsuite-catalog/Console/Command/CategoryPositionMigrate.php
```diff
@@ -0,0 +1,402 @@
...
+        }
+
+        $connection->beginTransaction();
+        try {
+            if (!empty($toDelete)) {
+                $connection->delete(
+                    $table,
+                    ['category_id = ?' => $categoryId, 'product_id IN (?)' => $toDelete]
+                );
+                $output->writeln("<comment>Deleted " . count($toDelete) . " products due to conflicts.</comment>");
```
romainruaud, line 342: It's not "products" but "products positions".

src/module-elasticsuite-catalog/Console/Command/CategoryPositionMigrate.php
```diff
@@ -0,0 +1,402 @@
...
+                $connection->update(
+                    $table,
+                    ['position' => $i],
+                    ['category_id = ?' => $categoryId, 'product_id = ?' => $pid]
+                );
+                $i++;
+            }
+
+            $connection->commit();
+            $output->writeln("<info>Category {$categoryId}: migrated " . count($normalized) . " products.</info>");
```
romainruaud, line 356: It's not "products" but "products positions".

romainruaud (changes requested):
The overall logic looks OK for me (however, I did not test it), but you can proceed moving some logics into dedicated models classes.

Regards

src/module-elasticsuite-catalog/Model/CategoryPositionMigrator.php
```diff
@@ -0,0 +1,351 @@
...
+        bool $migratePositive,
+        string $conflictStrategy
+    ): array {
+        $conflictStrategy = ($conflictStrategy === 'delete') ? 'delete' : 'reorder';
+
+        $legacyTable = $this->categoryProductResource->getMainTable();
+        $targetTable = $this->resource->getTableName(self::TARGET_TABLE_NAME);
+
+        // Fetch all legacy positions.
+        $select = $this->connection->select()
```
PierreGauthier: You should add pagination to this query, as some categories can contain a large number of products, and it may not always be possible to handle them in a single query.

In addition, you should add a condition on visible products, as it does not make sense to store positions for items that are not individually visible (for example, children of configurables). This can also significantly reduce the amount of data to insert.

PierreGauthier (changes requested):

src/module-elasticsuite-catalog/Model/CategoryPositionMigrator.php
```diff
@@ -0,0 +1,351 @@
...
+        $seenPositions = [];
+        $toDelete = [];
+        $nextPos = 1;
+
+        foreach ($updates as $u) {
+            $pid = $u['product_id'];
+            $pos = $u['position'];
+
+            if (isset($seenPositions[$pos])) {
+                if ($conflictStrategy === 'delete') {
```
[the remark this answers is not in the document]
vahonc: I'm not sure that it's a good way.
 
Reason:
- GROUP BY would arbitrarily collapse duplicate rows.
- We would lose control over which product is preserved and which is removed.
- It would depend on database internal behavior (non-deterministic without strict aggregation).
- It does not integrate cleanly with pagination and sequential normalization logic.

My current approach:
- Processes rows in sorted order.
- Keeps the first occurrence.
- Deletes only subsequent duplicates.
- Maintains deterministic and predictable behavior.

PierreGauthier (approved):

PR #3708, opened 2025-09-17
[Thesaurus] Allow using stemmed words to match rules

rbayet:


rbayet:
@romainruaud So enabled by default ?

PR #3767, opened 2025-12-04
[Thesaurus] Manage stemming in multi-word thesaurus (only)

rbayet:


rbayet:
@PierreGauthier for reference, sample data

image

You will need Elasticsuite Premium PR 410 for the Explain screen to work if need be.

PR #3789, opened 2026-02-06
[Indexing] Fixes #3685 - Prevent fallback on default store value when…

### LINKED ISSUE #3685: Issue with website-scoped attribute indexing (e.g., special_price) in multi-store environment
### Preconditions

We have encountered an issue with attribute indexing in Smile_ElasticsuiteCatalog related to website-scoped attributes (e.g., special_price), in a multi-store Magento 2 setup..

❗ Problem:

We had two websites: US and CA

If an attribute like special_price is defined with website scope, and:
	•	The value is not set/empty for a specific store view (e.g., canada_en),
	•	And the “Use Default Value” checkbox is not selected in the admin panel,

Then the attribute value from the default store view (All Store Views) is still indexed(inherited) by ElasticSuite for CA. Although it is indicated as empty for CA.

This results in incorrect values being returned in the search index and can lead to:
	•	Products incorrectly appearing in virtual categories,
	•	Filters and sorting behaving unexpectedly,
	•	Inconsistent behavior across store views.

Magento Version : 2.4.7-p6 Cloud EE

ElasticSuite Version : 2.11.14

Environment : Production/Developer

Third party modules :

### Steps to reproduce

	1.	Set special_price = 1300 for All Store Views (default).
	2.	In store view canada_en, leave special_price blank and do not check “Use Default Value”.
	3.	Reindex and inspect the Elasticsearch document for the product.
	4.	You’ll see special_price = 1300 still present in the index — but it should be omitted/null.

### Expected result

If the store view (e.g., canada_en) has no value for a website-scoped attribute and “Use Default” is not selected, then the attribute should be treated as null/empty and must not fall back to the default store view’s value.

This is consistent with Magento’s frontend and pricing logic.

### Actual result

Products incorrectly appearing in virtual categories

PierreGauthier:
… attribute store value is set to null

### DIFF
--- a/src/module-elasticsuite-catalog/Model/ResourceModel/Eav/Indexer/Fulltext/Datasource/AbstractAttributeData.php
+++ b/src/module-elasticsuite-catalog/Model/ResourceModel/Eav/Indexer/Fulltext/Datasource/AbstractAttributeData.php
@@ -140,12 +140,16 @@ class AbstractAttributeData extends Indexer
                     ['attribute_id', 'attribute_code']
                 )
                 ->where("entity.{$entityIdField} IN (?)", $entityIds)
-                ->where("t_attribute.attribute_id IN (?)", $attributeIds)
-                ->where("t_attribute.value IS NOT NULL");
+                ->where("t_attribute.attribute_id IN (?)", $attributeIds);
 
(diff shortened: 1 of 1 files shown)


rbayet:
Rebase required.

PR #3791, opened 2026-02-10
[Thesaurus] Do not always enforce SPELLING_TYPE_EXACT for rewritten queries

rbayet:


rbayet:
> 
> Second part: for the original query part, we might want to do the same, but I must admit I don't understand that legacy logic behind enforcing SPELLING_TYPE_EXACT for the original query. I get the flawed legacy for the rewritten queries : "how, if we have a synonym, then that means the rewritten query has an exact word" (ignoring the case of multi-terms queries).

After internal discussion, we'll keep the logic of switching the original query to SPELLING_TYPE_EXACT whatever the number of terms.
The original idea was probably (and especially for single term query) : 
* I'm searching for X that does not exist in the engine, so the term vector will tell me to enable fuzziness
* But I have a synonym rule, so I'll search also for Y
* And since I don't want the results for synonym Y to be polluted by fuzzy matches for X, I'll force the search for X to be EXACT

That original idea is also tangentially correct for multiple terms query, so we'll keep things as is.

PR #3817, opened 2026-03-31
Fix virtual category child filtering

szepeviktor:
Virtual category child filtering assumed `getChildrenCategories()` always returned a Magento category collection. In stores where the virtual root subtree returns an array of category objects instead, the layer filter fails while building pages such as `/our-brands`.

- **Problem**
  - `Smile\ElasticsuiteVirtualCategory\Model\Layer\Filter\Category::getChildrenCategories()` switched to the virtual root subtree and immediately called collection methods on the result.
  - That breaks when `getChildrenCategories()` returns an array rather than `\Magento\Catalog\Model\ResourceModel\Category\Collection`.

- **Change**
  - Preserve the existing collection path when the virtual root returns a category collection.
  - Add an array fallback that filters children with `array_filter(...)`.
  - Exclude the current category in both paths so subtree navigation behavior stays consistent.

- **Implementation details**
  - Added an explicit `Collection` type check before calling `clear()->addFieldToFilter(...)`.
  - Cached the current category ID once and reused it for both collection and array filtering paths.

### DIFF
--- a/src/module-elasticsuite-virtual-category/Model/Layer/Filter/Category.php
+++ b/src/module-elasticsuite-virtual-category/Model/Layer/Filter/Category.php
@@ -128,10 +128,19 @@ class Category extends \Smile\ElasticsuiteCatalog\Model\Layer\Filter\Category
                 $rootCategory = $this->virtualCategoryRoot->getVirtualCategoryRoot($currentCategory);
                 if ($rootCategory->getId()) {
                     $this->childrenCategories = $rootCategory->getChildrenCategories();
-                    $this->childrenCategories->clear()->addFieldToFilter(
-                        'entity_id',
-                        ['neq' => $currentCategory->getId()]
-                    );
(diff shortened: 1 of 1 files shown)


rbayet:
Hello @szepeviktor,

In order to reproduce it locally, is there a specific third-party module that should be used to provoke such a context where `getChildrenCategories()` would return an array rather than a `\Magento\Catalog\Model\ResourceModel\Category\Collection` ?

Additionally, could you add `@SuppressWarnings(PHPMD.ElseExpression)` to the PHPDoc of the `getChildrenCategories` so that the PHPMD check goes through ?

Regards,

szepeviktor:
> In order to reproduce it locally, is there a specific third-party module that should be used to provoke such a context

I am very sorry, Magento is too complex for me. And we have an ERP feeding crazy-inconsistent data.

Tag added!

PR #3819, opened 2026-04-03
[Tracker] Add child simple order item of configurable product

rbayet:
Useful for comparative performance of configurations or for visible simple children of configurable products


rbayet:
@romainruaud the change from the `$item` from a plain variable to a reference for private method `\Smile\ElasticsuiteTracker\Block\Variables\Page\Order::getOrderItemVariables` is a bit meh... but that's the least intrusive.

romainruaud (approved):

romainruaud:
LGTM you can merge

PR #3833, opened 2026-04-23
Fix condition check for mouseout event

groomershop-mt:
Currently this is causing JS errors:

self._getLastElement().hasClass is not a function. (In 'self._getLastElement().hasClass(self.options.selectClass)', 'self._getLastElement().hasClass' is undefined)

### DIFF
--- a/src/module-elasticsuite-core/view/frontend/web/js/form-mini.js
+++ b/src/module-elasticsuite-core/view/frontend/web/js/form-mini.js
@@ -301,7 +301,7 @@ define([
                                 self._resetResponseList(false);
                             })
                             .on('mouseout', function () {
-                                if (!self._getLastElement() && self._getLastElement().hasClass(self.options.selectClass)) {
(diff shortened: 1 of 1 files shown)


rbayet:
Hello @groomershop-mt,

I'm not able to reproduce the error in any context (because the `mouseleave` event is triggered instead/before ?) but indeed, if I remove all other event handlers except the `mouseenter`), I can see that the event handler is not doing what is supposed to do, that is removing the `selectClass` indicating focus on the last element of the vertical list of all result elements if the mouse is exiting the list, for instance through that last element.

For instance, going from here 
image
to there
image

Considering that our `form.mini.js` is heavily inspired [by the Magento one](https://github.com/magento/magento2/blob/6c67a9b503c5cd2ad414493b37d1aaaa60e3d385/app/code/Magento/Search/view/frontend/web/js/form-mini.js#L348-L352), you could probably also propose a similar PR on their github (for people not using Elasticsuite)

Regards

PR #3837, opened 2026-04-27
Fix category position migrator visibility join for CE/Mage-OS

mrrobot-magedev:
### Description

This PR fixes the visibility join in `CategoryPositionMigrator`.

The current query joins `catalog_product_entity_int` using:

```sql
cpei.row_id = ccp.product_id
```

However, in Magento Open Source / Mage-OS CE, product EAV backend tables use `entity_id`, not `row_id`.

This causes the category position migration command to fail on CE/Mage-OS installations because the column `catalog_product_entity_int.row_id` does not exist.

---

### Fix

Replaced:

```sql
cpei.row_id = ccp.product_id
```

with:

```sql
cpei.entity_id = ccp.product_id
```

---

### Steps to reproduce

1. Install ElasticSuite `2.11.x` on Magento Open Source / Mage-OS CE.
2. Run the category position migration CLI command.
3. Observe SQL error due to missing `row_id` column in `catalog_product_entity_int`.

---

### Expected result

Migration runs successfully and product visibility is correctly joined.

---

### Type of change

- Bug fix

---

### Notes

This change ensures compatibility with Magento Open Source / Mage-OS CE environments.

PR #3838, opened 2026-04-28
Fix category position migrator visibility join for CE/Mage-OS (2.10.x)

rbayet:
using dynamic linkField (CE/Mage-OS + EE compatibility)


rbayet:
Cherry-picks of #3837 from @mrrobot-magedev based on top of 2.10.x

PR #3839, opened 2026-04-29
[RuleEngine] Attribute Locator interface, refactor Optimizer locator with filter pool, update tests

vahonc:


src/module-elasticsuite-catalog-optimizer/Model/Rule/Attribute/OptimizerLocationProvider.php
```diff
@@ -14,22 +14,15 @@
 
 namespace Smile\ElasticsuiteCatalogOptimizer\Model\Rule\Attribute;
 
-use Magento\Framework\App\ResourceConnection;
-use Smile\ElasticsuiteCatalogOptimizer\Api\Data\OptimizerInterface;
 use Smile\ElasticsuiteCatalogRule\Api\Rule\Attribute\LocationProviderInterface;
+use Smile\ElasticsuiteCatalogOptimizer\Model\ResourceModel\Optimizer\CollectionFactory;
+use Smile\ElasticsuiteAbCampaign\Ui\Component\Optimizer\Listing\FilterCollectionNoCampaign;
```
rbayet: You cannot directly reference an Elasticsuite Premium module in the open source code base.
You filter class needs to be agnostic, so either a dummy class that does nothing and is overridden in the AB Campaign module, or an optional Filter class array through an interface

rbayet (changes requested):

src/module-elasticsuite-catalog-optimizer/etc/di.xml
```diff
@@ -153,4 +153,11 @@
         </arguments>
     </type>
 
+    <!-- No collection filters (empty by default) -->
+    <type name="Smile\ElasticsuiteCatalogOptimizer\Model\Rule\Attribute\OptimizerLocationProvider">
+        <arguments>
+            <argument name="filters" xsi:type="array"/>
+        </arguments>
+    </type>
```
rbayet: Please remove this, I don't think it would have an impact, but we don't need to spell it out that loud to people.

rbayet (dismissed):

src/module-elasticsuite-catalog-optimizer/etc/di.xml
vahonc: Done.

PR #3849, opened 2026-05-20
[Tracker] Add defer true to tracker script

vahonc:


rbayet:
Cannot work as is

- the link tag ensures that the tracking.js script is loaded after the DOM is loaded : :heavy_check_mark: 
- but the `<script>` tag that is in the `<head>` and which initializes the `smileTracker` object is executed immediately

It will throw a silent (because of the `try / catch`) error and the tag will never be sent.
If you comment the try / catch structure in `magento/vendor/smile/elasticsuite/src/module-elasticsuite-tracker/view/frontend/templates/config.phtml` 
```
    $scriptString = "\n";
    // $scriptString .= "try{";
    $scriptString .= "smileTracker.setConfig({";
    $scriptString .= "beaconUrl:'{$beaconUrl}',";
    $scriptString .= "telemetryUrl:'{$telemetryUrl}',";
    $scriptString .= "telemetryEnabled:'{$telemetryEnabled}',";
    $scriptString .= "sessionConfig:{$sessionConfig},";
    $scriptString .= "{$trackerApiConfig}";
    $scriptString .= "});";
    $scriptString .= "smileTracker.addPageVar('store_id','{$storeId}');";
    $scriptString .= "require(['{$userConsentScript}'],function(userConsent){";
    $scriptString .= "if(userConsent({$userConsentConfig})){";
    $scriptString .= "smileTracker.sendTag();";
    $scriptString .= "}";
    $scriptString .= "});";
    // $scriptString .= "}catch(err){;}\n";

    echo /* @noEscape */ $secureRenderer->renderTag('script', [], $scriptString, false);
```
you will be able to see the error in the browser console 
```
Uncaught ReferenceError: smileTracker is not defined
    <anonymous> https://ee245p9.docker.localhost/en/women/tops-women/jackets-women.html:39
[jackets-women.html:39:1](https://ee245p9.docker.localhost/en/women/tops-women/jackets-women.html)
    <anonyme> https://ee245p9.docker.localhost/en/women/tops-women/jackets-women.html:39
```

So since the link + script are supposed to be executed in a consecutive manner, either
- both are supposed to be blocking in the head
- or both are supposed to be executed after the DOM

The issue is that it seems that the `defer` attribute is only supposed to be supported for `script` tag with a `src` attribute, ie remote scripts (and not inline scripts).

rbayet (changes requested):
See detailed comment above

rbayet:
@vahonc possible solution : https://dev.to/k4ml/inline-script-defer-4dkf + https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script#defer ?

See also https://gist.github.com/jakub-g/385ee6b41085303a53ad92c7c8afd7a6

I must admit that considering the risks, I would fancy having theses changes as configurable.

rbayet:
Let's put that on the back burner for now.

PR #3860, opened 2026-05-31
Quote column names in Optimizer/Collection.php

wienczny:
Unquoted use of `from_date` in a raw WHERE clause triggers a MariaDB reserved-word syntax error. Quoting both date columns fixes compatibility with strict MariaDB versions

### DIFF
--- a/src/module-elasticsuite-catalog-optimizer/Model/ResourceModel/Optimizer/Collection.php
+++ b/src/module-elasticsuite-catalog-optimizer/Model/ResourceModel/Optimizer/Collection.php
@@ -131,8 +131,8 @@ class Collection extends \Magento\Framework\Model\ResourceModel\Db\Collection\Ab
         }
 
         $this->getSelect()
-            ->where('from_date is null or from_date <= ?', $date)
-            ->where('to_date is null or to_date >= ?', $date);
+            ->where('`from_date` is null or `from_date` <= ?', $date)
+            ->where('`to_date` is null or `to_date` >= ?', $date);
 
         return $this;
     }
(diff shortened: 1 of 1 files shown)


wienczny:
A more magento way to do this could be something like

```
$this->addFieldToFilter('from_date', [['null' => true], ['lteq' => $date]]);
$this->addFieldToFilter('to_date',   [['null' => true], ['gteq' => $date]]);
```

src/module-elasticsuite-catalog-optimizer/Model/ResourceModel/Optimizer/Collection.php
```diff
@@ -131,8 +131,8 @@ public function addIsActiveFilter($date = null)
         }
 
         $this->getSelect()
-            ->where('from_date is null or from_date <= ?', $date)
-            ->where('to_date is null or to_date >= ?', $date);
+            ->where('`from_date` is null or `from_date` <= ?', $date)
+            ->where('`to_date` is null or `to_date` >= ?', $date);
```
rbayet: If we want to be more precise, the proper way would be 
```
        $fromDateCol = $this->getConnection()->quoteIdentifier('from_date');
        $toDateCol = $this->getConnection()->quoteIdentifier('to_date');
        $this->getSelect()
            ->where(sprintf("%s is null or %s <= ?", $fromDateCol, $fromDateCol), $date)
            ->where(sprintf("%s is null or %s >= ?", $toDateCol, $toDateCol), $date);
```
with `$this->getConnection()->quoteIdentifier` being what `addFieldToFilter` does on the field(s)/first param.

rbayet (changes requested):

rbayet:
Hello @wienczny,

Thanks for your PR.
Would you be willing to apply my suggested changes and change the target branch to 2.10.x ?

Regards,

wienczny:
Updated by https://github.com/Smile-SA/elasticsuite/pull/3864

PR #3864, opened 2026-06-02
Quote column names in Optimizer/Collection.php

wienczny:
Unquoted use of from_date in a raw WHERE clause triggers a MariaDB reserved-word syntax error. Quoting both date columns fixes compatibility with strict MariaDB versions Use quoteIdentifier on field name because to_data and from_date fail on a strict mariadb. The sprintf mimics addFieldtoFilter without the overhead.


rbayet:
Thanks @wienczny !

PR #3869, opened 2026-06-15
[RuleEngine] Attributes used in rules monitor - only attribute "used for promo rules"

vahonc:


src/module-elasticsuite-catalog/etc/adminhtml/di.xml
rbayet, line 1: Isn't there a less intrusive way than a plugin on Magento\Framework\Data\Form\Element\AbstractElement ?

rbayet (changes requested):

src/module-elasticsuite-catalog/etc/adminhtml/di.xml
vahonc: I agree that targeting AbstractElement feels intrusive. However, the Product Attribute Edit page still relies on the legacy Block/Form rendering system rather than UI Components, so we cannot inject this CSS class via XML configuration.

rbayet: I'll have a look at it

src/module-elasticsuite-catalog-rule/Setup/CatalogRuleSetup.php
rbayet, line 1: I agree that this is the modern version of setup, but I would prefer that we keep going on with our own approach already used ie using `\Smile\ElasticsuiteCatalog\Setup\UpgradeData` + `\Smile\ElasticsuiteCatalog\Setup\CatalogSetup` and bumping the module version

rbayet (changes requested):

src/module-elasticsuite-catalog/etc/adminhtml/di.xml
rbayet: Hello @vahonc,

Please either "hardcode" the required CSS through 
* `src/module-elasticsuite-catalog/view/adminhtml/web/css/source/_module.less`
* or `src/module-elasticsuite-core/view/adminhtml/web/css/es-logo.css`

Or if you want to do something a bit more flexible, re-use the dedicated `\Smile\ElasticsuiteCatalog\Block\Plugin\Adminhtml\Product\Attribute\Edit\Tab\FrontPlugin`.
You have access there to the form and thus the field as en element in `\Smile\ElasticsuiteCatalog\Block\Plugin\Adminhtml\Product\Attribute\Edit\Tab\FrontPlugin::afterSetForm` through 
```
        $useInPromoRule = $form->getElement('is_used_for_promo_rules');
        if ($useInPromoRule) {
          ....
        }
```
Granted, adding the css class through `$useInPromoRule->addClass('es-esfeature__logo');` is probably not the most straight forward way, so you could also try to change the element renderer, for instance.

vahonc: Refactored by introducing a `WrapperCssClassDecorator` instance that decorates renderer of the native field element to append the `es-esfeature__logo` CSS class to its field wrapper.

src/module-elasticsuite-catalog-rule/Setup/CatalogRuleSetup.php
vahonc: Refactored according to the native ElasticSuite way for data setup/upgrade.

PR #3876, opened 2026-06-29
Fix PHP 8.5 deprecated use of null as array key for virtual category …

### LINKED ISSUE #3861: Potential problem with PHP 8.5 and using null as an array offset (which is deprecated)
While working on a Magento upgrade to 2.4.9 and PHP to 8.5 and running the [ampersand upgrade tool](https://github.com/idhlagency/ampersand-magento2-upgrade-patch-helper/) to check for potential issues, I noticed something in this module that could potentially trigger a deprecated warning when using PHP 8.5

It's around this code: https://github.com/Smile-SA/elasticsuite/blob/2.11.19-rc1/src/module-elasticsuite-virtual-category/Plugin/Catalog/Category/DataProviderPlugin.php#L86-L88

See the deprecated warning here on PHP 8.5 (not on older versions): https://3v4l.org/YGn9Z#v

Magento also came with some fallback in case the category id is `null`, diff between 2.4.8-p5 and 2.4.9:
```diff
diff -ur -N vendor_orig/magento/module-catalog/Model/Category/DataProvider.php vendor/magento/module-catalog/Model/Category/DataProvider.php
--- vendor_orig/magento/module-catalog/Model/Category/DataProvider.php	2026-06-01 08:42:39
+++ vendor/magento/module-catalog/Model/Category/DataProvider.php	2026-05-04 10:29:58
@@ -383,7 +383,7 @@
             $categoryData = $this->filterFields($categoryData);
             $categoryData = $this->convertValues($category, $categoryData);
 
-            $this->loadedData[$category->getId()] = $categoryData;
+            $this->loadedData[$category->getId() ?? ''] = $categoryData;
         }
```

### Preconditions
- Magento Version: 2.4.9
- ElasticSuite Version: 2.11.19-rc1
- Environment: N/A
- Third party modules: N/A

### Steps to reproduce
1. No idea if this can be triggered, but since you have an if condition for it, I assume it can?

### Expected result
1. Fully compatible with PHP 8.5

### Actual result
1. Potentially triggering deprecated warning

hostep:
…data provider.

Before PHP 8.5, if using `null` as array key, php silently converted this to an empty string: https://3v4l.org/9jqqo#v

As of PHP 8.5, this results in a deprecation warning, so we now need to do this conversion explicitly ourselves, and since Magento elevates deprecation warnings to real errors we no longer can create a new category in the backoffice of Magento when using PHP 8.5.

See https://github.com/Smile-SA/elasticsuite/issues/3861#issuecomment-4833074459 for steps to reproduce.

Fixes #3861

### DIFF
--- a/src/module-elasticsuite-virtual-category/Plugin/Catalog/Category/DataProviderPlugin.php
+++ b/src/module-elasticsuite-virtual-category/Plugin/Catalog/Category/DataProviderPlugin.php
@@ -82,34 +82,35 @@ class DataProviderPlugin
         $data = $proceed();
 
         $currentCategory = $dataProvider->getCurrentCategory();
+        $currentCategoryId = $currentCategory->getId() ?? '';
 
-        if ($currentCategory->getId() === null || $currentCategory->getLevel() < 2) {
-            $data[$currentCategory->getId()]['use_default']['is_virtual_category'] = true;
+        if ($currentCategoryId === '' || $currentCategory->getLevel() < 2) {
+            $data[$currentCategoryId]['use_default']['is_virtual_category'] = true;
         }
 
-        if ($currentCategory->getLevel() >= 2 && !isset($data[$currentCategory->getId()]['virtual_category_root'])) {
-            $data[$currentCategory->getId()]['virtual_category_root'] = $currentCategory->getPathIds()[1];
+        if ($currentCategory->getLevel() >= 2 && !isset($data[$currentCategoryId]['virtual_category_root'])) {
+            $data[$currentCategoryId]['virtual_category_root'] = $currentCategory->getPathIds()[1];
         }
 
-        $data[$currentCategory->getId()]['use_default']['show_use_store_positions'] = true;
-        if (!$currentCategory->getStoreId() || $currentCategory->getId() === null) {
-            $data[$currentCategory->getId()]['use_default']['show_use_store_positions'] = false;
+        $data[$currentCategoryId]['use_default']['show_use_store_positions'] = true;
+        if (!$currentCategory->getStoreId() || $currentCategoryId === null) {
+            $data[$currentCategoryId]['use_default']['show_use_store_positions'] = false;
         }
 
         // To restore global/"All store views" positions/blacklist.
-        $data[$currentCategory->getId()]['default']['sorted_products'] = [];
-        $data[$currentCategory->getId()]['default']['blacklisted_products'] = [];
+        $data[$currentCategoryId]['default']['sorted_products'] = [];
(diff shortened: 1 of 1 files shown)


src/module-elasticsuite-virtual-category/Plugin/Catalog/Category/DataProviderPlugin.php
```diff
@@ -82,34 +82,35 @@ public function aroundGetData(CategoryDataProvider $dataProvider, \Closure $proc
...
-            $data[$currentCategory->getId()]['virtual_category_root'] = $currentCategory->getPathIds()[1];
+        if ($currentCategory->getLevel() >= 2 && !isset($data[$currentCategoryId]['virtual_category_root'])) {
+            $data[$currentCategoryId]['virtual_category_root'] = $currentCategory->getPathIds()[1];
         }
 
-        $data[$currentCategory->getId()]['use_default']['show_use_store_positions'] = true;
-        if (!$currentCategory->getStoreId() || $currentCategory->getId() === null) {
-            $data[$currentCategory->getId()]['use_default']['show_use_store_positions'] = false;
+        $data[$currentCategoryId]['use_default']['show_use_store_positions'] = true;
+        if (!$currentCategory->getStoreId() || $currentCategoryId === null) {
```
rbayet: `$currentCategoryId === null` will no longer be satisfied since it is now an explicit empty string per line 85.
One needs to apply the same logic as on line 87 and replace the condition.

rbayet (changes requested):

src/module-elasticsuite-virtual-category/Plugin/Catalog/Category/DataProviderPlugin.php
hostep: Thank for spotting this @rbayet, I completely overlooked it. Fixed now.

rbayet:
Hello @hostep,

Out of curiosity, didn't you have initially the same error located in the legacy Magento code in vendor/magento/module-catalog/Model/Category/DataProvider.php on line 421 ?

Regards,

rbayet:
Hello @hostep,

Oh yeah you did, hence [your issue on Magento2 github](https://github.com/magento/magento2/issues/40908), that's related to the "null" frontend_input for "is_virtual_category".

Regards,

hostep:
@rbayet: yes, but these are 2 completely different bugs in my opinion.

rbayet:
Yes, for that specific, I'll try to have to look if we cannot tweak the attributes definition to mitigate that last error while I scan for other source of `deprecated use of null as array key`.

rbayet (approved):

PR #3879, opened 2026-07-03
[Tracker] Allow individual tracking sales of bundle and grouped products

rbayet:


rbayet:
Only product_id, sku, qty and row_total tracked.
Label and categories skipped for performance reasons.

PR #3887, opened 2026-07-20
Improve graphql filtering capabilities

PierreGauthier:


src/module-elasticsuite-catalog-graph-ql/DataProvider/Product/LayeredNavigation/Builder/Category.php
```diff
@@ -123,23 +123,33 @@ public function build(AggregationInterface $aggregation, ?int $storeId): array
...
+            $this->resourceConnection->getConnection()->fetchAll(
+                $this->categoryAttributeQuery->getQuery($categoryIds, ['name', 'is_active'], $storeId)
+            )
         );
+        $categoryLabels = \array_column($categoryAttributesValues, 'name', 'entity_id');
+        $categoryActive = \array_column($categoryAttributesValues, 'is_active', 'entity_id');
+
+        $categoryIds = \array_filter($categoryIds, function ($categoryId) use ($categoryActive) {
+            return !empty($categoryActive[$categoryId] ?? null);
+        });
```
rbayet, line 152: What are those changes related to ?
To also filter on categories that are no longer active ?

rbayet (changes requested):
I would prefer that you split the PR between
* one related to categories which is performance / accuracy oriented
* the other for the improved graphql filtering abilities per se

src/module-elasticsuite-catalog-graph-ql/DataProvider/Product/LayeredNavigation/Builder/Category.php
PierreGauthier: Yes, I use the same "post filter" we use on luma side for category aggregation, but inactive category should not be present il the filter aggregation anyway

PierreGauthier:
Split into : 
- https://github.com/Smile-SA/elasticsuite/pull/3890
- https://github.com/Smile-SA/elasticsuite/pull/3889

PR #3890, opened 2026-07-21
[GraphQl] Fix filtering on text attribute and add negation operator

PierreGauthier:


src/module-elasticsuite-catalog-graph-ql/etc/schema.graphqls
```diff
@@ -1,3 +1,17 @@
...
+}
+
+input FilterMatchTypeInput {
+    not_match: String @doc(description: "Exclure les produits dont le champ contient cette valeur.")
+}
+
+input FilterRangeTypeInput {
+    not_from: String @doc(description: "Exclure les valeurs supérieures ou égales à cette borne.")
+    not_to: String @doc(description: "Exclure les valeurs inférieures ou égales à cette borne.")
+}
```
rbayet: Why use french in the doc ?!

src/module-elasticsuite-core/Index/Mapping/Field.php
```diff
@@ -401,8 +402,8 @@ private function getFieldAnalyzers(): array
             $analyzers = [self::ANALYZER_KEYWORD];
         }
 
-        if ($this->isSearchable() || $this->isUsedForSortBy()) {
-            // Default search analyzer.
+        if ($this->isSearchable() || $this->isUsedForSortBy() || $this->isFilterable()) {
+            // Default search analyzer. Also needed for filterable text fields to support match queries.
```
rbayet, line 406: Have you checked that this doesn't put back pure filterable fields linked to pure filterable attributes back in the scope of the Fulltext Query Builder ?

rbayet (changes requested):

src/module-elasticsuite-core/Index/Mapping/Field.php
PierreGauthier: No, it can't, there is explicit condition on isSearchable to add copy_to in the mapping
and the multimatch query building is protected with le fieldFilter

PR #3898, opened 2026-07-31
[VirtualCategories] Handle null category IDs and broaden exception handling

### LINKED ISSUE #3900: [PHP 8.5.8] Deprecation Error / 404 on Product View
Hi team, 
I couldn't find an issue. Hope I don't buffer you. 🤭

PHP 8.5.8 (8.2 or 8.3)
Mage-OS 3.2.0 / Magento 2.4.9
ElasticSuite 2.12

### Steps to reproduce:

Navigate to a product page via a category page (so category context is active in the current session/request).

Leave the tab open until the server-side PHP session expires (or manually clear server sessions via
```
rm -rf var/session/*
```
Reload the product page directly or paste the product URL in a new browser window.

### Expected result:
The product page loads normally, fallback/default behavior takes over without category context.

### Actual result:
The page returns a 404 Not Found.

In var/log/exception.log, the following PHP 8 deprecation exception is logged:

```
main.CRITICAL: ErrorException: Using null as an array offset is deprecated, use an empty string instead in vendor/mage-os/module-catalog/Model/CategoryRepository.php:153
Stack trace:
#0 vendor/mage-os/module-catalog/Model/CategoryRepository.php(153): Spatie\Ignition\Ignition->renderError()
#1 vendor/smile/elasticsuite/src/module-elasticsuite-virtual-category/Plugin/Catalog/ProductPlugin.php(98): Magento\Catalog\Model\CategoryRepository->get()
#2 vendor/smile/elasticsuite/src/module-elasticsuite-virtual-category/Plugin/Catalog/ProductPlugin.php(98): Smile\ElasticsuiteVirtualCategory\Plugin\Catalog\ProductPlugin->aroundCanBeShowInCategory()
...
```
### Cause:
When the session expires or no category ID is present in the request context, $categoryId passed to aroundCanBeShowInCategory is null.

In Smile\ElasticsuiteVirtualCategory\Plugin\Catalog\ProductPlugin.php,
$this->categoryRepository->get($categoryId)
is called directly without verifying if $categoryId is set/not empty. In PHP 8.2+, passing null to CategoryRepository::get() triggers a deprecation notice (instances[null]), which halts execution and causes Magento to throw a 404 response.

### Proposed Fix:
.../vendor/smile/elasticsuite/src/module-elasticsuite-virtual-category/Plugin/Catalog/ProductPlugin.php
line 98
Wrap the repository call with a check for !empty($categoryId):
```
public function aroundCanBeShowInCategory(Product $product, \Closure $proceed, $categoryId)
{
    if (!empty($categoryId)) {
        try {
            $category = $this->categoryRepository->get($categoryId);
            if ((bool) $category->getIsVirtualCategory() === true) {
                return true;
            }
        } catch (NoSuchEntityException $e) {
            $category = null;
        }
    }

    return $proceed($categoryId);
}
```

Best, So-Egal.

rbayet:


So-Egal:
Hi @rbayet  / Team,

Thanks for providing the fix and updating the catch block to handle general \Exceptions!

Currently, you uses:

```
if (!$categoryId) {
    try {
        $category = $this->categoryRepository->get($categoryId);
        // ...
```
Two potential issues with this approach:

Inverted Logic: Using if (!$categoryId) means the block only executes when $categoryId is null or empty. When a user navigates normally via a category ($categoryId is set, e.g., 12), the check is skipped entirely, which bypasses the virtual category logic for active requests.

Unnecessary Repository Call & Exception Overhead: When $categoryId is null, executing
$this->categoryRepository->get(null) 
will always fail and hit the catch block.

Wouldn't it be cleaner and more performant to check if a valid $categoryId exists before calling the repository, avoiding the exception altogether?

Suggested approach:

```
public function aroundCanBeShowInCategory(Product $product, \Closure $proceed, $categoryId)
{
    if (!empty($categoryId)) {
        try {
            $category = $this->categoryRepository->get($categoryId);
            if ((bool) $category->getIsVirtualCategory() === true) {
                return true;
            }
        } catch (NoSuchEntityException | \Exception $e) {
            $category = null;
        }
    }
return $proceed($categoryId);
}
```


This ensures that:
* Virtual category checks run when a category ID is present.
* No unnecessary calls with `null` hit the `CategoryRepository` when the context/session is empty.

What do you think? Curious to hear your thoughts on this! 🚀

rbayet:
Oh yeah, good catch, I changed the [original fix](https://github.com/Smile-SA/elasticsuite/pull/3874/changes) from @Bashev with an early return `if (!$categoryId)` by conditioning the `try/catch` to the test, but I forgot to reverse the condition:
```
        if (!$categoryId) {
            return $proceed($categoryId);
        }

        try {
            $category = $this->categoryRepository->get($categoryId);
            if ((bool) $category->getIsVirtualCategory() === true) {
                return true;
            }
        } catch (NoSuchEntityException | \Exception $e) {
            $category = null;
        }

        return $proceed($categoryId);
```
to 
```
        if (!$categoryId) {
            try {
                $category = $this->categoryRepository->get($categoryId);
                if ((bool) $category->getIsVirtualCategory() === true) {
                    return true;
                }
            } catch (NoSuchEntityException | \Exception $e) {
                $category = null;
            }

            return $proceed($categoryId);}
```

And the mess up is transparent due to the catch, but yeah, this makes an useless repository load.

Regards,

rbayet:
Fix in #3901

PR #3909, opened 2026-08-19
[CatalogRule] Use a single terms query for SKU "is one of" conditions

Nuranto:
## Problem

`QueryBuilder::getSpecialAttributesSearchQuery()` expands a condition on `sku` into one `match` clause per value, whatever the operator. A rule holding a list of SKUs therefore costs as many Lucene leaf clauses as it has values.

On a production catalog, a virtual category with `sku is one of <3111 SKUs>` produced 3111 clauses. Every category whose query embeds that rule went over `indices.query.bool.max_clause_count` (5416 on that cluster) and silently returned zero products:

```json
{
  "type": "illegal_argument_exception",
  "reason": "Query rewrite failed: too many clauses",
  "caused_by": {
    "type": "too_many_nested_clauses",
    "reason": "maxClauseCount is set to 5416"
  }
}
```

## Steps to reproduce

1. Create a virtual category with a single rule condition `SKU is one of <a few thousand SKUs>`.
2. Browse it, or any category whose subtree references it.
3. The listing is empty and `var/log/system.log` shows the error above.

## Change

`()` and `!()` now emit a single `terms` clause. The `contains` path is left untouched.

`getSearchFieldName()` already resolves those two operators to the untouched field, which is mapped as `keyword` with a no-op normalizer. A `match` with `minimum_should_match: 100%` on a `keyword` field resolves to a single term, so N `match` clauses in a `should` are equivalent to one `terms` of N values.

|        | clauses      | DSL size    |
| ------ | ------------ | ----------- |
| before | 3111 `match` | 272 707 B   |
| after  | 1 `terms`    | 30 089 B    |

## Tests

`QueryBuilderTest`: added `testSpecialAttributesMultipleSkuIsOneOfQuery`, and updated `testSpecialAttributesSkuQuery` / `testSpecialAttributesMultipleSkuQuery` to the new contract. `testSpecialAttributesMultipleSkuAnalyzedQuery` covers the `contains` path and is unchanged.

### DIFF
--- a/src/module-elasticsuite-catalog-rule/Model/Rule/Condition/Product/QueryBuilder.php
+++ b/src/module-elasticsuite-catalog-rule/Model/Rule/Condition/Product/QueryBuilder.php
@@ -146,12 +146,24 @@ class QueryBuilder
 
             $clause = (substr($productCondition->getOperator(), 0, 1) === '!') ? 'mustNot' : 'should';
             // SKU values can be an array of value, due to the picker.
-            foreach (explode(',', $productCondition->getValue()) as $value) {
-                // Instead of just a "should" clause, we will have "should" or "must_not" depending on the rule operator.
+            $values = array_map('trim', explode(',', $productCondition->getValue()));
+
+            if (in_array($productCondition->getOperator(), ['()', '!()'])) {
+                // Those operators run on the untouched (keyword) field, where a single terms clause is
+                // equivalent to one match clause per value, but counts as a single clause in Lucene.
                 $queryParams[$clause][] = $this->prepareQuery(
-                    QueryInterface::TYPE_MATCH,
-                    ['field' => $fieldName, 'queryText' => trim($value), 'minimumShouldMatch' => "100%"]
+                    QueryInterface::TYPE_TERMS,
+                    ['field' => $fieldName, 'values' => $values]
                 );
+            } else {
+                foreach ($values as $value) {
+                    // Instead of just a "should" clause, we will have "should" or "must_not"
+                    // depending on the rule operator.
+                    $queryParams[$clause][] = $this->prepareQuery(
+                        QueryInterface::TYPE_MATCH,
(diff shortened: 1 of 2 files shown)


rbayet:
Hello @Nuranto,

> `sku is one of <3111 SKUs>`

My heart sank. :(
What about adding an ad-hoc boolean attribute and assign at Yes to those 3111 products and have the virtual category rule rely on that attribute instead ?

Anyway, I'll have a look.

Regards,

Nuranto:
Hi @rbayet 

😆 I know, mine too ! I'm the dev, not the user !

I already told them, they will do, but I think this should be fixed anyway.

Regards,

rbayet:
Hello @Nuranto,

Looks good to me, but it will probably generate a PHPMD Warning/Error due to the if ... else structure.
You can get around that by adding a `@SuppressWarnings(PHPMD.ElseExpression)` in the method's PHPDoc.

Regards,

src/module-elasticsuite-catalog-rule/Model/Rule/Condition/Product/QueryBuilder.php
```diff
@@ -146,12 +146,24 @@ private function getSpecialAttributesSearchQuery(ProductCondition $productCondit
...
+            if (in_array($productCondition->getOperator(), ['()', '!()'])) {
+                // Those operators run on the untouched (keyword) field, where a single terms clause is
+                // equivalent to one match clause per value, but counts as a single clause in Lucene.
                 $queryParams[$clause][] = $this->prepareQuery(
-                    QueryInterface::TYPE_MATCH,
-                    ['field' => $fieldName, 'queryText' => trim($value), 'minimumShouldMatch' => "100%"]
+                    QueryInterface::TYPE_TERMS,
+                    ['field' => $fieldName, 'values' => $values]
                 );
+            } else {
```
rbayet, line 158: Requires a `@SuppressWarnings(PHPMD.ElseExpression)` in the method's PHPDoc

rbayet (changes requested):

PR #3917, opened 2026-08-31
[Catalog][VirtualCategories] Fixes #3859, keep all positions when switching sort order

### LINKED ISSUE #3859: Admin category merchandiser: only pageSize (20) positions kept when switching sort order from "Name" to "Position"
While testing on a 2.4.9 local environment with the latest 2.12.x commit which includes both #3841 and #3858, I encountered a weird behavior when for a category 
* initially with "Product Name" as default sort order
* where all existing 48 (negative) positions were migrated as positive and contiguous positive positions from the legacy table to the `smile_virtualcategory_catalog_category_product_position`
if the category is edited in admin and the sort order is changed from "Product Name" to "Position" then only 20 products are shown in "Manual order"

If I put breakpoints in \Smile\ElasticsuiteVirtualCategory\Plugin\Catalog\Category\DataProviderPlugin at 
```
$data[$currentCategory->getId()]['default']['sorted_products'] = $this->getProductSavedPositions($globalCategory);
```
and
```
$data[$currentCategory->getId()]['sorted_products']         = $this->getProductSavedPositions($currentCategory);
```
I can see that, initially (when the category is first edited with sort order as "Product Name") the Ajax preview call  correctly sends by POST 
* a `sorted_products` var with the full list of 48 product positions
* all 48 `product_position[X]` vars
But of course, the preview return an array with 20 products sorted by "Product Name" (and a "size" of 48) ✔️ 

If I then switch the sort order of the category from "Product Name" to "Position" the Ajax preview call sends by POST 
* a `sorted_products` var reduced to 20 product positions
* only 20 `product_position[X]` vars
The Ajax preview call response is 
* an array of 32 products ❓
* sorted by "Position" ✔️ 
* a total size of 48 ✔️ 
Thus the preview is redrawn with only 20 products flagged as "Manual Sort" ❌ 

To me the issue is Javascript / products sorted related as the initial data is correct.

### Preconditions

Magento Version : 2.4.9 EE so far

ElasticSuite Version : 2.12.x (Elasticsuite OpenSource only)

Environment : Developer

Third party modules : None

### Steps to reproduce

1. Have a category with 50 manually assigned products without any (Elasticsuite) position
2. Change its Default sort order to "Product Name" and save it
3. Check that catalog_category_product has positions whatever type of position (negative, zero, positive) 
4. Truncate the `smile_virtualcategory_catalog_category_product_position` table to start afresh
5. Migrate the product positions (all type of position) for that category or the whole catalog
6. Check that you have positive and contiguous positions for the 50 products and for that category in `smile_virtualcategory_catalog_category_product_position`
7. Refresh that category in admin
8. Check that POST vars of the ajax preview call are consistent with the expected number of manually sorted products
9. Check that the ajax preview call returns only 20 products sorted by "Product Name"
10. Switch the Default sort order of the category to "Position" and inspect the refreshed Preview

### Expected result

1. The Preview Ajax call should return by design 50 (ignoring the pageSize) products
2. The refreshed Preview area should show 50 manually "Manual Sort" products

### Actual result

1. The Preview Ajax call returns only 20 + X (< 50) products 
2. The refreshed Preview area show only 20 "Manual Sort" products, the rest are in Automatic Sort

vahonc:


src/module-elasticsuite-virtual-category/Model/Preview.php
```diff
@@ -102,6 +102,20 @@ public function __construct(
...
 
+    /**
+     * {@inheritDoc}
+     */
+    public function getData(): array
+    {
+        $data = parent::getData();
+        // Only sort_by "position" guarantees getSortedProducts() returned every manually positioned
+        // product (see getSortedProductIds() below). Any other sort only returns the current page,
+        // so the client must not use it to decide which manual positions are still valid.
```
rbayet: Please replace "client" par "frontend layer"

rbayet: Please also replace by a whole comment block, ie 
```
/*
  * This is my multiline comment that [...]
  *  and also this things and [...].
  */ 
```

src/module-elasticsuite-catalog/view/adminhtml/web/js/form/element/product-sorter.js
```diff
@@ -207,16 +207,23 @@ define([
...
             this.currentSize(Math.max(this.currentSize(), this.products().length));
 
-            var productIds = products.map(function (product) { return product.getId() });
-            var editPositions = this.editPositions();
-
-            for (var productId in editPositions) {
-                if ($.inArray(parseInt(productId, 10), productIds) < 0) {
-                    delete editPositions[productId];
+            // The returned product list only contains every manually positioned product (as opposed to
+            // just the current preview page) when the backend flags the response as such. Pruning
```
rbayet: The formulation is unclear.
Go for something like that : 
=> 
<< When the backend flags the response with 'is_manual_sort = true', the returned product list contains every manually positioned product. Otherwise only the (unapplied) positions of products on the current preview page are provided.
So only prune editPositions to eliminate obsolete positions when 'is_manual_sort = true'. >>

rbayet: Same goes, max 2 lines with "single line comment //".
Beyond, please use block comments

rbayet (changes requested):

rbayet:
@vahonc Tested, perfect, can you squash your commits ?

vahonc:
Done.

rbayet:
Replaced by https://github.com/Smile-SA/elasticsuite/pull/3919

PR #3922, opened 2026-09-21
[Catalog] Fixes #3826, scope ajax facets to the requested category

### LINKED ISSUE #3826: AJAX filter requests return empty results, ajaxFilterModifier missing from category/search providers
### Preconditions

Magento Version : 2.4.8-p4 open source

ElasticSuite Version : 2.11.18.1

Environment : Both developer and production

Third party modules :

## Summary

The AJAX endpoint `/catalog/navigation_filter/ajax` (used by the filter search input and "Show more" button) returns an empty JSON array `[]` for attributes that are not pinned as layered navigation attributes. Once a filter value is already selected the same endpoint works correctly.

## Root cause

The `ajaxFilterModifier` is registered on the base `FilterableAttributes` type,
but the `categoryFilterableAttributesProvider` and `searchFilterableAttributesProvider`
virtual types override the `modifiersPool` argument without including it:

```xml
<!-- Base type — has ajaxFilterModifier -->
<type name="FilterableAttributes">
    <arguments>
        <argument name="modifiersPool" xsi:type="array">
            <item name="ajaxFilterModifier" xsi:type="object">...AjaxFilter</item>
        </argument>
    </arguments>
</type>

<!-- Virtual types — override modifiersPool WITHOUT ajaxFilterModifier -->
<virtualType name="categoryFilterableAttributesProvider" type="FilterableAttributes">
    <arguments>
        <argument name="modifiersPool" xsi:type="array">
            <item name="coverageModifier" xsi:type="object">...Coverage</item>
            <item name="statisticsModifier" xsi:type="object">...Statistics</item>
            <!-- ajaxFilterModifier is missing here -->
        </argument>
    </arguments>
</virtualType>
```
This causes two problems during AJAX filter requests:

The AjaxFilter modifier never runs, so it cannot filter the attributes list down to the single requested attribute or set the aggregation size to 0 (unlimited).

The Coverage modifier runs unchecked and calculates coverage rates against the wrong product scope. In the AJAX controller context the product collection is not properly scoped to the current category, so coverage is computed against the entire catalog. Attributes whose coverage drops below facetMinCoverageRate in that global scope are removed — even though they already passed the coverage check on the initial page load.

Attributes that are pinned as "layered nav attributes" via LayeredNavAttributesProvider are unaffected because they skip the coverage check, which is why the bug only manifests for non-pinned attributes.

## Steps to reproduce
Navigate to a category page that has a filterable attribute with more values than facetMaxSize (so the "Show more" button / search input appears)
The attribute must not be pinned as a layered navigation attribute for that category Without selecting any filter value, click "Show more" or type in the filter search input Observe that the AJAX response is [] and no additional filter values appear 
If you first check one of the filter values and then repeat step 3, it works
correctly (because currentFilterValue is set, which changes the behavior).

Expected behavior
The AJAX endpoint should return all available filter values regardless of whether a value is currently selected.

## Suggested fix
Add ajaxFilterModifier to both virtual type modifier pools, and have the Coverage modifier skip its check when the AjaxFilter modifier has already narrowed the attributes list to a single requested filter (since it already
passed coverage on the initial page render).

See diff below.

---

## Diff

```diff
diff --git a/src/module-elasticsuite-catalog/Search/Request/Product/Aggregation/Provider/FilterableAttributes/Modifier/Coverage.php b/src/module-elasticsuite-catalog/Search/Request/Product/Aggregation/Provider/FilterableAttributes/Modifier/Coverage.php
index abcdefg..1234567 100644
--- a/src/module-elasticsuite-catalog/Search/Request/Product/Aggregation/Provider/FilterableAttributes/Modifier/Coverage.php
+++ b/src/module-elasticsuite-catalog/Search/Request/Product/Aggregation/Provider/FilterableAttributes/Modifier/Coverage.php
@@ -14,6 +14,7 @@
 namespace Smile\ElasticsuiteCatalog\Search\Request\Product\Aggregation\Provider\FilterableAttributes\Modifier;
 
 use Smile\ElasticsuiteCatalog\Model\Attribute\Source\FilterDisplayMode;
+use Smile\ElasticsuiteCatalog\Model\Attribute\LayeredNavAttributesProvider;
 use Smile\ElasticsuiteCatalog\Search\Request\Product\Aggregation\Provider\FilterableAttributes\ModifierInterface;
 use Smile\ElasticsuiteCatalog\Search\Request\Product\Coverage\ProviderFactory as CoverageProviderFactory;
 use Smile\ElasticsuiteCore\Search\Request\Builder as SearchRequestBuilder;
@@ -31,6 +32,11 @@ class Coverage implements ModifierInterface
      */
     private $coverageRequestBuilder;
 
+    /**
+     * @var \Magento\Framework\App\RequestInterface
+     */
+    private $request;
+
     /**
      * @var \Smile\ElasticsuiteCatalog\Search\Request\Product\Coverage\ProviderFactory
      */
@@ -46,12 +52,15 @@ class Coverage implements ModifierInterface
      * @param SearchRequestBuilder         $coverageRequestBuilder       Coverage Request builder.
      * @param CoverageProviderFactory      $coverageProviderFactory      Coverage provider factory.
      * @param LayeredNavAttributesProvider $layeredNavAttributesProvider Layered navigation attributes provider.
+     * @param \Magento\Framework\App\RequestInterface $request          HTTP Request.
      */
     public function __construct(
         SearchRequestBuilder $coverageRequestBuilder,
         CoverageProviderFactory $coverageProviderFactory,
-        LayeredNavAttributesProvider $layeredNavAttributesProvider
+        LayeredNavAttributesProvider $layeredNavAttributesProvider,
+        \Magento\Framework\App\RequestInterface $request
     ) {
+        $this->request                      = $request;
         $this->coverageRequestBuilder       = $coverageRequestBuilder;
         $this->coverageProviderFactory      = $coverageProviderFactory;
         $this->layeredNavAttributesProvider = $layeredNavAttributesProvider;
@@ -63,6 +72,12 @@ class Coverage implements ModifierInterface
     public function modifyAttributes($storeId, $requestName, $attributes, $query, $filters, $queryFilters)
     {
         $relevantAttributes = [];
+
+        // Skip coverage check during AJAX filter requests — the attribute already
+        // passed coverage on the initial page render.
+        if ($this->request->isAjax() && $this->request->getParam('filterName')) {
+            return $attributes;
+        }
+
         $coverageRates      = $this->getCoverageRates($storeId, $requestName, $query, $filters, $queryFilters);
 
         foreach ($attributes as $attribute) {
diff --git a/src/module-elasticsuite-catalog/etc/frontend/di.xml b/src/module-elasticsuite-catalog/etc/frontend/di.xml
index abcdefg..1234567 100644
--- a/src/module-elasticsuite-catalog/etc/frontend/di.xml
+++ b/src/module-elasticsuite-catalog/etc/frontend/di.xml
@@ -218,6 +218,7 @@
             <argument name="attributeList" xsi:type="object">Smile\ElasticsuiteCatalog\Search\Request\Product\Aggregation\Provider\FilterableAttributes\Search\AttributeList</argument>
             <argument name="modifiersPool" xsi:type="array">
+                <item name="ajaxFilterModifier" xsi:type="object">Smile\ElasticsuiteCatalog\Search\Request\Product\Aggregation\Provider\FilterableAttributes\Modifier\AjaxFilter</item>
                 <item name="coverageModifier" xsi:type="object">Smile\ElasticsuiteCatalog\Search\Request\Product\Aggregation\Provider\FilterableAttributes\Modifier\Coverage</item>
                 <item name="statisticsModifier" xsi:type="object">Smile\ElasticsuiteCatalog\Search\Request\Product\Aggregation\Provider\FilterableAttributes\Modifier\Statistics</item>
             </argument>
@@ -228,6 +229,7 @@
             <argument name="attributeList" xsi:type="object">Smile\ElasticsuiteCatalog\Search\Request\Product\Aggregation\Provider\FilterableAttributes\Category\AttributeList</argument>
             <argument name="modifiersPool" xsi:type="array">
+                <item name="ajaxFilterModifier" xsi:type="object">Smile\ElasticsuiteCatalog\Search\Request\Product\Aggregation\Provider\FilterableAttributes\Modifier\AjaxFilter</item>
                 <item name="coverageModifier" xsi:type="object">Smile\ElasticsuiteCatalog\Search\Request\Product\Aggregation\Provider\FilterableAttributes\Modifier\Coverage</item>
                 <item name="statisticsModifier" xsi:type="object">Smile\ElasticsuiteCatalog\Search\Request\Product\Aggregation\Provider\FilterableAttributes\Modifier\Statistics</item>
             </argument>
```

klo-chi:
Fixes #3826.

### Preconditions

- Magento 2.4.x, ElasticSuite 2.10.x / 2.11.x / 2.12.x (reproduced on 2.11.18.1, code unchanged up to 2.12.1.1)
- Stores > Configuration > Catalog > Catalog > Layered Navigation > Display Category Filter = **No** (`catalog/layered_navigation/display_category = 0`)
- An attribute filter with more options than its facet max size, so the "Show more" link is displayed

### Steps to reproduce

1. On Luma, open a category, e.g. Gear > Bags (the case @rbayet reproduced in #3826).
2. Click "Show more" on a facet (or type in the facet search box).

**Expected:** the extra options of the current category, with the same counts as on the page.
**Actual:** every option of the catalog with catalog-wide counts. Picking an option that is not in the category leads to an empty product list.

### Root cause

With Display Category Filter = No, the category restriction only comes from the search context current category (`Model/Product/Search/Request/Container/Filter/CurrentCategory`).

`Controller/Navigation/Filter/Ajax::initLayer()` calls `Layer::setCurrentCategory($category)`. Magento's implementation calls `getCurrentCategory()` first to compare ids, which returns the store root category. `Plugin/Layer/CategoryPlugin::afterGetCurrentCategory()` stores it in the search context because the context is still empty, and since be691bc75a it no longer overwrites a non-empty context. So the context keeps the root category and the facet request is not scoped to the requested category.

With Display Category Filter = Yes the issue does not show because `Model/Layer/Filter/Category::apply()` sets the search context from the `cat` parameter.

### Fix

Set the search context current category explicitly right after `setCurrentCategory()` in the ajax controller.

The search context is injected as a new optional trailing constructor argument with an `ObjectManager` fallback (same backward compatibility pattern as `Model/Preview.php` in the virtual category module), so subclasses and existing DI configuration keep working.

Side note: for a search request (`q`) that also carries `cat`, the facets are now scoped to `cat` when Display Category Filter = No. The renderer only adds `cat` on category pages (level > 1), so this does not happen with the standard layered navigation, and it matches what the Category filter already does when the setting is Yes.

### Tests

On a Magento 2.4.7-p6 / ElasticSuite 2.11.18.1 store, Display Category Filter = No, for every attribute facet of two categories (an anchor parent and a leaf), with and without an attribute filter applied, comparing the "Show more" ajax response with the facet items rendered on the category page (labels and counts):

| | Mismatching facets |
|---|---|
| Without the fix | 25 of 28 |
| With the fix, Display Category Filter = No | 0 of 28 |
| With the fix, Display Category Filter = Yes | 0 of 28 |

Also checked:
- `php -l` on PHP 7.4, 8.1 and 8.3.
- phpcs (MEQP2 severity 9 and SmileLab) and phpmd with the rulesets used by the quality workflow: no new violations.


rbayet:
Correctly fixes the issue, the two facets (pun intended) "Too many irrelevant options" and "No options at all" (when clicking on "Show more") of the problem are linked to the attribute facet coverage rate 
* with a high coverage rate (90), the "No options at all" is more likely
* with a low coverage rate (1), the "Too many irrelevant options" is more likely

